PR #125, opened 2017-12-18
add tests for protocol core

SaitTalhaNisanci:


internal/protocol/core_test.go
```diff
@@ -0,0 +1,324 @@
...
+
+func TestDataEntryView_CreationTime(t *testing.T) {
+	entryView := EntryView{}
+	entryView.creationTime = 5
+	if result := entryView.CreationTime(); result != 5 {
+		t.Fatal("entryView CreationTime failed")
+	}
+}
+
+func TestDataEntryView_ExpirationTime(t *testing.T) {
```
mustafaiman: What are you testing here and the other tests like this one?

SaitTalhaNisanci: @mustafaiman  these tests will be changed.  I will update this PR.

PR #126, opened 2017-12-19
fix data race in heartbeat

SaitTalhaNisanci:


asimarslan:
verify

sancar:
Fixed in https://github.com/hazelcast/hazelcast-go-client/pull/128

PR #135, opened 2017-12-29
add more test to serialization, fix bug

furkansenharputlu:
- There was a bug in registering class definitions. It is fixed.


internal/serialization/portable_serializer_test.go
```diff
@@ -229,8 +285,34 @@ func TestPortableSerializer2(t *testing.T) {
...
+	config2.AddPortableFactory(2, &PortableFactory1{})
+
+	service2 := NewSerializationService(config2)
+	ret3, _ := service2.ToObject(data)
+
+	if !reflect.DeepEqual(ret, expectedRet) || !reflect.DeepEqual(ret2, expectedRet2) || !reflect.DeepEqual(ret3, expectedRet) {
+		t.Errorf("ReadObject() failed!")
+	}
+
+}
```
mustafaiman: As far as I understand, there are two tests here.
* One serialization service can serialize and deserialize two different portables without messing things up
* Two different serialization services can read and write the same portable
Can you split the different tests and name them accordingly?

PR #141, opened 2018-01-09
fix memory leakage in invocationService

SaitTalhaNisanci:
Go subroutine in invocation to detect invocation timeout would still be alive after invocation was completed, which was causing memory leakage. So when invocation gets an error or response from server we needed to terminate the go subroutine.


SaitTalhaNisanci:
verify

PR #143, opened 2018-01-10
unregister invocations when there is any kind of error

SaitTalhaNisanci:
Unregistering an invocation was done only when server responds, however there are some other cases such as 'connection closed'. That was causing a memory leak. Since every error will go to handleException we unregister any invocation that comes there.  Unregistering the same invocation twice doesnt have any side effects.

### DIFF
--- a/internal/invocation.go
+++ b/internal/invocation.go
@@ -83,19 +83,20 @@ func (invocation *Invocation) Result() (*ClientMessage, error) {
 }
 
 type InvocationService struct {
-	client                    *HazelcastClient
-	quit                      chan struct{}
-	nextCorrelation           int64
-	responseWaitings          map[int64]*Invocation
-	eventHandlers             map[int64]*Invocation
-	sending                   chan *Invocation
-	responseChannel           chan *ClientMessage
-	cleanupConnectionChannel  chan *ConnectionAndError
-	removeEventHandlerChannel chan int64
-	notSentMessages           chan int64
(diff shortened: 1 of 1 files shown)


SaitTalhaNisanci:
verify

internal/invocation.go
```diff
@@ -326,6 +333,7 @@ func (invocationService *InvocationService) cleanupConnectionInternal(connection
 
 }
 func (invocationService *InvocationService) handleException(invocation *Invocation, err error) {
+	invocationService.unRegisterInvocationChannel <- invocation.request.CorrelationId()
```
mustafaiman: why remove the invocation so early? Maybe it is retryable. Can't we retry it with the same invocation id?

SaitTalhaNisanci: if it is retryable then it will be reregistered, so in both cases we need to unregister the previous invocation.

SaitTalhaNisanci: @mustafaiman  This is how it will be reregistered.

https://github.com/hazelcast/hazelcast-go-client/pull/143/files#diff-012e965d1afb9578566f072d6fd469ccR362

https://github.com/hazelcast/hazelcast-go-client/pull/143/files#diff-012e965d1afb9578566f072d6fd469ccR155

mustafaiman: I thought maybe we could re-use the same correlation id for retrying but apparently Java client also does the same thing. So I'm ok.

PR #149, opened 2018-01-15
serialize int as int64

furkansenharputlu:
The `int` type is usually 32 bits wide on 32-bit systems and 64 bits wide on 64-bit systems. Therefore, we decided it to serialize as `int64` type.


furkansenharputlu:
verify

PR #153, opened 2018-02-07
add list codecs

SaitTalhaNisanci:
adds list codecs, #151


internal/protocol/list_containsall.go
```diff
@@ -0,0 +1,57 @@
...
+	. "github.com/hazelcast/hazelcast-go-client/internal/serialization"
+
+	. "github.com/hazelcast/hazelcast-go-client/internal/common"
+)
+
+type ListContainsAllResponseParameters struct {
+	Response bool
+}
+
+func ListContainsAllCalculateSize(name *string, values []*Data) int {
```
furkansenharputlu: Also, this pr includes many `[]*Data`s. They should be `[]Data`.

Issue #152, opened 2018-02-07
Set


SaitTalhaNisanci:
- [x] Backport the related Hazelcast Java interface into go repo.
- [x] Document the interface GoDocs.
- [x] Generate the related proxy codecs of client protocol
- [x] Implement proxy and write unit tests
- [ ] Write a sample for this proxy

PR #154, opened 2018-02-07
add set codecs

SaitTalhaNisanci:
adds set codecs, #152.

### DIFF
--- a/internal/protocol/set_addall.go
+++ b/internal/protocol/set_addall.go
@@ -1,0 +1,57 @@
+// Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package protocol
+
+import (
+	. "github.com/hazelcast/hazelcast-go-client/internal/serialization"
+
+	. "github.com/hazelcast/hazelcast-go-client/internal/common"
+)
+
+type SetAddAllResponseParameters struct {
+	Response bool
+}
+
+func SetAddAllCalculateSize(name *string, valueList []*Data) int {
+	// Calculates the request payload size
+	dataSize := 0
+	dataSize += StringCalculateSize(name)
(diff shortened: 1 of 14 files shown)


internal/protocol/set_addall.go
```diff
@@ -0,0 +1,57 @@
...
+	. "github.com/hazelcast/hazelcast-go-client/internal/serialization"
+
+	. "github.com/hazelcast/hazelcast-go-client/internal/common"
+)
+
+type SetAddAllResponseParameters struct {
+	Response bool
+}
+
+func SetAddAllCalculateSize(name *string, valueList []*Data) int {
```
furkansenharputlu: This should not be pointer array. Like this `[]Data`.

internal/protocol/set_addall.go
```diff
@@ -0,0 +1,57 @@
...
+	dataSize := 0
+	dataSize += StringCalculateSize(name)
+	dataSize += INT_SIZE_IN_BYTES
+	for _, valueListItem := range valueList {
+		dataSize += DataCalculateSize(valueListItem)
+	}
+	return dataSize
+}
+
+func SetAddAllEncodeRequest(name *string, valueList []*Data) *ClientMessage {
```
furkansenharputlu: Here, also. There are many `[]*Data`s to be changed to `[]Data`.

SaitTalhaNisanci:
@furkansenharputlu 
For codecs we are using flyweight model, so for example DataCalculateSize takes a pointer and if we dont use `[]*Data` we would be copying the content of Data when calculating the size which is not necessary i.e. `DataCalculateSize(*data)`. Can you explain why you think it should be the other way ?

SaitTalhaNisanci:
For example https://github.com/hazelcast/hazelcast-go-client/pull/154/files#diff-421d1d29fb9bb75de9fba568f5d9203bR49 here 'ReadData()' returns a pointer and if we didnt make it a pointer slice we would be copying the content of Data and taking the address afterwards, am I missing something ?

furkansenharputlu:
When I run this code:
```package main

import "fmt"

type student struct {
	name string
}

func main() {
	var array []student = []student{student{"furkan"}, student{"talha"}}
	x(array)
	fmt.Println(array)
}

func x(pass []student) {
	pass[1].name = "talha nişancı"
}

```

The result is:
```[{furkan} {talha nişancı}]```

This is why I am saying we should use `[]Data`. I think passed slice is not deeply copied.

SaitTalhaNisanci:
Your example is about slices so. Everything is pass by value in go except slices and maps, your example is about `*[]Data` vs `[]Data` not about `[]*Data` vs `[]Data`.  I am talking about this https://play.golang.org/p/rFJpczSd6Hn

furkansenharputlu:
I understand your reasoning. We both think same about slices. They are not deeply copied while passing as parameter. However, you try to keep consistency in codecs so you use slice of pointers. 
Ok 👍

PR #156, opened 2018-02-08
add List interface with Godocs

SaitTalhaNisanci:
#151.
This pr adds IList interface with its docs.

### DIFF
--- a/core/IList.go
+++ b/core/IList.go
@@ -1,0 +1,107 @@
+// Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License")
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


SaitTalhaNisanci:
verify

sancar:
There seem to be different decisions made between Ilist and ISet when picking a method. 
`getAll` and `toSlice` for example seems to be doing the same thing. In one pr `getAll` used, in other `toSlice`. 
Can we introduce an ICollection interface and make ISet, IList, IQueue extend that? This way we can avoid different names for same functionality more easily.

core/IList.go
```diff
@@ -0,0 +1,103 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package core
+
+// Concurrent, distributed implementation of List.
```
asimarslan: In golang context word `List` does not mean much.  go list is total different from Hazelcast API. 
I think we need a better explanation here to replace List with something more generic like array like container (or some better formal expression)

PR #159, opened 2018-02-10
add ReplicatedMap interface with GoDocs

furkansenharputlu:
#158

### DIFF
--- a/core/replicated_map.go
+++ b/core/replicated_map.go
@@ -1,0 +1,118 @@
+// Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License")
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


core/replicated_map.go
```diff
@@ -0,0 +1,118 @@
...
+package core
+
+import (
+	. "github.com/hazelcast/hazelcast-go-client/serialization"
+	"time"
+)
+
+// ReplicatedMap is a map-like data structure with weak consistency
+// and values locally stored on every node of the cluster.
+//
```
SaitTalhaNisanci: There shouldnt be any space here otherwise it wont be generated with goDoc

core/replicated_map.go
```diff
@@ -0,0 +1,118 @@
...
+// Whenever a value is written asynchronously, the new value will be internally
+// distributed to all existing cluster members, and eventually every node will have
+// the new value.
+//
```
SaitTalhaNisanci: should remove the space

core/replicated_map.go
furkansenharputlu: No, it generates. The space is used for new paragraph.

core/replicated_map.go
furkansenharputlu: Also here, it is for new paragraph.

PR #162, opened 2018-02-12
add Queue interface with Godocs

SaitTalhaNisanci:
#161.

### DIFF
--- a/core/IQueue.go
+++ b/core/IQueue.go
@@ -1,0 +1,106 @@
+// Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License")
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package core
+
+// IQueue is a concurrent, blocking, distributed, observable queue. Queue is not a partitioned data-structure.
+// All of the Queue content is stored in a single machine (and in the backup).
+// Queue will not scale by adding more members in the cluster.
+type IQueue interface {
+
+	// IDistributedObject is the base interface for all distributed objects.
+	IDistributedObject
+
+	// AddAll adds all the items. If items slice changes during this operation, behavior is unspecified.
+	// AddAll returns true if the queue has changed, false otherwise.
+	AddAll(items []interface{}) (changed bool, err error)
+
+	// AddItemListener adds an item listener for this queue.
+	// The listener will be invoked for any add/remove item event.
+	// AddItemListener returns the registrationID of the listener.
+	AddItemListener(listener interface{}, includeValue bool) (registrationID *string, err error)
+
+	// Clear removes all of the items in this queue.
+	Clear() (err error)
+
+	// Contains returns true if the given item is in this queue.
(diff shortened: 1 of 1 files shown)


sancar:
What is the difference between `offer` and `add`?

SaitTalhaNisanci:
In terms of functionality there isnt any difference, should we remove add or offer ? would it be okay to keep both of them?

core/IQueue.go
```diff
@@ -0,0 +1,111 @@
...
+// All of the Queue content is stored in a single machine (and in the backup).
+// Queue will not scale by adding more members in the cluster.
+type IQueue interface {
+
+	// IDistributedObject is the base interface for all distributed objects.
+	IDistributedObject
+
+	// Add adds the given item to the end of the queue. Operation is successful only
+	// if queue has required capacity.
+	// Add returns true if the item is added, false otherwise.
```
mustafaiman: If the item cannot be added this method should set `err` with an appropriate error. If `err` is set, return value should not matter. If the item is added it returns `true`, otherwise `err` is set. Therefore this method never returns `false`.

core/IQueue.go
```diff
@@ -0,0 +1,111 @@
...
+	// returning nil.
+	// PollWithTimeout returns the head of this queue, nil if the queue is empty after timeoutInMilliSeconds milliseconds.
+	PollWithTimeout(timeoutInMilliSeconds int64) (item interface{}, err error)
+
+	// Put inserts the item at the end of this queue. It blocks until
+	// space becomes available if necessary.
+	Put(item interface{}) (err error)
+
+	// RemainingCapacity returns the number of additional items this queue can contain.
+	RemainingCapacity() (remainingCapacity int64, err error)
```
mustafaiman: this returns int32 on the Java side. see https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientQueueProxy.java#L208

sancar:
@SaitTalhaNisanci I think we should remove `add` it is the version that throws an exception. 
In go client, `offer` seems enough. 

I see that you choose not to include `remove()` also. It is the same functinality with `poll()` but once throws exception when item is not avaliable. With same reasoning, we could avoid adding `add` .

sancar:
1. remove(object)
2. remove()
3. poll()

There are 3 methods. The functionality of the 1st one is different from others, it should stay. 
2nd and 3rd ones are same and only differs via the exception. I think we should remove 2nd one.

core/IQueue.go
```diff
@@ -0,0 +1,102 @@
...
+	Poll() (item interface{}, err error)
+
+	// PollWithTimeout retrieves and removes the head of this queue,
+	// if the queue is empty it waits timeoutInMilliSeconds milliseconds before
+	// returning nil.
+	// PollWithTimeout returns the head of this queue, nil if the queue is empty after timeoutInMilliSeconds milliseconds.
+	PollWithTimeout(timeoutInMilliSeconds int64) (item interface{}, err error)
+
+	// Put inserts the item at the end of this queue. It blocks until
+	// space becomes available if necessary.
```
Serdaro: It has been used 'room' when referring to inserting items to the queue and queue has room for this operation; just to be consistent within the integrity of code comments, maybe it is better to change this sentence as "It blocks until there is available room in the queue, if necessary."

core/IQueue.go
```diff
@@ -0,0 +1,106 @@
...
+
+// IQueue is a concurrent, blocking, distributed, observable queue. Queue is not a partitioned data-structure.
+// All of the Queue content is stored in a single machine (and in the backup).
+// Queue will not scale by adding more members in the cluster.
+type IQueue interface {
+
+	// IDistributedObject is the base interface for all distributed objects.
+	IDistributedObject
+
+	// AddAll adds all the items. If items slice changes during this operation, behaviour is unspecified.
```
Serdaro: ....behavior...

PR #165, opened 2018-02-12
add MultiMap interface with GoDocs

furkansenharputlu:
#164

### DIFF
--- a/core/multi_map.go
+++ b/core/multi_map.go
@@ -1,0 +1,146 @@
+// Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License")
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package core
+
+import "time"
+
+// MultiMap is a specialized map whose keys can be associated with multiple values.
+type MultiMap interface {
+	// IDistributedObject is the base interface for all distributed objects.
+	IDistributedObject
+
+	// Stores a key-value pair in the multi-map.
+	// It returns true if size of the multi-map is increased, false if the multi-map
+	// already contains the key-value pair.
+	Put(key interface{}, value interface{}) (increased bool, err error)
+
+	// Get returns a slice of values associated with the specified key.
+	Get(key interface{}) (values []interface{}, err error)
+
+	// Remove removes an association of the specified value with the specified key. Calling this method does not affect
+	// other values associated with the same key.
+	// It returns true if the value was detached from the specified key, false if it was not.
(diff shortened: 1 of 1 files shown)


sancar:
`delete` seems missing

core/multi_map.go
```diff
@@ -0,0 +1,143 @@
...
+
+// MultiMap is a specialized map whose keys can be associated with multiple values.
+type MultiMap interface {
+	// IDistributedObject is the base interface for all distributed objects.
+	IDistributedObject
+
+	// Stores a key-value pair in the multi-map.
+	// It returns true if size of the multi-map is increased, false if the multi-map
+	// already contains the key-value pair.
+	Put(key interface{}, value interface{}) (increased bool, err error)
```
SaitTalhaNisanci: I think it would be better to use a more descriptive return name here, maybe "*changed*" ?

core/multi_map.go
```diff
@@ -0,0 +1,143 @@
...
+	Get(key interface{}) (values []interface{}, err error)
+
+	// Remove removes an association of the specified value with the specified key. Calling this method does not affect
+	// other values associated with the same key.
+	// It returns true if the value was detached from the specified key, false if it was not.
+	Remove(key interface{}, value interface{}) (removed bool, err error)
+
+	// RemoveAll detaches all values from the specified key.
+	// It returns a slice of old values that were associated with this key prior to this method call.
+	RemoveAll(key interface{}) (oldValues []interface{})
```
SaitTalhaNisanci: This method should also return an error.

core/multi_map.go
```diff
@@ -0,0 +1,143 @@
...
+	ContainsEntry(key interface{}, value interface{}) (found bool, err error)
+
+	// Clear removes all entries from this multi-map.
+	Clear() (err error)
+
+	// Size returns the total number of values in this multi-map.
+	Size() (size int32, err error)
+
+	// ValueCount returns the number of values associated with the specified key.
+	ValueCount(key interface{}) (valueCount int32)
```
SaitTalhaNisanci: This method should also return an error

furkansenharputlu:
@sancar What do you mean by `delete`, a method ? I cannot see it in Java client

core/multi_map.go
furkansenharputlu: I think "*increased*" is descriptive enough.

sancar:
@furkansenharputlu 
It seems `delete` is recently added(3.10). Here:
https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMultiMapProxy.java#L153

core/multi_map.go
asimarslan: `changed` is not correct as it includes decrease.

We can use something like `mapSizeIncreased`

core/multi_map.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Stores a key-value pair in the multi-map.
+	// It returns true if size of the multi-map is increased, false if the multi-map
+	// already contains the key-value pair.
+	Put(key interface{}, value interface{}) (increased bool, err error)
+
+	// Get returns a slice of values associated with the specified key.
+	Get(key interface{}) (values []interface{}, err error)
+
+	// Remove removes an association of the specified value with the specified key. Calling this method does not affect
```
asimarslan: First `Remove` is not needed.  
"Removes and association ... " is better

PR #166, opened 2018-02-14
add ICollection

SaitTalhaNisanci:
ICollection has common methods for `List`,`Queue`and `Set`. 
Users can cast their collection to `ICollection` and execute the same code on them.


SaitTalhaNisanci:
As the client team we have decided that `ICollection` is not necessary.

PR #168, opened 2018-02-15
add Topic interface with GoDocs

SaitTalhaNisanci:
#167.

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -240,3 +240,31 @@ type ILifecycleListener interface {
 	// LifecycleStateChanged is called when instance's state changes. No blocking calls should be made in this method.
 	LifecycleStateChanged(string)
 }
+
+// TopicMessageListener is a listener for ITopic.
+// Provided that a TopicMessageListener is not registered twice, a TopicMessageListener will never be called concurrently. So there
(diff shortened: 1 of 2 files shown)


core/api.go
```diff
@@ -240,3 +240,31 @@ type ILifecycleListener interface {
...
+}
+
+// TopicMessage is a message for ITopic.
+type TopicMessage interface {
+
+	// MessageObject returns the published message.
+	MessageObject() interface{}
+
+	// PublishTime returns the time when the message is published.
+	PublishTime() int64
```
asimarslan: unit of time is MilliSeconds. Other API docs did not mention too. This can be default of Java context but golang uses second or nanosecond.
see time.Unix()

core/api.go
```diff
@@ -240,3 +240,31 @@ type ILifecycleListener interface {
...
+
+	// PublishMember returns the member that published the message.
+	// The member can be nil if:
+	//    - The message was sent by a client and not a member.
+	//    - The member, that sent the message, left the cluster before the message was processed.
+	PublishMember() IMember
```
asimarslan: should be `PublishingMember`

PR #176, opened 2018-03-02
add validateAndSerialize methods to simplify proxies

SaitTalhaNisanci:
This PR adds ValidateAndSerialize helper methods so that proxies can be simplified. 
This PR also changes the errors to HazelcastErrors.

### DIFF
--- a/internal/proxy.go
+++ b/internal/proxy.go
@@ -15,6 +15,8 @@
 package internal
 
 import (
+	"github.com/hazelcast/hazelcast-go-client/core"
+	. "github.com/hazelcast/hazelcast-go-client/internal/common"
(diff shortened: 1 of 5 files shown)


internal/proxy.go
```diff
@@ -41,6 +43,56 @@ func (proxy *proxy) ServiceName() string {
 	return *proxy.serviceName
 }
 
+func (proxy *proxy) ValidateAndSerialize1(arg1 interface{}) (arg1Data *Data, err error) {
```
mustafaiman: I think there is no need for `1` at the end of the method name.

mustafaiman: Also validate methods should be private

PR #184, opened 2018-03-05
Queue implementation and unit tests

asimarslan:


internal/queue.go
```diff
@@ -0,0 +1,224 @@
...
+	itemData, err := queue.validateAndSerializeSlice(items)
+	if err != nil {
+		return false, err
+	}
+	request := QueueContainsAllEncodeRequest(queue.name, itemData)
+	responseMessage, err := queue.Invoke(request)
+	return queue.DecodeToBoolAndError(responseMessage, err, QueueContainsAllDecodeResponse)
+}
+
+func (queue *QueueProxy) DrainTo(slice *[]interface{}) (movedAmount int32, err error) {
```
SaitTalhaNisanci: You can use this here :
```
func (queue *QueueProxy) DrainTo(slice *[]interface{}) (movedAmount int32, err error) {
	request := QueueDrainToEncodeRequest(queue.name)
	responseMessage, err := queue.Invoke(request)
	resultSlice, err := queue.DecodeToInterfaceSliceAndError(responseMessage, err, QueueDrainToDecodeResponse)
	if err != nil {
		return 0, err
	}
	*slice = append(*slice, resultSlice...)
	return int32(len(resultSlice)), nil
}
```

internal/queue.go
```diff
@@ -0,0 +1,224 @@
...
+	itemData, err := queue.validateAndSerialize(item)
+	if err != nil {
+		return false, err
+	}
+	request := QueueOfferEncodeRequest(queue.name, itemData, 0)
+	responseMessage, err := queue.Invoke(request)
+	return queue.DecodeToBoolAndError(responseMessage, err, QueueOfferDecodeResponse)
+}
+
+func (queue *QueueProxy) OfferWithTimeout(item interface{}, timeoutInMilliSeconds int64) (added bool, err error) {
```
mustafaiman: We can offer a TimeUnit choice. See: https://github.com/hazelcast/hazelcast-go-client/blob/master/internal/map.go#L106

asimarslan: That timeUnit exist in Java too. I'll keep it same as with Java interface and have only millis in queue otherwise we need to add these timeUnits in many more places.

mustafaiman: You say you will keep this as the same as Java. Java Client's Queue implementation has `TimeUnit` too: https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientQueueProxy.java#L183

asimarslan: You are right, all of them have. Let me update these.

internal/queue.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return 0, err
+	}
+	*slice = append(*slice, resultSlice...)
+	return int32(len(resultSlice)), nil
+}
+
+func (queue *QueueProxy) DrainToWithMaxSize(slice *[]interface{}, maxElements int32) (movedAmount int32, err error) {
+	if slice == nil {
+		return 0, core.NewHazelcastNilPointerError(NIL_ARG_IS_NOT_ALLOWED, nil)
```
SaitTalhaNisanci: Here we can use `NIL_SLICE_IS_NOT_ALLOWED` to be consistent.

internal/queue.go
SaitTalhaNisanci: We can do it similar to `PutTransient` in map. https://github.com/hazelcast/hazelcast-go-client/blob/master/internal/map.go#L62

PR #185, opened 2018-03-05
implement Replicated Map with tests

SaitTalhaNisanci:
#158.

This PR is continuation of #180.

### DIFF
--- a/internal/common/util.go
+++ b/internal/common/util.go
@@ -43,12 +43,6 @@ func GetIpAndPort(addr string) (string, int32) {
 	return addr, int32(port)
 }
 
-func CheckNotNil(v interface{}) bool {
-	return v != nil
-}
-func CheckNotEmpty(v []interface{}) bool {
-	return len(v) > 0
-}
 func GetTimeInMilliSeconds(time int64, duration time.Duration) int64 {
 	var timeInMillis int64 = time * duration.Nanoseconds() / (1000000)
 	if time > 0 && timeInMillis == 0 {
--- a/internal/replicated_map.go
+++ b/internal/replicated_map.go
@@ -1,0 +1,267 @@
+// Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License")
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package internal
+
+import (
+	. "github.com/hazelcast/hazelcast-go-client/core"
+	. "github.com/hazelcast/hazelcast-go-client/internal/common"
+	. "github.com/hazelcast/hazelcast-go-client/internal/protocol"
+	"github.com/hazelcast/hazelcast-go-client/internal/serialization"
+	. "github.com/hazelcast/hazelcast-go-client/serialization"
+	"math/rand"
+	"time"
+)
+
+type ReplicatedMapProxy struct {
+	*proxy
+	targetPartitionId int32
+}
+
+func newReplicatedMapProxy(client *HazelcastClient, serviceName *string, name *string) *ReplicatedMapProxy {
+	partitionCount := client.PartitionService.PartitionCount()
+	targetPartitionId := rand.Int31n(partitionCount)
+	return &ReplicatedMapProxy{proxy: &proxy{client, serviceName, name}, targetPartitionId: targetPartitionId}
+}
+func (rmp *ReplicatedMapProxy) Put(key interface{}, value interface{}) (oldValue interface{}, err error) {
+	return rmp.PutWithTtl(key, value, ttlUnlimited, time.Second)
+}
+
+func (rmp *ReplicatedMapProxy) PutWithTtl(key interface{}, value interface{}, ttl int64, ttlTimeUnit time.Duration) (oldValue interface{}, err error) {
+	keyData, valueData, err := rmp.validateAndSerialize2(key, value)
+	if err != nil {
+		return nil, err
+	}
+	request := ReplicatedMapPutEncodeRequest(rmp.name, keyData, valueData, ttl)
+	responseMessage, err := rmp.InvokeOnKey(request, keyData)
+	return rmp.DecodeToObjectAndError(responseMessage, err, ReplicatedMapPutDecodeResponse)
+}
+
+func (rmp *ReplicatedMapProxy) PutAll(entries map[interface{}]interface{}) (err error) {
+	if entries == nil {
+		return NewHazelcastNilPointerError(NIL_MAP_IS_NOT_ALLOWED, nil)
+	}
+	pairs := make([]*Pair, len(entries))
+	index := 0
+	for key, value := range entries {
+		keyData, valueData, err := rmp.validateAndSerialize2(key, value)
+		if err != nil {
+			return err
+		}
+		pairs[index] = NewPair(keyData, valueData)
+		index++
+	}
+	request := ReplicatedMapPutAllEncodeRequest(rmp.name, pairs)
+	_, err = rmp.InvokeOnRandomTarget(request)
+	return err
+}
+
+func (rmp *ReplicatedMapProxy) Get(key interface{}) (value interface{}, err error) {
+	keyData, err := rmp.validateAndSerialize(key)
+	if err != nil {
+		return nil, err
+	}
+	request := ReplicatedMapGetEncodeRequest(rmp.name, keyData)
+	responseMessage, err := rmp.InvokeOnKey(request, keyData)
+	return rmp.DecodeToObjectAndError(responseMessage, err, ReplicatedMapGetDecodeResponse)
+}
+
+func (rmp *ReplicatedMapProxy) ContainsKey(key interface{}) (found bool, err error) {
+	keyData, err := rmp.validateAndSerialize(key)
+	if err != nil {
+		return false, err
+	}
+	request := ReplicatedMapContainsKeyEncodeRequest(rmp.name, keyData)
+	responseMessage, err := rmp.InvokeOnKey(request, keyData)
+	return rmp.DecodeToBoolAndError(responseMessage, err, ReplicatedMapContainsKeyDecodeResponse)
+}
+
+func (rmp *ReplicatedMapProxy) ContainsValue(value interface{}) (found bool, err error) {
+	valueData, err := rmp.validateAndSerialize(value)
+	if err != nil {
+		return false, err
+	}
+	request := ReplicatedMapContainsValueEncodeRequest(rmp.name, valueData)
+	responseMessage, err := rmp.InvokeOnKey(request, valueData)
+	return rmp.DecodeToBoolAndError(responseMessage, err, ReplicatedMapContainsValueDecodeResponse)
+}
+
+func (rmp *ReplicatedMapProxy) Clear() (err error) {
+	request := ReplicatedMapClearEncodeRequest(rmp.name)
+	_, err = rmp.InvokeOnRandomTarget(request)
+	return err
+}
+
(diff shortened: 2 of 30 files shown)


internal/replicated_map.go
```diff
@@ -0,0 +1,256 @@
...
+	partitionCount := client.PartitionService.PartitionCount()
+	targetPartitionId := rand.Int31n(partitionCount)
+	return &ReplicatedMapProxy{proxy: &proxy{client, serviceName, name}, targetPartitionId: targetPartitionId}
+}
+func (rmp *ReplicatedMapProxy) Put(key interface{}, value interface{}) (oldValue interface{}, err error) {
+	keyData, valueData, err := rmp.validateAndSerialize2(key, value)
+	if err != nil {
+		return nil, err
+	}
+	request := ReplicatedMapPutEncodeRequest(rmp.name, keyData, valueData, TTL)
```
mustafaiman: `TTL` here comes from map.go. This is not good practice. I dont think this constant adds value. Straight `0` would be more readable. If you really want to use a constant, name it better (`TTL_UNLIMITED`) and make it private to this proxy.

internal/replicated_map.go
```diff
@@ -0,0 +1,256 @@
...
+	targetPartitionId int32
+}
+
+func newReplicatedMapProxy(client *HazelcastClient, serviceName *string, name *string) *ReplicatedMapProxy {
+	partitionCount := client.PartitionService.PartitionCount()
+	targetPartitionId := rand.Int31n(partitionCount)
+	return &ReplicatedMapProxy{proxy: &proxy{client, serviceName, name}, targetPartitionId: targetPartitionId}
+}
+func (rmp *ReplicatedMapProxy) Put(key interface{}, value interface{}) (oldValue interface{}, err error) {
+	keyData, valueData, err := rmp.validateAndSerialize2(key, value)
```
mustafaiman: You may reuse `PutWithTtl` here.

internal/common/util.go
```diff
@@ -46,9 +47,18 @@ func GetIpAndPort(addr string) (string, int32) {
 func CheckNotNil(v interface{}) bool {
 	return v != nil
 }
+
+func AssertNotNil(argument interface{}, argName string) error {
```
mustafaiman: I dont see this function being used in this PR.

internal/replicated_map.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (rmp *ReplicatedMapProxy) EntrySet() (resultPairs []IPair, err error) {
+	request := ReplicatedMapEntrySetEncodeRequest(rmp.name)
+	responseMessage, err := rmp.InvokeOnPartition(request, rmp.targetPartitionId)
+	return rmp.DecodeToPairSliceAndError(responseMessage, err, ReplicatedMapEntrySetDecodeResponse)
+}
+
+func (rmp *ReplicatedMapProxy) AddEntryListener(listener interface{}) (registrationID *string, err error) {
+	request := ReplicatedMapAddEntryListenerEncodeRequest(rmp.name, rmp.isSmart())
+	eventHandler := func(clientMessage *ClientMessage) {
```
mustafaiman: can you extract event handler creation out of this method?

internal/replicated_map.go
```diff
@@ -0,0 +1,256 @@
...
+	})
+}
+
+func (rmp *ReplicatedMapProxy) AddEntryListenerWithPredicate(listener interface{}, predicate IPredicate) (registrationID *string, err error) {
+	predicateData, err := rmp.validateAndSerializePredicate(predicate)
+	if err != nil {
+		return nil, err
+	}
+	request := ReplicatedMapAddEntryListenerWithPredicateEncodeRequest(rmp.name, predicateData, rmp.isSmart())
+	eventHandler := func(clientMessage *ClientMessage) {
```
mustafaiman: extract event handler creation out of this method

tests/proxy/map/map_test.go
```diff
@@ -42,14 +43,24 @@ func TestMain(m *testing.M) {
...
 	m.Run()
-	mp.Clear()
+	clear()
 	client.Shutdown()
 	remoteController.ShutdownCluster(cluster.ID)
 }
 
+func initialize() {
+	mp, _ = client.GetMap("myMap")
+	rmp, _ = client.GetReplicatedMap("myReplicatedMap")
```
mustafaiman: `rmp` is not used anywhere in this test. Should be removed.

tests/proxy/replicated_map/replicated_map_test.go
```diff
@@ -0,0 +1,345 @@
...
+	client, _ = hazelcast.NewHazelcastClient()
+	rmp, _ = client.GetReplicatedMap("myReplicatedMap")
+	m.Run()
+	rmp.Clear()
+	client.Shutdown()
+	remoteController.ShutdownCluster(cluster.ID)
+}
+
+func TestReplicatedMapProxy_Name(t *testing.T) {
+	name := "myReplicatedMap"
```
mustafaiman: no need for this variable. In fact, eliminating the variable would make the test more readable.

tests/proxy/replicated_map/replicated_map_test.go
```diff
@@ -0,0 +1,345 @@
...
+	AssertEqualf(t, nil, name, rmp.Name(), "Name() failed")
+}
+
+func TestReplicatedMapProxy_ServiceName(t *testing.T) {
+	serviceName := common.SERVICE_NAME_REPLICATED_MAP
```
mustafaiman: no need for `serviceName` variable. Use the name directly in the assertion.

tests/proxy/replicated_map/replicated_map_test.go
```diff
@@ -0,0 +1,345 @@
...
+	AssertEqualf(t, nil, serviceName, rmp.ServiceName(), "ServiceName() failed")
+}
+
+func TestReplicatedMapProxy_PartitionKey(t *testing.T) {
+	name := "myReplicatedMap"
```
mustafaiman: again no need for the variable.

tests/proxy/replicated_map/replicated_map_test.go
```diff
@@ -0,0 +1,345 @@
...
+	defer rmp.Clear()
+	testMap := make(map[interface{}]interface{})
+	for i := 0; i < 10; i++ {
+		testMap["testingKey"+strconv.Itoa(i)] = "testingValue" + strconv.Itoa(i)
+	}
+	err := rmp.PutAll(testMap)
+	AssertErrorNil(t, err)
+	entryList, err := rmp.EntrySet()
+	AssertErrorNil(t, err)
+	for _, pair := range entryList {
```
mustafaiman: if `entryList` did not include all the required elements, this test would still pass.
It will be better if you write a utility method that will check equality of two slices and use that. Another utility method may also be necessary for getting the slice from map.

tests/proxy/replicated_map/replicated_map_test.go
```diff
@@ -0,0 +1,345 @@
...
+func TestReplicatedMapProxy_Values(t *testing.T) {
+	defer rmp.Clear()
+	var expecteds = make([]string, 10)
+	var ret = make([]string, 10)
+	for i := 0; i < 10; i++ {
+		rmp.Put(strconv.Itoa(i), strconv.Itoa(i))
+		expecteds[i] = strconv.Itoa(i)
+	}
+	values, err := rmp.Values()
+	for j := 0; j < 10; j++ {
```
mustafaiman: what if `values` was broken so that it returned more elements than 10. This test would still pass. It is better if you write a utility method for this kind of comparisons and use it in these kind of tests. Something like `AssertIterableEqual`.

tests/proxy/replicated_map/replicated_map_test.go
```diff
@@ -0,0 +1,345 @@
...
+func TestReplicatedMapProxy_AddEntryListenerWithPredicate(t *testing.T) {
+	var wg *sync.WaitGroup = new(sync.WaitGroup)
+	entryListener := &entryListener{wg: wg}
+	registrationId, err := rmp.AddEntryListenerWithPredicate(entryListener, core.Equal("this", "value"))
+	AssertEqual(t, err, nil, nil)
+	wg.Add(1)
+	rmp.Put("key123", "value")
+	timeout := WaitTimeout(wg, Timeout)
+	AssertEqualf(t, nil, false, timeout, "AddEntryListenerWithPredicate failed")
+	rmp.RemoveEntryListener(registrationId)
```
mustafaiman: you dont assert the values that entry listener got here.

tests/proxy/replicated_map/replicated_map_test.go
```diff
@@ -0,0 +1,345 @@
...
+
+func TestReplicatedMapProxy_AddEntryListenerToKey(t *testing.T) {
+	var wg *sync.WaitGroup = new(sync.WaitGroup)
+	entryListener := &entryListener{wg: wg}
+	registrationId, err := rmp.AddEntryListenerToKey(entryListener, "key1")
+	AssertEqual(t, err, nil, nil)
+	wg.Add(1)
+	rmp.Put("key1", "value1")
+	timeout := WaitTimeout(wg, Timeout)
+	AssertEqualf(t, nil, false, timeout, "AddEntryListenerToKey failed")
```
mustafaiman: the same as `AddEntryListenerWithPredicate` above.

tests/proxy/replicated_map/replicated_map_test.go
SaitTalhaNisanci: @mustafaiman  I think these tests are not necessary actually.(Name,ServiceName,PartitionKey). They are already tested here https://github.com/hazelcast/hazelcast-go-client/blob/master/tests/proxy/map/map_test.go#L53. Do you think they are necessary ?

tests/proxy/map/map_test.go
```diff
@@ -32,6 +32,7 @@ import (
 
 var mp IMap
 var mp2 IMap
+var rmp ReplicatedMap
```
mustafaiman: not used

tests/proxy/map/map_test.go
```diff
@@ -42,14 +43,23 @@ func TestMain(m *testing.M) {
...
 	remoteController.ShutdownCluster(cluster.ID)
 }
 
+func initialize() {
+	mp, _ = client.GetMap("myMap")
+}
+
+func clear() {
+	mp.Clear()
+	rmp.Clear()
```
mustafaiman: not used

tests/proxy/replicated_map/replicated_map_test.go
mustafaiman: Agree, not really neccessary.

tests/proxy/replicated_map/replicated_map_test.go
```diff
@@ -0,0 +1,323 @@
...
+	testKey := "testingKey"
+	testValue := "testingValue"
+	rmp.Put(testKey, testValue)
+	rmp.Destroy()
+	rmp, _ := client.GetReplicatedMap("myReplicatedMap")
+	res, err := rmp.Get(testKey)
+	AssertNilf(t, err, res, "replicatedMap Destroy() failed")
+}
+
+func TestReplicatedMapProxy_Put(t *testing.T) {
```
asimarslan: There are missing tests like TestX_withNil tests

SaitTalhaNisanci: @asimarslan  I dont think we need those tests here, they go to the same method which is already tested.

asimarslan: Those validation methods are tested but your proxy method may not use it correctly so We need to test if a proxy method reject a nil value or not. So we do need these tests.

tests/proxy/replicated_map/replicated_map_test.go
```diff
@@ -0,0 +1,323 @@
...
+func TestReplicatedMapProxy_Get(t *testing.T) {
+	defer rmp.Clear()
+	testKey := "testingKey"
+	testValue := "testingValue"
+	_, err := rmp.Put(testKey, testValue)
+	res, err := rmp.Get(testKey)
+	AssertEqualf(t, err, res, testValue, "replicatedMap Get() failed")
+}
+
+func TestReplicatedMapProxy_ContainsKey(t *testing.T) {
```
asimarslan: We also need test with different values. The methods that take values should be tested for both exist and not exists  cases.
Test Contain for an empty map and/or Remove a non exists key.
See queue_test.go

PR #186, opened 2018-03-08
log member list after a change in the list

### LINKED ISSUE #183: Member List
We should print the member list when the client connects to server.

SaitTalhaNisanci:
fixes #183.

The format is as follows:
```
Members {size:1} [
	 Member [127.0.0.1]:5701 - 4514d75d-513d-43c2-b5ae-a40e30d7a55c
]
```

### DIFF
--- a/internal/cluster.go
+++ b/internal/cluster.go
@@ -15,6 +15,7 @@
 package internal
 
 import (
+	"fmt"
 	"github.com/hazelcast/hazelcast-go-client/config"
 	"github.com/hazelcast/hazelcast-go-client/core"
 	"github.com/hazelcast/hazelcast-go-client/internal/common"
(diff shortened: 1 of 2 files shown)


asimarslan:
@SaitTalhaNisanci Log format is changed but not updated on other clients, Can you update this PR to comply with Java client log.

internal/cluster.go
```diff
@@ -181,9 +182,22 @@ func (clusterService *ClusterService) initMembershipListener(connection *Connect
...
+	clusterService.logMembers()
 	log.Println("Registered membership listener with Id ", *registrationId)
 	return nil
 }
+
+func (clusterService *ClusterService) logMembers() {
+	members := clusterService.members.Load().([]*Member)
+	membersInfo := fmt.Sprintf("New member list:\n\nMembers[%d]{\n", len(members))
+	for _, member := range members {
+		memberInfo := fmt.Sprintf("\t Member [%s:%d]\n", member.Address().Host(), member.Address().Port())
```
asimarslan: A member string should be same as : https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/instance/AbstractMember.java#L181

internal/cluster.go
```diff
@@ -181,9 +182,28 @@ func (clusterService *ClusterService) initMembershipListener(connection *Connect
...
+	clusterService.logMembers()
 	log.Println("Registered membership listener with Id ", *registrationId)
 	return nil
 }
+
+func (clusterService *ClusterService) logMembers() {
+	members := clusterService.members.Load().([]*Member)
+	membersInfo := fmt.Sprintf("\n\nMembers {size:%d} [\n", len(members))
+	for _, member := range members {
+		memberInfo := fmt.Sprintf("\t Member [%s]:%d - %s", member.Address().Host(), member.Address().Port(),
```
asimarslan: What about creating a `Member.String()`so that it can be reused in future?

PR #187, opened 2018-03-09
Ringbuffer codecs, implementation and test

furkansenharputlu:


internal/ringbuffer.go
```diff
@@ -0,0 +1,119 @@
...
+	itemData, err := rp.validateAndSerialize(item)
+	if err != nil {
+		return
+	}
+	request := RingbufferAddEncodeRequest(rp.name, int32(overflowPolicy.Policy()), itemData)
+	responseMessage, err := rp.Invoke(request)
+	return rp.DecodeToInt64AndError(responseMessage, err, RingbufferAddDecodeResponse)
+}
+
+func (rp *RingbufferProxy) AddAll(items []interface{}, overflowPolicy core.OverflowPolicy) (lastSequence int64, err error) {
```
mustafaiman: overflowPolicy is hardcoded as `Overwrite` in sync AddAll.

internal/ringbuffer.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	readCount, itemsData, itemSeqs := RingbufferReadManyDecodeResponse(responseMessage)()
+	items, err = collection.DataToObjectCollection(itemsData, rp.client.SerializationService)
+	if err != nil {
+		return
+	}
+	return items, readCount, itemSeqs, nil
+}
+
+func (rp *RingbufferProxy) checkSequence(value int64, argName string) (err error) {
```
mustafaiman: `AssertNotNegative` would be more informative for this method's name.

internal/ringbuffer.go
```diff
@@ -0,0 +1,119 @@
...
+func (rp *RingbufferProxy) ReadOne(sequence int64) (item interface{}, err error) {
+	if err = rp.checkSequence(sequence, "sequence"); err != nil {
+		return
+	}
+	request := RingbufferReadOneEncodeRequest(rp.name, sequence)
+	responseMessage, err := rp.Invoke(request)
+	return rp.DecodeToObjectAndError(responseMessage, err, RingbufferReadOneDecodeResponse)
+}
+
+func (rp *RingbufferProxy) ReadMany(startSequence int64, minCount int32, maxCount int32, filter interface{}) (items []interface{}, readCount int32, itemSequences []int64, err error) {
```
mustafaiman: There is no sync api for ReadMany in other clients.

internal/ringbuffer.go
```diff
@@ -0,0 +1,119 @@
+package internal
```
SaitTalhaNisanci: Can you add the licence here.
You can add licence to default file template in your IDE.

tests/proxy/ringbuffer/ringbuffer_test.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func TestRingbufferProxy_Capacity(t *testing.T) {
+	defer destroyAndCreate()
+	fillRingbuffer(capacity)
+	cap, err := ringbuffer.Capacity()
+	AssertEqualf(t, err, cap, capacity, "ringbuffer Capacity() failed")
+}
+
+func TestRingbufferProxy_AddSize(t *testing.T) {
```
mustafaiman: this is not a good test. checking multiple things at once. I would split size and add tests.

mustafaiman:
I see this PR implements a synchronous `ReadMany` method although other clients have only asynchronous version of it. Is this intentional?

internal/ringbuffer.go
```diff
@@ -0,0 +1,119 @@
...
+func newRingbufferProxy(client *HazelcastClient, serviceName *string, name *string) (*RingbufferProxy, error) {
+	parSpecProxy, err := newPartitionSpecificProxy(client, serviceName, name)
+	if err != nil {
+		return nil, err
+	}
+	return &RingbufferProxy{parSpecProxy}, nil
+}
+
+func (rp *RingbufferProxy) Capacity() (capacity int64, err error) {
+	request := RingbufferCapacityEncodeRequest(rp.name)
```
SaitTalhaNisanci: In java it looks like there is a different implementation 
https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java#L107

internal/ringbuffer.go
```diff
@@ -0,0 +1,119 @@
...
+	return rp.DecodeToInt64AndError(responseMessage, err, RingbufferHeadSequenceDecodeResponse)
+}
+
+func (rp *RingbufferProxy) RemainingCapacity() (remainingCapacity int64, err error) {
+	request := RingbufferRemainingCapacityEncodeRequest(rp.name)
+	responseMessage, err := rp.Invoke(request)
+	return rp.DecodeToInt64AndError(responseMessage, err, RingbufferRemainingCapacityDecodeResponse)
+}
+
+func (rp *RingbufferProxy) Add(item interface{}, overflowPolicy core.OverflowPolicy) (sequence int64, err error) {
```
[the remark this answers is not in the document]
furkansenharputlu: Yes, we considered that but we decided to make it like this.

internal/ringbuffer.go
```diff
@@ -0,0 +1,119 @@
...
+
+import (
+	"fmt"
+	"github.com/hazelcast/hazelcast-go-client/core"
+	"github.com/hazelcast/hazelcast-go-client/internal/common/collection"
+	. "github.com/hazelcast/hazelcast-go-client/internal/protocol"
+)
+
+type RingbufferProxy struct {
+	*partitionSpecificProxy
```
SaitTalhaNisanci: In java this is not `partitionSpecificProxy`, as far as I see in some clients this is `partititionSpecificProxy` and in some it is not, what should this be?

furkansenharputlu: I took Node.js as reference.

internal/ringbuffer.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return &RingbufferProxy{parSpecProxy}, nil
+}
+
+func (rp *RingbufferProxy) Capacity() (capacity int64, err error) {
+	request := RingbufferCapacityEncodeRequest(rp.name)
+	responseMessage, err := rp.Invoke(request)
+	return rp.DecodeToInt64AndError(responseMessage, err, RingbufferCapacityDecodeResponse)
+}
+func (rp *RingbufferProxy) Size() (size int64, err error) {
```
SaitTalhaNisanci: It looks like in map codecs `Size()` returns `int32` however in this one it is `int64`. In java it is also like that, is there a reason for that ?

internal/ringbuffer.go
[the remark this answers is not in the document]
asimarslan: There is no method override in golang so it's not nice to have a seperate named method. A single one already does the job.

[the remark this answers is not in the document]
mustafaiman: I assumed there is going to be an AddAsync method in the future but there is no need for that. So ignore my comment.

internal/ringbuffer.go
asimarslan: go is more like python. You can see each method both sync and async.

internal/ringbuffer.go
mustafaiman: This is from Javadoc:
```
 * A Ringbuffer currently is a replicated but not partitioned data-structure. So all data is stored in a single partition;
 * comparable to the IQueue implementation. But we'll provide an option to partition the data in the near future.
```
In practice, it is partition specific. If it becomes partitioned in the future, we can change this easily. It is implementation detail anyway.

internal/ringbuffer.go
asimarslan: It caches the capacity value so that client won't ask it server again to improve performance.

internal/ringbuffer.go
asimarslan: Yes you are right there is a inconsistency but it is out of scope of this PR.

furkansenharputlu:
verify

PR #189, opened 2018-03-12
fix Ringbuffer readMany to make it more backward compatible API

asimarslan:


internal/ringbuffer.go
```diff
@@ -139,3 +135,35 @@ func (rp *RingbufferProxy) checkCounts(minCount int32, maxCount int32) (err erro
...
+func NewLazyReadResultSet(readCount int32, items []*Data, itemSeqs []int64, ss *SerializationService) *LazyReadResultSet {
+	return &LazyReadResultSet{readCount: readCount, dataItems: items, itemSequences: itemSeqs, serializationService: ss}
+}
+
+func (rs *LazyReadResultSet) ReadCount() int32 {
+	return rs.readCount
+}
+
+func (rs *LazyReadResultSet) Get(index int32) (result interface{}, err error) {
+	return rs.serializationService.ToObject(rs.dataItems[index])
```
SaitTalhaNisanci: What if the index is out of boundaries ?

SaitTalhaNisanci: Also why dont we store the object after converting the corresponding data so that when the same index is called more than once we can directly return the result. Is there any reason to avoid doing that ?

internal/ringbuffer.go
```diff
@@ -139,3 +135,35 @@ func (rp *RingbufferProxy) checkCounts(minCount int32, maxCount int32) (err erro
...
+}
+
+func (rs *LazyReadResultSet) Sequence(index int32) int64 {
+	if rs.itemSequences == nil {
+		return SEQUENCE_UNAVAILABLE
+	}
+	return rs.itemSequences[index]
```
SaitTalhaNisanci: Again shouldnt we check the index range?

internal/ringbuffer.go
```diff
@@ -139,3 +135,35 @@ func (rp *RingbufferProxy) checkCounts(minCount int32, maxCount int32) (err erro
...
+
+func (rs *LazyReadResultSet) ReadCount() int32 {
+	return rs.readCount
+}
+
+func (rs *LazyReadResultSet) Get(index int32) (result interface{}, err error) {
+	return rs.serializationService.ToObject(rs.dataItems[index])
+}
+
+func (rs *LazyReadResultSet) Sequence(index int32) int64 {
```
SaitTalhaNisanci: In java this method throws `UnsupportedOperationException` when `itemSequences`is nil. Would it be better to return an error too in this method ?

internal/ringbuffer.go
asimarslan: we need to add index validation

core/ringbuffer.go
```diff
@@ -121,3 +121,23 @@ type policy int32
 func (op policy) Policy() policy {
 	return op
 }
+
+// The result of a Ringbuffer.ReadMany operation
```
SaitTalhaNisanci: In go the convention for docs is to start with what you are documenting, for example here it should be `ReadResultSet is the result of a .... ` and in methods it is the same `ReadCount returns the... ` etc

internal/ringbuffer.go
SaitTalhaNisanci: As far as I understand it is like this in Java too,
```
    public E get(int index) {
        rangeCheck(index);
        final Data item = items[index];
        return serializationService.toObject(item);
    }
```
why is it like that ?

SaitTalhaNisanci: I think we need index validation in the proxies as well but server will return an error in that case but here it will panic so this is more vital I guess.

internal/ringbuffer.go
asimarslan: Yes but in java you don't return a value when you rise an exception but in go you add an error parameter and return the value anyway so there is no point for an extra error parameter

internal/ringbuffer.go
```diff
@@ -139,3 +135,35 @@ func (rp *RingbufferProxy) checkCounts(minCount int32, maxCount int32) (err erro
...
 }
+
+type LazyReadResultSet struct {
+	readCount            int32
+	dataItems            []*Data
+	itemSequences        []int64
+	serializationService *SerializationService
+}
+
+const SEQUENCE_UNAVAILABLE int64 = -1
```
furkansenharputlu: `sequenceUnavailable` is convenient for Go.

internal/ringbuffer.go
```diff
@@ -139,3 +135,66 @@ func (rp *RingbufferProxy) checkCounts(minCount int32, maxCount int32) (err erro
...
+	items                []interface{}
+	itemSequences        []int64
+	serializationService *SerializationService
+}
+
+const sequenceUnavailable int64 = -1
+
+func NewLazyReadResultSet(readCount int32, itemsData []*Data, itemSeqs []int64, ss *SerializationService) (rs *LazyReadResultSet) {
+	rs = &LazyReadResultSet{readCount: readCount, itemSequences: itemSeqs, serializationService: ss}
+	rs.items = rs.convertToInterfaceSlice(itemsData)
```
SaitTalhaNisanci: In the corresponding codec if we created `[]interface{}` instead of `[]*Data` we wouldnt need this.  We iterate the same slice twice because of that. 

Doing that for only that codec is not trivial unless we do it manually. 
So I am not sure which is better.

asimarslan: The codecs are designed to be code generated, in java you can not even edit them. Let's keep it as `[]*Data`

PR #192, opened 2018-03-23
Org website samples

furkansenharputlu:


samples/org_website_samples/queue_sample.go
```diff
@@ -35,8 +35,7 @@ func queueSampleRun() {
 	queue.PollWithTimeout(5, time.Second)
 	//Indefinitely blocking Operations
 	queue.Put("yetanotheritem")
-	item, _ := queue.Take()
-	fmt.Println(item)
+	fmt.Println(queue.Take())
```
SaitTalhaNisanci: This is also going to print the error, if there isnt error it will print `<nil>`

asimarslan: I think that's Ok

PR #197, opened 2018-03-24
PNCounter interface with GoDocs

furkansenharputlu:
This pr adds PNCounter interface with GoDocs. #193


furkansenharputlu:
verify

PR #198, opened 2018-03-26
add PNCounter codecs

furkansenharputlu:
proxy codecs #193


internal/protocol/consts.go
```diff
@@ -24,9 +24,15 @@ const CLIENT_TYPE = "GOO"
 func DataCalculateSize(d *Data) int {
 	return len(d.Buffer()) + INT_SIZE_IN_BYTES
 }
+
 func StringCalculateSize(str *string) int {
 	return len(*str) + INT_SIZE_IN_BYTES
 }
+
+func Int64CalculateSize(v int64) int {
```
furkansenharputlu: This is to not change codecs

PR #201, opened 2018-03-26
implement flakeIdGenerator with tests

SaitTalhaNisanci:
This pr implements flakeIdGenerator proxy with tests.
In config, pattern matcher is not implemented and left for another PR for future.
This pr also changes the server version to `3.10-BETA-1` in `start-rc.sh` since the server should be at least *3.10* for `flakeIdGenerator`
#194.


tests/proxy/flake_id_generator/flake_id_generator_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	wg.Wait()
+	for _, localIds := range localIdsSlice {
+		for id, _ := range localIds {
+			ids[id] = struct{}{}
+		}
+	}
+	// if there were duplicate IDs generated, there will be less items in the map than expected
+	AssertEqualf(t, nil, len(ids), numRoutines*idsInRoutine, "FlakeIdGenerator NewId() returned duplicate ids")
+}
```
mustafaiman: Add a test to ensure that the generator returns the appropriate error when there are no members in the cluster that has a join id smaller than 2^16

tests/proxy/flake_id_generator/flake_id_generator_test.go
```diff
@@ -37,18 +38,30 @@ const (
...
+	script := "function assignOverflowedNodeId() {" +
+		"   instance_" + strconv.Itoa(instanceNum) + ".getCluster().getLocalMember().setMemberListJoinVersion(100000);" +
+		"   return instance_" + strconv.Itoa(instanceNum) + ".getCluster().getLocalMember().getMemberListJoinVersion();" +
+		"}" +
+		"result=\"\"+assignOverflowedNodeId();"
+	return remoteController.ExecuteOnController(clusterId, script, 1)
 }
 
 func TestFlakeIdGeneratorProxy_ConfigTest(t *testing.T) {
+	cluster, err := remoteController.CreateCluster("3.9", DefaultServerConfig)
```
mustafaiman: Does this mean cluster version is going to be 3.9? 3.9 does not even support flake ids.

SaitTalhaNisanci:
verify

Issue #205, opened 2018-03-29
Projections & Aggregations


SaitTalhaNisanci:
- [X] Backport the related Hazelcast Java interface into go repo.
- [X] Document the interface GoDocs.
- [X] Generate the related proxy codecs of client protocol
- [X] Implement proxy and write unit tests
- [] Write a sample for this proxy

PR #206, opened 2018-03-29
add aggregate and project apis to map

SaitTalhaNisanci:
This pr only adds aggregator and projection methods to `IMap`.
#205.

### DIFF
--- a/core/map.go
+++ b/core/map.go
@@ -70,6 +70,26 @@ type IMap interface {
 	// Size returns the number of entries in this map.
 	Size() (size int32, err error)
 
+	// Aggregate applies the aggregation logic on all map entries and returns the result.
+	// Fast-Aggregations are the successor of the Map-Reduce Aggregators.
+	// They are equivalent to the Map-Reduce Aggregators in most of the use-cases, but instead of running on the Map-Reduce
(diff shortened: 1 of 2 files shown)


sancar:
IAggregator and IProjection does not have to be IdentifiedDataSerializable. It can be Portable or Custom serializable also. Bad news is that IPredicate is wrong also. It is same. 
The methods accepting these should probably be `interface{}`

core/map.go
```diff
@@ -71,6 +71,23 @@ type IMap interface {
 	// Size returns the number of entries in this map.
 	Size() (size int32, err error)
 
+	// Aggregate applies the aggregation logic on all map entries and returns the result.
+	// Fast-Aggregations are the successor of the Map-Reduce Aggregators.
+	// They are equivalent to the Map-Reduce Aggregators in most of the use-cases, but instead of running on the Map-Reduce
+	// engine they run on the Query infrastructure. Their performance is tens to hundreds times better due to the fact
+	// that they run in parallel for each partition and are highly optimized for speed and low memory consumption.
+	// Aggregate returns the result of the given aggregator.
+	Aggregate(aggregator IAggregator) (result interface{}, err error)
```
furkansenharputlu: I think `I` is not necessary at the beginning of `IAggregator `.

PR #207, opened 2018-03-30
implement PNCounter and test

furkansenharputlu:
#193 

This pr also solves some data races in internal.


furkansenharputlu:
verify

furkansenharputlu:
verify

furkansenharputlu:
verify

internal/connection_manager.go
```diff
@@ -189,7 +187,9 @@ func (connectionManager *connectionManager) clusterAuthenticator(connection *Con
 			connection.serverHazelcastVersion = serverHazelcastVersion
 			connection.endpoint.Store(address)
 			connection.isOwnerConnection = asOwner
+			connectionManager.lock.Lock()
```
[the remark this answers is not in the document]
sancar: Client should not send authentication to same address twice. We should avoid doing that in the first place.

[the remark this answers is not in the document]
sancar: It think we should merge this part as it is done in this prd. 
It seems we need to fix more here and it is not really job of this prd. If tests are passing with this change I am OK. Lets discuss the solution and prepare a second prd for this.

internal/invocation.go
```diff
@@ -223,14 +228,14 @@ func (invocationService *invocationService) send(invocation *invocation, connect
...
-			log.Println("the following error occured while trying to send the invocation ", err)
+			log.Println("the following error occured while trying to send the invocation: ", err)
 			invocationService.handleException(invocation, err)
 		}
 	}()
 }
 func (invocationService *invocationService) sendInvocation(invocation *invocation) invocationResult {
 	if invocationService.isShutdown.Load().(bool) {
-		invocation.err <- NewHazelcastClientNotActiveError("Client is shut down", nil)
+		invocation.err <- NewHazelcastClientNotActiveError("client is shut down", nil)
```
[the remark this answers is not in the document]
furkansenharputlu: In Golang, error strings should start with lower case and should not end with punctuation.
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

internal/pn_counter.go
```diff
@@ -0,0 +1,253 @@
...
+	response, err := c.invokeGetInternal(c.emptyAddresses, nil, target)
+	if err != nil {
+		return
+	}
+	value, replicaTimestamps, _ := PNCounterGetDecodeResponse(response)()
+	c.updateObservedReplicaTimestamps(replicaTimestamps)
+	return value, nil
+}
+
+func (c *PNCounterProxy) operation(delta int64, getBeforeUpdate bool) (value int64, err error) {
```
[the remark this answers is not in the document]
furkansenharputlu: What can be better than `operation` ? Do you have any suggestion ?

internal/pn_counter.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+	return response, nil
+}
+
+func (c *PNCounterProxy) chooseTargetReplica(excludedAddresses map[IAddress]struct{}) (IAddress, error) {
+	replicaAddresses, err := c.getReplicaAddresses(excludedAddresses)
+	if err != nil || len(replicaAddresses) == 0 {
+		return nil, err
+	}
+	rand.Seed(time.Now().UnixNano())
```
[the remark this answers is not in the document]
furkansenharputlu: If seed is called once, `rand.Intn` returns always same result so I preferred this way to be able to get different random values.

internal/connection_manager.go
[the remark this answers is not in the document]
mustafaiman: But these locks were not introduced by this PRD. There was already a `connect` mechanism and this PR changes it. I do not know if the old mechanism had problems but I think this change introduces more bugs than it solves.

internal/pn_counter.go
```diff
@@ -0,0 +1,253 @@
...
+	return c, nil
+}
+
+func (c *PNCounterProxy) Get() (currentValue int64, err error) {
+	target, err := c.getCRDTOperationTarget(c.emptyAddresses)
+	if err != nil {
+		return
+	}
+	if target == nil {
+		err = NewHazelcastNoDataMemberInClusterError("cannot invoke operations on a CRDT because the cluster does not contain any data members", nil)
```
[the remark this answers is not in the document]
mustafaiman: ignore this

internal/pn_counter.go
```diff
@@ -0,0 +1,253 @@
...
+	return c.operation(1, true)
+}
+
+func (c *PNCounterProxy) Reset() {
+	atomic.StorePointer(&c.observedClock, unsafe.Pointer(NewVectorClock()))
+}
+
+func (c *PNCounterProxy) getCRDTOperationTarget(excludedAddresses map[IAddress]struct{}) (IAddress, error) {
+	c.targetSelectionMutex.RLock()
+	defer c.targetSelectionMutex.RUnlock()
```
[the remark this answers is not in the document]
mustafaiman: You can just store whatever is read into a local variable before releasing the lock. Then release the lock and return the stored value. `defer` is not unavoidable.

internal/pn_counter.go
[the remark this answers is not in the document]
mustafaiman: what about `getAndUpdateTimestamps`?

internal/pn_counter.go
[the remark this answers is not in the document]
mustafaiman: I think it should not be that way. When I run https://play.golang.org/p/Vsoo4PBTlVe , it gives another random integer at each line.

tests/proxy/pn_counter/pn_counter_test.go
```diff
@@ -0,0 +1,210 @@
...
+	wg.Wait()
+	currentValue, err := counter.Get()
+	AssertEqualf(t, err, currentValue, int64(delta), "PNCounter has race condition")
+
+}
+
+func TestPNCounter_HazelcastNoDataMemberInClusterError(t *testing.T) {
+	client.Shutdown()
+	remoteController.ShutdownCluster(cluster.ID)
+	cluster, err = remoteController.CreateCluster("3.10", LiteMemberConfig)
```
mustafaiman: As far as I know, supplying a version number to `CreateCluster` does not have an effect in terms of server version number. It implies an effect that is actually not there. Moreover, these tests are going to run with later versions too. Maybe we should let it be `nil`.

tests/util.go
```diff
@@ -82,6 +82,25 @@ const DefaultServerConfig = `
 </hazelcast>
 `
 
+const LiteMemberConfig = `
```
mustafaiman: I do not like the idea of having xml configs inlined like this. I think required xml configs should be in their respective `.xml` file. One reason being that you lose syntax highlighting in xml which makes it harder to read. Also, these are not really `util` kind of things that are some fundamental helpers used by multiple tests. Rather, each one is expected to be used by a single test most of the time.

On the other hand, this way it is easier to find which tests use a specific configuration.

I would say put here a helper function that will read the content of a given xml file into a string. Keep actual xml files close to the tests that use them.

tests/proxy/pn_counter/pn_counter_test.go
```diff
@@ -0,0 +1,210 @@
...
+	currentValue, err := counter.Get()
+	AssertEqualf(t, err, currentValue, delta, "PNCounter.Get failed")
+}
+
+func TestPNCounter_GetAndAdd(t *testing.T) {
+	defer destroyAndCreate()
+	var delta int64 = 5
+	previousValue, err := counter.GetAndAdd(delta)
+	AssertEqualf(t, err, previousValue, int64(0), "PNCounter.GetAndAdd failed")
+}
```
mustafaiman: this test does not check if the counter value actually increased by 5. Other tests that return previous value have the same problem, too.

tests/proxy/pn_counter/pn_counter_test.go
```diff
@@ -0,0 +1,210 @@
...
+func TestPNCounter_HazelcastNoDataMemberInClusterError(t *testing.T) {
+	client.Shutdown()
+	remoteController.ShutdownCluster(cluster.ID)
+	cluster, err = remoteController.CreateCluster("3.10", LiteMemberConfig)
+	remoteController.StartMember(cluster.ID)
+	client, _ = hazelcast.NewHazelcastClient()
+	counter, _ = client.GetPNCounter(counterName)
+	var delta int64 = 5
+	_, err = counter.AddAndGet(delta)
+	if _, ok := err.(*core.HazelcastNoDataMemberInClusterError); !ok {
```
mustafaiman: Can you add a `assertExpectError(t *Testing, err Error, msg string)` assertion to `util` and use that for this kind of checks.

tests/proxy/pn_counter/pn_counter_test.go
```diff
@@ -0,0 +1,210 @@
...
+	client, _ = hazelcast.NewHazelcastClient()
+	counter, _ = client.GetPNCounter(counterName)
+	var delta int64 = 5
+	counter.GetAndAdd(delta)
+	target := client.GetCluster().GetMember(counter.(*internal.PNCounterProxy).GetCurrentTargetReplicaAddress())
+	remoteController.TerminateMember(cluster.ID, target.Uuid())
+	counter.Reset()
+	currentValue, err := counter.AddAndGet(delta)
+	AssertEqualf(t, err, currentValue, int64(delta), "PNCounter.Reset failed")
+}
```
mustafaiman: I think the following tests should be in their own file
- `TestPNCounter_HazelcastNoDataMemberInClusterError`
- `TestPNCounter_HazelcastConsistencyLostError`
- `TestPNCounter_Reset`

The problem is that these tests mess with global state(client instance). If these tests fail, they will not be able to restore the global state. So the rest of the tests are going to fail. Tests should be isolated as much as possible.

internal/vector_clock.go
```diff
@@ -0,0 +1,120 @@
...
+	for i := int32(0); i < stateSize; i++ {
+		replicaId, err := input.ReadUTF()
+		if err != nil {
+			return err
+		}
+		timestamp, err := input.ReadInt64()
+		if err != nil {
+			return err
+		}
+		v.mutex.Lock()
```
mustafaiman: do we need the lock here, given that `ReadData` is going to be called only once for each `VectorClock` instance?

internal/vector_clock.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	containsAll := true
+	v.mutex.RLock()
+	for replicaId, _ := range v.replicaTimestamps {
+		if _, containsAll = other.replicaTimestamps[replicaId]; !containsAll {
+			break
+		}
+	}
+	v.mutex.RUnlock()
+	return anyTimestampGreater || containsAll
```
mustafaiman: this check is not true. if two vector clocks are exactly the same, this will return true although it needs to return false. 

You actually do not need the second loop. At this point, it is enough to check if `v` has more entries than `other` because you already know that `v` has all the entries that `other` has by the first loop.

internal/vector_clock.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (v *VectorClock) SetReplicaTimestamp(replicaId *string, timestamp int64) {
+	v.mutex.Lock()
+	v.replicaTimestamps[replicaId] = timestamp
+	v.mutex.Unlock()
+}
+
+func (v *VectorClock) TimestampForReplica(replicaId *string) int64 {
```
mustafaiman: this method does not tell if there is indeed a timestamp for `replicaId`. I suggest that this method returns two results `val, ok`.
I do not remember if a timestamp can be negative or 0 but it is better to be on the safe side.

internal/pn_counter.go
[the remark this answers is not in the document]
furkansenharputlu: Why do we try to avoid `defer`? Also, I think storing to a local variable is reverse to logic of the concurrency. After releasing lock, what if other goroutine writes and changes the value. We won't be able to get the new value of shared variable.

internal/pn_counter.go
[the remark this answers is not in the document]
furkansenharputlu: The thing confusing me is that when you run this code again, it prints same results.

internal/vector_clock.go
furkansenharputlu: This `v.replicaTimestamps[replicaId]` already returns `val, ok`.
Like:
`val,ok :=v.replicaTimestamps[replicaId]`

tests/proxy/pn_counter/pn_counter_test.go
furkansenharputlu: In Go, tests in same package run as if they are in the same test file. To make this I need to create 3 packages for 3 different tests so I preferred this way.

tests/proxy/pn_counter/pn_counter_test.go
furkansenharputlu: It should be like this:
 `AssertExpectedError(t, err, reflect.TypeOf((*core.HazelcastNoDataMemberInClusterError)(nil)).Elem(),
		"PNCounter.AddAndGet should return HazelcastNoDataMemberInClusterError")`

I think using an extra method does not make it easy or nice-looking.

tests/proxy/pn_counter/pn_counter_test.go
furkansenharputlu: `string` can't be `nil` but I will make it `""`.

internal/pn_counter.go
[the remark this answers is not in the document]
mustafaiman: Because it is an online playground, probably `time.Now()` call returns a cached value within specific intervals. So when you run the same sample within a short amount of time, the seed stays the same. It is expected to get the same set of random numbers then.

tests/proxy/pn_counter/pn_counter_test.go
mustafaiman: Can you pass `*core.HazelcastNoDataMemberInClusterError` directly instead of `reflect.TypeOf((*core.HazelcastNoDataMemberInClusterError)(nil)).Elem()`? Since `reflect.TypeOf` can take it as an argument, I am tempted to think that you can pass that type to another function and do the refecltion within you function. But I am not familiar with reflection in Golang so I will take your word for it.

I can agree that it does not look nicer that way but I still see some value in it.
- Having the assert method clearly shows that it is the intention of this test to ensure that the error is of that type.
- Less error prone to write since you delegate `if` branching here. 
- If you are fine with a more general error message such as `Expected $x error, but got $y`, you can avoid writing error description every time. I think general error message is good because all the information about the error is still easily reachable. Specific error message does not provide extra value.
- Easier to read.

tests/proxy/pn_counter/pn_counter_test.go
mustafaiman: Do you know how such situation is handled in other Go libraries?

internal/pn_counter.go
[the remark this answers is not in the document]
mustafaiman: `operation` does not give me any clue about what this function does. When I am reading another function that uses `operation`, I find myself jumping to `operation` code eveytime to remember what it does.

tests/proxy/pn_counter/pn_counter_test.go
furkansenharputlu:
````
func TestUnreadRuneAtEOF(t *testing.T) {
	// UnreadRune/ReadRune should error at EOF (was a bug; used to panic)
	r := NewReader(strings.NewReader("x"))
	r.ReadRune()
	r.ReadRune()
	r.UnreadRune()
	_, _, err := r.ReadRune()
	if err == nil {
		t.Error("expected error at EOF")
	} else if err != io.EOF {
		t.Error("expected EOF; got", err)
	}
}
````
It is a built in `Go` library test. Tests are written by using `if` statement.

File: internal/pn_counter.go [hunk 15]
```diff
@@ -0,0 +1,253 @@
...
+			err = lastError
+		} else {
+			err = NewHazelcastNoDataMemberInClusterError("cannot invoke operations on a CRDT because the cluster does not contain any data members", nil)
+		}
+		return
+	}
+	request := PNCounterGetEncodeRequest(c.name, (*VectorClock)(atomic.LoadPointer(&c.observedClock)).EntrySet(), target.(*Address))
+	response, lastError = c.invokeOnAddress(request, target.(*Address))
+	if lastError != nil {
+		log.Printf("error occurred while invoking operation on target %v, choosing different target", target)
```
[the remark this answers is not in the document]
mustafaiman: I think there is no default support for log levels in Golang. So I am skipping this for now. But I think we should come up with a solution for log level support.

tests/proxy/pn_counter/pn_counter_test.go
mustafaiman: This is a minor point. I am not strongly for one side.

PR #208, opened 2018-04-03
refactor predicates

SaitTalhaNisanci:
This pr moves predicates to `core/predicates` from `core` so that they can be used as:
``` predicates.And()```
instead of 
```core.And()```

### DIFF
--- a/core/predicates/predicates.go
+++ b/core/predicates/predicates.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package core
+package predicates
 
 import (
 	. "github.com/hazelcast/hazelcast-go-client/internal/predicates"
--- a/samples/map/predicate/predicate_example.go
+++ b/samples/map/predicate/predicate_example.go
@@ -17,7 +17,7 @@ package main
 import (
 	"fmt"
 	"github.com/hazelcast/hazelcast-go-client"
-	"github.com/hazelcast/hazelcast-go-client/core"
+	"github.com/hazelcast/hazelcast-go-client/core/predicates"
 	"strconv"
 )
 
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by furkansenharputlu** [comment]
I think we should create a `struct` named `Predicates` and these helper functions should be part of this struct. `Projection` and `Aggregate` are also like `Predicate`. Otherwise, we will create 3 packages having just one file and this will be a little weird.

**Comment by SaitTalhaNisanci** [comment]
I dont think doing a workaround like that is convenient here because it is not `go` way of doing that. Having packages with a single file seem okay in my opinion, there are a lot of packages with  only one go file in go source code as well.

**Comment by furkansenharputlu** [comment]
I made an example for projections:
```
// Projections is a utility variable to create basic Projection instances.
var Projections = &ProjectionHelpers{}


// ProjectionHelpers is a utility struct to create basic Projection instances.
// Do not create instance of this struct. It is already created with name Projections. Use it.
type ProjectionHelpers struct {
}

// SingleAttribute returns a projection that extracts the value of the given `attributePath`.
func (p *ProjectionHelpers) SingleAttribute(attributePath string) *SingleAttributeProjection {
	return NewSingleAttributeProjection(attributePath)
}

// MultiAttribute returns a projection that extracts the value of the given `attributePaths`.
// The attribute values will be returned as an []interface{} slice from each projection call.
func (p *ProjectionHelpers) MultiAttribute(attributePaths ...string) *MultiAttributeProjection {
	return NewMultiAttributeProjection(attributePaths)
}

// Identity returns a projection that does no transformation.
// Since 3.10.
func (p *ProjectionHelpers) Identity() *IdentityProjection {
	return NewIdentityProjection()
}
```

asimarslan:
I think we should only use variables if we actually need to keep some state otherwise we should just use packages.
See this [SO question& answer](https://stackoverflow.com/questions/44991854/equivalence-of-static-methods-in-go?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa) :

PR #209, opened 2018-04-03
remove chaining from configs

SaitTalhaNisanci:
This pr removes chaining from configs

This pr fixes a typo in samples as well.


mustafaiman:
I don't think it is very common in Golang to use method chaining. Specifically because, methods sometimes return error and you cannot handle that error happened int the middle of the method chain. I suggest keeping chaining out of this codebase as much as possible.

mustafaiman:
@sancar wdyt? 🔼

sancar:
Personally, I see no harm to use chaining where possible. The config methods that needs to return error is relatively low.

SaitTalhaNisanci:
Chaining is not idiomatic in go since methods return errors in general, but as @sancar said in our case it is only a few methods. https://stackoverflow.com/questions/27297702/go-method-chaining-and-error-handling


How should we proceed with this ? @sancar  @mustafaiman

sancar:
We can drop chaining for go. I am OK with this.

SaitTalhaNisanci:
Okay I will drop chaining from existing methods in this PR then.

PR #210, opened 2018-04-03
combine running tests and add dependencies in build sh

### LINKED ISSUE #202: build.sh running tests
Currently tests are run twice, the first round is with `-race` flag to check if there is any race condition and the second round is to generate coverage report, we need to combine these two.

SaitTalhaNisanci:
Fixes #202.

This pr combines race and coverage flags into one command to avoid running tests twice.
If thrift exists it wont be downloaded again.
There seems to be a problem with thrift repository therefore it is changed to its github mirror.
Test output is redirected to both stdout and test.out

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -23,11 +23,11 @@ export PACKAGE_LIST=$(go list $CLIENT_IMPORT_PATH/... | grep -vE ".*/tests|.*/co
 echo $PACKAGE_LIST
 
 
-if [ -d $GOPATH/src/git.apache.org/thrift.git/ ]; then
+if [ -d $GOPATH/src/github.com/apache/thrift/  ]; then
     echo "thrift already exists, not downloading."
 else
-    go get git.apache.org/thrift.git/lib/go/thrift
-    pushd $GOPATH/src/git.apache.org/thrift.git/
+    go get github.com/apache/thrift/lib/go/thrift
+    pushd $GOPATH/src/github.com/apache/thrift
     git fetch --tags --quiet
     git checkout 0.10.0
     popd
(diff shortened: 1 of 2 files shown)


SaitTalhaNisanci:
verify

SaitTalhaNisanci:
verify

build.sh
```diff
@@ -36,35 +42,47 @@ bash ./start-rc.sh
...
-echo "mode: set" > coverage.out
+
+echo "mode: atomic" > coverage.out
+
 for pkg in $(go list $CLIENT_IMPORT_PATH/...);
 do
     if [[ $pkg != *"vendor"* ]]; then
       echo "testing... $pkg"
-      go test -v -coverprofile=tmp.out -coverpkg ${PACKAGE_LIST} $pkg >> test.out
+      go test -race -covermode=atomic  -v -coverprofile=tmp.out ${pkg}
```
asimarslan: We don't need atomic run count. `covermode` `set` is enough.

build.sh
```diff
@@ -36,35 +42,47 @@ bash ./start-rc.sh
...
 
 # Run tests (JUnit plugin) with race detection
-for pkg in $(go list $CLIENT_IMPORT_PATH/...);
-do
-    if [[ $pkg != *"vendor"* ]]; then
-      echo "testing with race detection on ... $pkg"
-      go test -race -v  $pkg
-    fi
-done
+#for pkg in $(go list $CLIENT_IMPORT_PATH/...);
```
asimarslan: Please remove commented out code block

build.sh
SaitTalhaNisanci: When running race and coverage together the most secure way is to do it with atomic run count otherwise when doing coverage race detector gives race condition since the extra lines that are added for coverage are not atomic .

build.sh
SaitTalhaNisanci: I will clean up the script after we agree on the essentials

PR #211, opened 2018-04-05
replace IPredicate with interface{}

SaitTalhaNisanci:
Predicates were only `IdentifiedDataSerializable` which is different than the Java client. So we have decided to change it to `interface`.


SaitTalhaNisanci:
verify

PR #212, opened 2018-04-08
Lock implementation

byonchev:


byonchev:
Messed up git tree, reopened as #213

PR #213, opened 2018-04-08
lock implementation

byonchev:


sancar:
Hi @byonchev , thanks for the pr, but this implementation have a problem regarding usage of thread id. 
In hazelcast, locks are designed to be re-entrant. Meaning,  same thread can acquire the lock multiple times. 
If we use same thread id everywhere that means that every go-routine can acquire the lock inside same client. And that makes the lock practically useless. 
We need to able to assign different thread id's to locks used by different go-routine's. There seems to be a hacky way to access go-routine id but it is not advised to be used. 
https://blog.sgmansfield.com/2015/12/goroutine-ids/

We have two different solutions here:
1. Implement a non reentrant lock service on server side for specifically for go client. 
2. Make thread id parametric in ILock interface so that user passes go-routine id. This sounds like avoiding problem for somebody else to solve but an application developer can numerate their go channels and pass appropriate id. But definitely, this design is more error prone than the first one. 

Again, thanks for the pr but we can't merge it as it is.

PR #214, opened 2018-04-09
[DONT MERGE] fix

sancar:


internal/cluster.go
```diff
@@ -363,10 +360,19 @@ func (cs *clusterService) GetMemberByUUID(uuid string) core.Member {
 	return nil
 }
 
+func (cs *clusterService) getOwnerConnectionAddress() *protocol.Address {
+	address, ok := cs.ownerConnectionAddress.Load().(*protocol.Address)
```
SaitTalhaNisanci: When can this loading return an error? I think it will always return `ok=true`.

internal/connection.go
```diff
@@ -178,21 +176,21 @@ func (c *Connection) close(err error) {
...
-		", endpoint=%s:%d"+
+		", endpoint=%s"+
 		", lastReadTime=%s"+
 		", lastWriteTime=%s"+
 		", closedTime=%s"+
 		", lastHeartbeatRequested=%s"+
 		", lastHeartbeatReceived=%s"+
 		", connected server version=%s", c.isAlive(), c.connectionID,
-		c.endpoint.Load().(*protocol.Address).Host(), c.endpoint.Load().(*protocol.Address).Port(),
+		c.endpoint.Load().(core.Address).String(),
```
SaitTalhaNisanci: I dont think `.String()` is necessary here

internal/connection_manager.go
```diff
@@ -29,30 +28,33 @@ const (
...
-	client              *HazelcastClient
-	connections         map[string]*Connection
-	lock                sync.RWMutex
-	nextConnID          int64
-	connectionListeners atomic.Value
-	listenerLock        sync.Mutex
-}
+type connectionManager interface {
+	//returns a snapshot of active connections
+	getActiveConnections() map[string]*Connection
```
SaitTalhaNisanci: the convention of documentation is `method name ....` so they should be in the form of `getActiveConnections returns ...`

internal/connection_manager.go
```diff
@@ -29,30 +28,33 @@ const (
...
 
-func newConnectionManager(client *HazelcastClient) *connectionManager {
-	cm := connectionManager{client: client,
-		connections: make(map[string]*Connection),
-	}
-	cm.connectionListeners.Store(make([]connectionListener, 0)) //Initialize
-	return &cm
-}
+	//returns connection if available, null otherwise
```
SaitTalhaNisanci: nil

internal/connection_manager.go
```diff
@@ -29,30 +28,33 @@ const (
...
 
-type connectionManager struct {
-	client              *HazelcastClient
-	connections         map[string]*Connection
-	lock                sync.RWMutex
-	nextConnID          int64
-	connectionListeners atomic.Value
-	listenerLock        sync.Mutex
-}
+type connectionManager interface {
```
SaitTalhaNisanci: Using an interface in the same package seems a little bit odd to me. I feel like it would be better if we actually had separate packages for these and the communication was via interfaces.

internal/connection_manager.go
```diff
@@ -104,35 +105,63 @@ func (cm *connectionManager) connectionClosed(connection *Connection, cause erro
...
+	}
 }
 
-func (cm *connectionManager) getConnection(address core.Address, asOwner bool) *Connection {
-	cm.lock.RLock()
-	defer cm.lock.RUnlock()
-	return cm.getConnectionInternal(address, asOwner)
+//internal definitions and methods called inside connection manager process
+
+type connectionManagerImpl struct {
```
SaitTalhaNisanci: I think this name does not suit go well. It should be `connectionManager` and there is also an interface named `connectionManager` but as I said earlier, I dont think having an interface and struct in the same package is go way.

internal/connection.go
sancar: Interesting. Why it is used for every other field here ? I am removing all of them then.

internal/cluster.go
sancar: if ownerConnectionAddress is nil, ok becomes false

internal/connection.go
SaitTalhaNisanci: Yeah `String()` method is already called when trying to print a struct etc.

internal/connection_manager.go
sancar: This suits my purpose perfectly. I don't need another package for this. Why introduce one ?

All other services access connection manager over its interface. connectionManagerImpl is hidden to anyone else.

internal/connection_manager.go
sancar: ok, I will change this

internal/connection_manager.go
sancar: ok, I will change this

internal/connection_manager.go
```diff
@@ -148,55 +177,46 @@ func (cm *connectionManager) getConnectionInternal(address core.Address, asOwner
...
+	return nil, err
+}
 
+func (cm *connectionManagerImpl) NextConnectionID() int64 {
+	cm.nextConnectionID = cm.nextConnectionID + 1
+	return cm.nextConnectionID
 }
 
-func (cm *connectionManager) authenticate(connection *Connection, asOwner bool) error {
+func (cm *connectionManagerImpl) encodeRequest(asOwner bool) *protocol.ClientMessage {
```
SaitTalhaNisanci: I think we can name this as `encodeAuthenticationRequest` as this method only encodes Authentication requests.

internal/protocol/core.go
```diff
@@ -76,9 +81,7 @@ func (m *Member) Attributes() map[string]string {
 }
 
 func (m *Member) String() string {
-	memberInfo := fmt.Sprintf("Member [%s]:%d - %s", m.Address().Host(), m.Address().Port(),
-		m.UUID(),
-	)
+	memberInfo := fmt.Sprintf("Member %s - %s", m.address.String(), m.UUID())
```
SaitTalhaNisanci: `.String()` is unnecessary, `m.address` is enough.

internal/connection_manager.go
sancar: ok

internal/protocol/core.go
sancar: ok

internal/connection_manager.go
furkansenharputlu: I think we should have a `connection` package. Then, this interface will be more meaningful.

internal/connection_manager.go
```diff
@@ -63,36 +65,35 @@ func (cm *connectionManager) addListener(listener connectionListener) {
...
+}
+
+func (cm *connectionManagerImpl) getActiveConnections() map[string]*Connection {
 	connections := make(map[string]*Connection)
-	cm.lock.RLock()
-	defer cm.lock.RUnlock()
+	cm.connectionsMutex.RLock()
+	defer cm.connectionsMutex.RUnlock()
 	for k, v := range cm.connections {
 		connections[k] = v
```
furkansenharputlu: Why don't we use write lock here?

furkansenharputlu: Why don't we use write lock here?

sancar: It looks confusing I guess.
`connections` is a local variable. I am only reading from cm.connections. Not writing to it. So a read lock is enough here.

internal/cluster.go
```diff
@@ -361,10 +358,19 @@ func (cs *clusterService) GetMemberByUUID(uuid string) core.Member {
 	return nil
 }
 
+func (cs *clusterService) getOwnerConnectionAddress() *proto.Address {
+	address, ok := cs.ownerConnectionAddress.Load().(*proto.Address)
```
furkansenharputlu: `return  cs.ownerConnectionAddress.Load().(*proto.Address)` is enough I think.

sancar: if you omit error parameter, when address is null, cast fails with panic.
It says that nil can not be cast to (*proto.Address)

PR #219, opened 2018-04-11
fix bug in readBool and ReadUInt8

SaitTalhaNisanci:


internal/common/buffer_util_test.go
```diff
@@ -58,7 +58,7 @@ func TestReadBool(t *testing.T) {
 	buf := []byte{0, 0, 0}
 	WriteBool(buf, 0, true)
 	WriteBool(buf, 2, true)
-	if ReadBool(buf, 0) != true && ReadBool(buf, 2) != true {
+	if ReadBool(buf, 0) != true || ReadBool(buf, 2) != true {
```
furkansenharputlu: This should be `OR`. No need to change.

internal/common/buffer_util_test.go
```diff
@@ -67,7 +67,7 @@ func TestReadUInt8(t *testing.T) {
 	buf := []byte{0, 0, 0}
 	WriteUInt8(buf, 1, 5)
 	WriteUInt8(buf, 2, 12)
-	if ReadUInt8(buf, 1) != 5 && ReadUInt8(buf, 2) != 12 {
+	if ReadUInt8(buf, 1) != 5 || ReadUInt8(buf, 2) != 12 {
```
furkansenharputlu: same

internal/common/buffer_util_test.go
furkansenharputlu: If one of them is not `true`. It should fail.

SaitTalhaNisanci: That is why it should be 'or'. Currently it is 'and' so if one of them fails it doesnt fail

furkansenharputlu: I am confused. Ignore this.

PR #222, opened 2018-04-12
refactor constant names according to Go convention

furkansenharputlu:


furkansenharputlu:
verify

SaitTalhaNisanci:
Currently all constants are public, we should change it either in this PR or in another one. It is probably better to change it in this one, what do you think ?

furkansenharputlu:
In Java,

PR #223, opened 2018-04-12
Example idiomatic go

asim:
This is a PR to demonstrate idiomatic Go. I believe the current hazelcast client is likely written by those well versed in Java and this shows in the way the client is written. I believe you would benefit significantly by hiring people with Go experience to provide a more natural client. This PR serves merely as a quick hacked example based on my years of experience. 

Do not merge the PR. Please close once having read.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Hazelcast Go Client  
+# Go Hazelcast Client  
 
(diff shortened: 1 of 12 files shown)


asimarslan:
Hi @asim ,

Thanks for your feedback. We will review your comments.

PR #225, opened 2018-04-13
remove unnecessary interfaces from api

furkansenharputlu:
#245

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -50,58 +50,6 @@ type Pair interface {
 	Value() interface{}
 }
 
-// DistributedObjectInfo contains name and service name of distributed objects.
-type DistributedObjectInfo interface {
-	// Name returns the name of distributed object.
-	Name() string
-
-	// ServiceName returns the service name of distributed object.
-	ServiceName() string
-}
-
-// Error contains error information that occurred in the server.
-type Error interface {
-	// ErrorCode returns the error code.
-	ErrorCode() int32
-
-	// ClassName returns the class name where error occurred.
-	ClassName() string
-
-	// Message returns the error message.
-	Message() string
-
-	// StackTrace returns a slice of StackTraceElement.
-	StackTrace() []StackTraceElement
-
-	// CauseErrorCode returns the cause error code.
-	CauseErrorCode() int32
-
-	// CauseClassName returns the cause class name.
-	CauseClassName() string
-}
-
-type StackTraceElement interface {
-	// DeclaringClass returns the fully qualified name of the class containing
-	// the execution point represented by the stack trace element.
-	DeclaringClass() string
-
-	// MethodName returns the name of the method containing the execution point
-	// represented by this stack trace element.
-	MethodName() string
-
-	// FileName returns the name of the file containing the execution point
-	// represented by the stack trace element, or nil if
-	// this information is unavailable.
-	FileName() string
-
-	// LineNumber returns the line number of the source line containing the
-	// execution point represented by this stack trace element, or
-	// a negative number if this information is unavailable. A value
-	// of -2 indicates that the method containing the execution point
-	// is a native method.
-	LineNumber() int32
-}
-
 // EntryView represents a readonly view of a map entry.
 type EntryView interface {
 	// Key returns the key of the entry.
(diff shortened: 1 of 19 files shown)


sancar:
What is the reasoning behind removing IItemEvent ?
For example, should we remove IEntryEvent interface also for same reason ?

furkansenharputlu:
In Hazelcast Java, `ItemEvent` is an implementation in `core` package. There is no need to use interface in Go. I made same for `EntryEvent` and `MapEvent`.

furkansenharputlu:
verify

internal/cluster.go
```diff
@@ -351,8 +351,7 @@ func (cs *clusterService) GetMember(address core.Address) core.Member {
 	membersList := cs.members.Load().([]*protocol.Member)
 	for _, member := range membersList {
 		if *member.Address().(*protocol.Address) == *address.(*protocol.Address) {
-			copyMember := *member
-			return &copyMember
+			return member
```
SaitTalhaNisanci: What is the reason for this change?

core/api.go
```diff
@@ -141,30 +89,52 @@ type EntryView interface {
...
+	Name() string
+
+	// Member returns the member that fired this event.
+	Member() Member
+
+	// EventType returns the type of entry event.
+	EventType() int32
+
+	// ToString returns a string representation of this event.
+	ToString() string
```
SaitTalhaNisanci: Why is this `ToString` ? In Go it should be `String` which is `ToString` in Java.

internal/map.go
```diff
@@ -539,29 +539,30 @@ func (mp *mapProxy) AddEntryListenerToKeyWithPredicate(listener interface{}, pre
...
 	valueData *serialization.Data, mergingValueData *serialization.Data, eventType int32, uuid *string,
 	numberOfAffectedEntries int32, includedValue bool, listener interface{}) {
+	member := mp.client.ClusterService.GetMemberByUUID(*uuid)
 	key, _ := mp.toObject(keyData)
 	oldValue, _ := mp.toObject(oldValueData)
 	value, _ := mp.toObject(valueData)
 	mergingValue, _ := mp.toObject(mergingValueData)
-	entryEvent := protocol.NewEntryEvent(key, oldValue, value, mergingValue, eventType, uuid)
-	mapEvent := protocol.NewMapEvent(eventType, uuid, numberOfAffectedEntries)
+	entryEvent := protocol.NewEntryEvent(mp.name, member, eventType, key, oldValue, value, mergingValue)
```
SaitTalhaNisanci: Why are we changing this order?

core/api.go
```diff
@@ -141,30 +89,52 @@ type EntryView interface {
 	TTL() time.Duration
 }
 
+// AbstractMapEvent is base for a map event.
+type AbstractMapEvent interface {
```
SaitTalhaNisanci: I couldnt find the usage for `AbstractMapEvent` on the client side. Can you show where it is used on the client side?

core/api.go
```diff
@@ -260,19 +217,19 @@ type LifecycleListener interface {
...
 	// OnMessage is invoked when a message is received for the added topic. Note that topic guarantees message ordering.
 	// Therefore there is only one thread invoking OnMessage.
-	OnMessage(message TopicMessage)
+	OnMessage(message Message)
 }
 
-// TopicMessage is a message for Topic.
-type TopicMessage interface {
+// Message is a message for Topic.
+type Message interface {
```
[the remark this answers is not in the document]
furkansenharputlu: It is `Message` in Java.

internal/cluster.go
furkansenharputlu: Previous one seems more complex. We do not need to protect `Member`. We give it as interface to user.

internal/map.go
furkansenharputlu: Here, `member` and `name` were missing. I added them.

core/api.go
furkansenharputlu: In Java server side, some interfaces are declared and used in client side. Otherwise, same interface will be created twice. Therefore, that there is no `AbstractMapEvent` on the client side is normal.

mustafaiman: Here `AbstractMapEvent` is an interface. On Java side, it is named abstract because it is an abstract class. 
I think it is okay to have `MapEvent` and `EntryEvent` derive from a base interface since they have common methods but naming is not right. I would change its name to `Event`.

internal/protocol/core.go
```diff
@@ -354,13 +354,46 @@ func (e *StackTraceElement) LineNumber() int32 {
 	return e.lineNumber
 }
 
+type AbstractMapEvent struct {
+	name      *string
```
SaitTalhaNisanci: We should use `string` instead of `*string`here and in the rest of this PR.

core/api.go
```diff
@@ -220,26 +177,26 @@ type EntryMergedListener interface {
 	EntryMerged(EntryEvent)
```
SaitTalhaNisanci: event

PR #226, opened 2018-04-13
change comments according to go conventions

SaitTalhaNisanci:
This pr changes some comments according to go conventions.

Some typos are also fixed.

### DIFF
--- a/core/pn_counter.go
+++ b/core/pn_counter.go
@@ -54,65 +54,65 @@ type PNCounter interface {
 	IDistributedObject
 
 	// Get returns the current value of the counter.
-	// It returns HazelcastNoDataMemberInClusterError if the cluster does not contain any data members.
-	// It returns HazelcastUnsupportedOperationError if the cluster version is less  than 3.10.
-	// It returns HazelcastConsistencyLostError if the session guarantees have been lost.
+	// It returns HazelcastNoDataMemberInClusterError if the cluster does not contain any data members,
+	// HazelcastUnsupportedOperationError if the cluster version is less  than 3.10,
+	// HazelcastConsistencyLostError if the session guarantees have been lost.
 	Get() (currentValue int64, err error)
 
 	// GetAndAdd adds the given value to the current value.
-	// It returns the previous value.
-	// It returns HazelcastNoDataMemberInClusterError if the cluster does not contain any data members.
-	// It returns HazelcastUnsupportedOperationError if the cluster version is less  than 3.10.
-	// It returns HazelcastConsistencyLostError if the session guarantees have been lost.
+	// It returns the previous value,
+	// HazelcastNoDataMemberInClusterError if the cluster does not contain any data members,
+	// HazelcastUnsupportedOperationError if the cluster version is less  than 3.10,
+	// HazelcastConsistencyLostError if the session guarantees have been lost.
 	GetAndAdd(delta int64) (previousValue int64, err error)
 
 	// AddAndGet adds the given value to the current value.
-	// It returns the updated value.
-	// It returns HazelcastNoDataMemberInClusterError if the cluster does not contain any data members.
-	// It returns HazelcastUnsupportedOperationError if the cluster version is less  than 3.10.
-	// It returns HazelcastConsistencyLostError if the session guarantees have been lost.
+	// It returns the updated value,
+	// HazelcastNoDataMemberInClusterError if the cluster does not contain any data members,
+	// HazelcastUnsupportedOperationError if the cluster version is less  than 3.10,
+	// HazelcastConsistencyLostError if the session guarantees have been lost.
 	AddAndGet(delta int64) (updatedValue int64, err error)
 
 	// GetAndSubtract subtracts the given value to the current value.
-	// It returns the previous value.
-	// It returns HazelcastNoDataMemberInClusterError if the cluster does not contain any data members.
-	// It returns HazelcastUnsupportedOperationError if the cluster version is less  than 3.10.
-	// It returns HazelcastConsistencyLostError if the session guarantees have been lost.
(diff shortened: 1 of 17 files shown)


core/pn_counter.go
```diff
@@ -54,65 +54,65 @@ type PNCounter interface {
...
+	// HazelcastUnsupportedOperationError if the cluster version is less  than 3.10,
+	// HazelcastConsistencyLostError if the session guarantees have been lost.
 	Get() (currentValue int64, err error)
 
 	// GetAndAdd adds the given value to the current value.
-	// It returns the previous value.
-	// It returns HazelcastNoDataMemberInClusterError if the cluster does not contain any data members.
-	// It returns HazelcastUnsupportedOperationError if the cluster version is less  than 3.10.
-	// It returns HazelcastConsistencyLostError if the session guarantees have been lost.
+	// GetAndAdd returns the previous value.
```
furkansenharputlu: These changes are unnecessary. `It` references `GetAndAdd `.
Look at this. This is a built-in method and uses `it` while referencing method name.

````
// grow grows the buffer to guarantee space for n more bytes.
// It returns the index where bytes should be written.
// If the buffer can't grow it will panic with ErrTooLarge.
func (b *Buffer) grow(n int) int {
````

There are many like this. This is not about `Go` convention.

File: core/pn_counter.go [hunk 2]
```diff
@@ -54,65 +54,65 @@ type PNCounter interface {
...
+	// HazelcastConsistencyLostError if the session guarantees have been lost.
 	GetAndAdd(delta int64) (previousValue int64, err error)
 
 	// AddAndGet adds the given value to the current value.
-	// It returns the updated value.
-	// It returns HazelcastNoDataMemberInClusterError if the cluster does not contain any data members.
-	// It returns HazelcastUnsupportedOperationError if the cluster version is less  than 3.10.
-	// It returns HazelcastConsistencyLostError if the session guarantees have been lost.
+	// AddAndGet returns the updated value.
+	// AddAndGet returns HazelcastNoDataMemberInClusterError if the cluster does not contain any data members,
```
furkansenharputlu: same here

core/predicates/predicates.go
```diff
@@ -12,98 +12,103 @@
...
 
+// Package predicates is a utility package to create predicates.
 package predicates
 
 import (
 	. "github.com/hazelcast/hazelcast-go-client/internal/predicates"
 )
 
-// Sql is a helper function for creating SqlPredicate.
+// Sql returns a SqlPredicateSql with the given sql.
```
furkansenharputlu: These changes are also unnecessary. There is nothing wrong at previous one.

core/ringbuffer.go
```diff
@@ -115,9 +115,10 @@ const (
 	OverflowPolicyFail policy = 1
 )
 
-// This type is not used by user.
+// policy type is not used by user.
```
furkansenharputlu: This is not a GoDoc. It is a comment. Is there any  Go convention in comments ?

core/ringbuffer.go
```diff
@@ -115,9 +115,10 @@ const (
...
 type policy int32
 
+// Policy returns the policy type.
```
furkansenharputlu: This is not a public function so it is not necessary to add GoDoc.

serialization/api.go
```diff
@@ -82,334 +82,334 @@ type Serializer interface {
...
-	// Write writes object to ObjectDataOutput.
+	// Write writes an object to ObjectDataOutput.
 	Write(output DataOutput, object interface{}) error
 }
 
 // IData is the basic unit of serialization. It stores binary form of an object serialized
 // by SerializationService's ToData() method.
 type IData interface {
-	// Returns byte array representation of internal binary format.
+	// Buffer returns byte array representation of internal binary format.
```
furkansenharputlu: This is a built-in Go method and its GoDoc.
```
// Return the list of export search paths for this GccgoInstallation.
func (inst *GccgoInstallation) SearchPaths() (paths []string) {
```

serialization/api.go
```diff
@@ -82,334 +82,334 @@ type Serializer interface {
...
 }
 
 // DataOutput provides serialization methods.
 type DataOutput interface {
-	// Returns the head position in the byte array.
+	// Position returns the head position in the byte array.
 	Position() int32
 
-	// Sets the head position in the byte array.
+	// SetPosition sets the head position in the byte array.
```
furkansenharputlu: This is a built-in Go method and its GoDoc.
```
// Return the list of export search paths for this GccgoInstallation.
func (inst *GccgoInstallation) SearchPaths() (paths []string) {
```

core/pn_counter.go
Serdaro: Agreed.

core/replicated_map.go
```diff
@@ -39,7 +39,7 @@ type ReplicatedMap interface {
 	// PutWithTtl associates a given value to the specified key and replicates it to the
 	// cluster. If there is an old value, it will be replaced by the specified
 	// one and returned from the call. In addition, you have to specify a ttl and its time unit
-	// to define when the value is outdated and thus should be removed from the
+	// to define when the value is outdated and this should be removed from the
```
Serdaro: I think "thus" was used correctly in that context, tho it is redundant. You can either change "this" to "thus", or just remove "this".

core/predicates/predicates.go
SaitTalhaNisanci: We have used `returns` pattern for all `New` functions so we should keep it consistent.

serialization/api.go
SaitTalhaNisanci: The convention is to start with the method name

PR #228, opened 2018-04-13
fix some warnings from vet

SaitTalhaNisanci:
This pr fixes some vet warnings.

### DIFF
--- a/internal/common/buffer_util_test.go
+++ b/internal/common/buffer_util_test.go
@@ -26,7 +26,7 @@ func TestWriteInt32(t *testing.T) {
 	expectedBuf := []byte{0, 5, 0, 0, 0, 0, 0, 0}
 
 	if bytes.Compare(buf, expectedBuf) != 0 {
-		t.Errorf("WriteInt32() makes ", buf, " expected ", expectedBuf)
+		t.Error("WriteInt32() makes ", buf, " expected ", expectedBuf)
 	}
 }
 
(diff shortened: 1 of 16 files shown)


internal/common/buffer_util_test.go
```diff
@@ -26,7 +26,7 @@ func TestWriteInt32(t *testing.T) {
 	expectedBuf := []byte{0, 5, 0, 0, 0, 0, 0, 0}
 
 	if bytes.Compare(buf, expectedBuf) != 0 {
-		t.Errorf("WriteInt32() makes ", buf, " expected ", expectedBuf)
+		t.Error("WriteInt32() makes ", buf, " expected ", expectedBuf)
```
furkansenharputlu: This change is unnecessary. This is a built-in test. They make same job.
```
func TestNewEqual(t *testing.T) {
	// Different allocations should not be equal.
	if errors.New("abc") == errors.New("abc") {
		t.Errorf(`New("abc") == New("abc")`)
	}
	if errors.New("abc") == errors.New("xyz") {
		t.Errorf(`New("abc") == New("xyz")`)
	}

	// Same allocation should be equal to itself (not crash).
	err := errors.New("jkl")
	if err != err {
		t.Errorf(`err != err`)
	}
}
````

SaitTalhaNisanci: `Errorf` should be used when there is formatting, when there is no formatting `Error` should be used. `go vet` complains about this.

PR #229, opened 2018-04-13
rearrange imports according to standards

SaitTalhaNisanci:
This pr refactors imports and removes unused ones.

PR #230, opened 2018-04-13
use goimports instead of gofmt

SaitTalhaNisanci:
This pr adds `goimports` to `build.sh`. We were using `gofmt` to detect formatting problems but `goimports` does everything `gofmt` does and it also adds missing imports or removes unused imports and it groups imports.

Build will fail if `goimports` detects any problems.

This pr depends on #229.


SaitTalhaNisanci:
verify

sancar:
@SaitTalhaNisanci Can you change this also in localTests.sh

SaitTalhaNisanci:
This pr is no longer needed as we are using gometalinter.

PR #232, opened 2018-04-16
remove dots from imports

SaitTalhaNisanci:
This PR removes dot imports.

This is from https://github.com/golang/go/wiki/CodeReviewComments#import-dot:

In this case, the test file cannot be in package foo because it uses bar/testutil, which imports foo. So we use the 'import .form to let the file pretend to be part of package foo even though it is not. Except for this one case, ```do not use import . in your programs. It makes the programs much harder to read because it is unclear whether a name like Quux is a top-level identifier in the current package or in an imported package.```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -17,7 +17,7 @@ package config
 
 import (
 	"github.com/hazelcast/hazelcast-go-client/core"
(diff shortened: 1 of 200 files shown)


SaitTalhaNisanci:
verify

SaitTalhaNisanci:
verify

PR #233, opened 2018-04-16
refactor receiver names according to Go convention

furkansenharputlu:
This pr refactors receiver names according to Go [convention](https://github.com/golang/go/wiki/CodeReviewComments#receiver-names).


sancar:
I did not really like this convention. I prefer long descriptive names. But since we are following the go convention, I will give my OK anyway.

PR #234, opened 2018-04-16
remove I prefix from interface names

furkansenharputlu:
This pr removes `I` prefix from some interface names. We decided to make Go Client more compatible with Go conventions.
#245

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -17,8 +17,8 @@ package core
 
 import "time"
 
-// IAddress represents an address of a member in the cluster.
-type IAddress interface {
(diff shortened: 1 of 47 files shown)


SaitTalhaNisanci:
I think we shouldnt use *Impl. As the convention is not to use dot imports I dont think there will be any problems when they are the same name. For example `Address` in core would be `core.Address` and `Address` in protocol would be `protocol.Address`.

furkansenharputlu:
The thing that an interface and its implemented struct have same name is weird I think. I used ‘Impl’. What do you think ? @sancar @mustafaiman .

PR #235, opened 2018-04-16
Time related API's converted to use time.* types.

sancar:
To represent fix time in point time.Time
To represent a duration time.Duration is used. 

Internal usages where duration used with wrong resolution is also fixed.


sancar:
run-lab-run

PR #243, opened 2018-04-27
change some pkg names according to go convention

SaitTalhaNisanci:
This pr removes _ or - from package names, it also removes plural package names.

This pr also shortens `protocol` to `proto` since the convention is to use short package names.

Some packages such as `common` should also be changed as they are not good names for packages but it will be fixed in another PR as it is not trivial.  Note that this is already fixed #250.


SaitTalhaNisanci:
verify

PR #247, opened 2018-04-27
implement projection and test

furkansenharputlu:
This pr implements `SingleAttribute` projection and tests.
Also, it adds `Project` and `ProjectWithPredicate` functions to `Map`.
 #205


core/projection/projection.go
```diff
@@ -0,0 +1,22 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package projection
+
+import "github.com/hazelcast/hazelcast-go-client/internal/projection"
+
+// SingleAttribute returns a projection that extracts the value of the given `attributePath`.
```
mustafaiman: `attributePath must not be empty`. Does this piece of comment apply here? I think it should be mentioned.

internal/projection/projections_impl.go
```diff
@@ -0,0 +1,52 @@
...
+package projection
+
+import (
+	"github.com/hazelcast/hazelcast-go-client/internal/common"
+	"github.com/hazelcast/hazelcast-go-client/serialization"
+)
+
+const singleAttributeProjectionID = 0
+
+type SingleAttribute struct {
```
mustafaiman: This is Javadoc for `SingleAttributeProjection`
```

/**
 * Projection that extracts the values of the given attribute and returns it.
 * <p>
 * The attributePath does not support the [any] operator.
 * The input object has to be an instance of Extractable in order for the projection to work.
 *
 * @param <I> type of the input
 */
```
I think some of this information applies here.

PR #248, opened 2018-04-27
fix bugs in linter.sh

SaitTalhaNisanci:
get gometalinter if it doesnt exist and install the dependencies. 

Also we shouldnt run `gometalinter` when it detects problems as it cannot solve the problems.

### DIFF
--- a/linter.sh
+++ b/linter.sh
@@ -1,14 +1,28 @@
 #!/usr/bin/env bash
 
-gometalinter ./... 2>&1 | read; [ $? == 1 ]
+if ! [ -d $GOPATH/src/github.com/alecthomas/gometalinter/  ]; then
+    echo "downloading gometalinter"
+    go get -u github.com/alecthomas/gometalinter
+    gometalinter --install
+    if [  "$?" != "0" ] ; then
+        echo "error when installing gometalinter"
+    exit 1
+    fi
+fi
+
+gometalinter ./... 2>&1 > gometalinter-err.out
 
 
-if [ "$?" = "1" ]; then
+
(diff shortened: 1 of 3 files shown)


linter.sh
```diff
@@ -1,14 +1,23 @@
...
 
-gometalinter ./... 2>&1 | read; [ $? == 1 ]
+if [ -d $GOPATH/src/github.com/alecthomas/gometalinter/  ]; then
+    echo "gometalinter already exists, not downloading."
+else
+    go get -u github.com/alecthomas/gometalinter
+    gometalinter --install
+fi
+
+gometalinter ./... 2>&1 > temp.err
```
mustafaiman: can we use a meaningful name instead of `temp.err`.  Suggestion: `gometalinter-err.out`

linter.sh
```diff
@@ -1,14 +1,23 @@
 #!/usr/bin/env bash
 
-gometalinter ./... 2>&1 | read; [ $? == 1 ]
+if [ -d $GOPATH/src/github.com/alecthomas/gometalinter/  ]; then
+    echo "gometalinter already exists, not downloading."
```
furkansenharputlu: It is better not to print this because for each run, it will print this. When it is ok, it is enough to print `code format is ok!`.

linter.sh
```diff
@@ -1,14 +1,23 @@
...
+else
+    go get -u github.com/alecthomas/gometalinter
+    gometalinter --install
```
furkansenharputlu: We may say here something like `gometalinter is downloading ...`

SaitTalhaNisanci:
verify

SaitTalhaNisanci:
verify

Issue #245, opened 2018-04-27
Refactor


SaitTalhaNisanci:
- [X] Make receiver names shorter 
- [X] Fix go vet errors.
- [ ] Make package names shorter
- [ ] Remove unnecessary names from Methods like Hazelcast from `hazelcast.NewHazelcastClient()`
- [X] Add dot imports
- [X] Use `goimports` 
- [ ] Fix warnings from [https://github.com/alecthomas/gometalinter](https://github.com/alecthomas/gometalinter)
- [ ] Refactor tests 
- [X] Refactor constant names to camel case
- [ ] Remove `I` prefix from `IMap` etc
- [ ] Remove unused interfaces
- [ ] Change package names such as `Common`

PR #251, opened 2018-04-29
remove Get prefixes from some function names

furkansenharputlu:
This pr removes `Get` prefixes from some function names in api. (`GetMap`, `GetList` etc.) 
#245


furkansenharputlu:
We decided to close this pr because the functions(`GetMap`, `GetList` etc.) are not getters.

PR #252, opened 2018-05-02
fixing opening multiple connections to same address

sancar:
connection opening made under write lock as fix.

### DIFF
--- a/internal/connection_manager.go
+++ b/internal/connection_manager.go
@@ -35,7 +35,7 @@ type connectionManager struct {
 	lock                sync.RWMutex
 	nextConnID          int64
 	connectionListeners atomic.Value
-	mu                  sync.Mutex
+	listenerLock        sync.Mutex
 }
 
 func newConnectionManager(client *HazelcastClient) *connectionManager {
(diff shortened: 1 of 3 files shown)


internal/connection_manager.go
```diff
@@ -138,23 +154,31 @@ func (cm *connectionManager) ConnectionCount() int32 {
 	return int32(len(cm.connections))
 }
 
-func (cm *connectionManager) openNewConnection(address core.Address, resp chan *Connection, asOwner bool) error {
+func (cm *connectionManager) getOrConnectInternal(address core.Address, asOwner bool) (*Connection, error) {
+	cm.lock.Lock()
```
SaitTalhaNisanci: Instead of using write lock and checking if connection exists, would it be better to first check if connection exists with `getConnection` which is guarded under read lock instead of `getConnectionInternal` which is not, and then if it doesnt exist require write locks and connect.

SaitTalhaNisanci: I guess that would result in opening multiple connections though so nevermind

internal/connection_manager.go
```diff
@@ -46,13 +46,13 @@ func newConnectionManager(client *HazelcastClient) *connectionManager {
 	return &cm
 }
 
-func (cm *connectionManager) nextConnectionID() int64 {
+func (cm *connectionManager) NextConnectionID() int64 {
```
SaitTalhaNisanci: As far as I see this is made public just for the sake of tests. I cant think of any better way off the top of my head other than putting them into the same package, which is complicated for now.

internal/connection_manager.go
sancar: I have put the reading connection under readlock where this method is called. This is the optimization part. 
We need to check the `getConnectionInternal` under writeLock again to make sure not to open multiple connections.

Issue #254, opened 2018-05-03
validate listener arguments 


furkansenharputlu:
In listener methods, there is no type validation of listener arguments. User can give any type as a listener argument. For example, user can pass `5` as a listener and doesn't get error. Listener methods should return proper `HazelcastIllegalArgumentError` when wrong type is passed.

PR #255, opened 2018-05-03
validate listener arguments

furkansenharputlu:
This pr validates listener argument types. #254

### DIFF
--- a/core/map.go
+++ b/core/map.go
@@ -338,26 +338,30 @@ type Map interface {
 	PutTransient(key interface{}, value interface{}, ttl time.Duration) (err error)
 
 	// AddEntryListener adds a continuous entry listener for this map.
-	// To receive an event, you should implement a corresponding interface for that event such as
-	// EntryAddedListener, EntryRemovedListener, EntryUpdatedListener.
+	// To receive an event, listener should implement a corresponding interface for that event such as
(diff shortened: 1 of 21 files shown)


mustafaiman:
Can't we change the signature of `AddEntryListener(listener interface{}, includeValue bool)` to `AddEntryListener(listener EntryListener{}, includeValue bool)` and use some inheritance to have other listeners derive from `EntryListener`? This is what typing is for. We first ignore types in our API and then try to patch the same functionality with extra code.

core/map.go
```diff
@@ -338,26 +338,30 @@ type Map interface {
 	PutTransient(key interface{}, value interface{}, ttl time.Duration) (err error)
 
 	// AddEntryListener adds a continuous entry listener for this map.
-	// To receive an event, you should implement a corresponding interface for that event such as
-	// EntryAddedListener, EntryRemovedListener, EntryUpdatedListener.
+	// To receive an event, listener should implement a corresponding interface for that event such as
+	// EntryAddedListener, EntryRemovedListener, EntryUpdatedListener,EntryEvictedListener,
+	// EntryMergedListener, MapEvictedListener, MapEvictedListener.
```
SaitTalhaNisanci: `MapEvictedListener` is written twice in this one and the following ones. `MapCleared` is missing.

PR #263, opened 2018-05-14
remove I from some remaining interfaces

SaitTalhaNisanci:


internal/heartbeat.go
```diff
@@ -136,10 +136,10 @@ func (hbs *heartBeatService) shutdown() {
 	close(hbs.cancel)
 }
 
-type IOnHeartbeatStopped interface {
+type OnHeartbeatStopped interface {
```
furkansenharputlu: Can you merge and rename these listeners like in Java ?
```
public interface ConnectionHeartbeatListener {

    /**
     * This event will be fired when the heartbeat is resumed for a connection to a member.
     */
    void heartbeatResumed(Connection connection);

    /**
     * This event will be fired when no heartbeat response is received for
     * {@link com.hazelcast.client.spi.properties.ClientProperty#HEARTBEAT_TIMEOUT} milliseconds from the member.
     */
    void heartbeatStopped(Connection connection);
}
```

PR #270, opened 2018-06-04
update codecs

SaitTalhaNisanci:
This pr makes `calculateSize` methods private and makes some small changes.


asimarslan:
@SaitTalhaNisanci does this PR make calculateSize methods public or private ?

SaitTalhaNisanci:
@asimarslan  It makes it private. If a method starts with a lower case it is private otherwise public.

PR #273, opened 2018-06-07
Fix client copy message logic

sancar:
When a message retried, the fields like acquiresResorce and
isRetryable are forgotten to be copied.
We make sure those fields are also carried to new client message
with this pr.

related to https://github.com/hazelcast/hazelcast/pull/13266


internal/proto/message.go
```diff
@@ -70,6 +70,13 @@ func NewClientMessage(buffer []byte, payloadSize int) *ClientMessage {
 	return clientMessage
 }
 
+func (m *ClientMessage) CloneMessage() *ClientMessage {
+	copiedBuffer := append([]byte{}, m.Buffer...)
```
furkansenharputlu: Why do you used `append` here?

sancar: It turns out that I don't need to use generics. I have updated the prd to use copy instead.

PR #279, opened 2018-06-13
implement cloud discovery

### LINKED ISSUE #276: Hazelcast Cloud Discovery
- [x] Implement Hazelcast Cloud Discovery
- [x] Write tests
- [ ] Sample Code

### LINKED ISSUE #278: Address Provider and Translator
We should use address provider and translators as in java client to make code more organized.

SaitTalhaNisanci:
This pr implements hazelcast cloud discovery with tests. fixes #276.

Address provider and translators are added for better modularity. fixes #278.

Currently the only way to enable cloud discovery is through programmatic configuration, it should be extended further after the implementation of `hazelcastProperty`.

### DIFF
--- a/internal/cluster.go
+++ b/internal/cluster.go
@@ -44,11 +44,12 @@ type clusterService struct {
 	uuid                   atomic.Value
 	ownerConnectionAddress atomic.Value
 	listeners              atomic.Value
+	addressProviders       []AddressProvider
 	mu                     sync.Mutex
 	reconnectChan          chan struct{}
 }
 
-func newClusterService(client *HazelcastClient, config *config.Config) *clusterService {
+func newClusterService(client *HazelcastClient, config *config.Config, addressProviders []AddressProvider) *clusterService {
 	service := &clusterService{client: client, config: config, reconnectChan: make(chan struct{}, 1)}
 	service.ownerConnectionAddress.Store(&proto.Address{})
 	service.members.Store(make([]*proto.Member, 0))       //Initialize
@@ -60,6 +61,7 @@ func newClusterService(client *HazelcastClient, config *config.Config) *clusterS
 	for _, membershipListener := range client.ClientConfig.MembershipListeners() {
 		service.AddListener(membershipListener)
 	}
+	service.addressProviders = addressProviders
 	service.client.ConnectionManager.addListener(service)
 	go service.process()
 	return service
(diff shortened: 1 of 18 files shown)


internal/discovery/hazelcast_cloud_discovery.go
```diff
@@ -0,0 +1,138 @@
...
+	return hzCloud
+}
+
+func (hzC *HazelcastCloud) discoverNodesInternal() (map[string]core.Address, error) {
+	return hzC.callService()
+}
+
+func (hzC *HazelcastCloud) callService() (map[string]core.Address, error) {
+	url := hzC.endPointURL
+	timeout := time.Duration(hzC.connectionTimeoutInMillis) * time.Millisecond
```
sancar: Why do we need this time conversion stuff ?
It looks like if we did not convert it in `HazelcastClient.getConnectionTimeoutMillis`(https://github.com/hazelcast/hazelcast-go-client/pull/279/files#diff-c9be6944b26f165a41c612366fa2b4e1R200), then it would not be necessary to convert here, right ?

sancar: We should keep the type as Duration as much as possible to avoid confusion

internal/discovery/hazelcast_cloud_discovery.go
```diff
@@ -0,0 +1,138 @@
...
+	endPointURL               string
+	connectionTimeoutInMillis int64
+	discoverNodes             nodeDiscoverer
+}
+
+func NewHazelcastCloud(cloudToken string, connectionTimeoutInMillis int64) *HazelcastCloud {
+	hzCloud := &HazelcastCloud{}
+	hzCloud.endPointURL = cloudURL + cloudToken
+	hzCloud.connectionTimeoutInMillis = connectionTimeoutInMillis
+	hzCloud.discoverNodes = hzCloud.discoverNodesInternal
```
sancar: nodeDiscoverer type and this line looks confusing. Why do we need that ? 
Are we trying to avoid a public method ? `DiscoverNodes`

sancar: Hmm. It looks like it to do mock testing
https://github.com/hazelcast/hazelcast-go-client/pull/279/files#diff-64b25b464ccecbf58536edcc6ed4ffb8R41

internal/cluster.go
```diff
@@ -69,6 +71,22 @@ func (cs *clusterService) start() error {
...
+	for _, member := range memberList {
+		addresses = append(addresses, member.Address())
+	}
+
+	providerAddresses := make([]core.Address, 0)
+	for _, provider := range cs.addressProviders {
+		providerAddresses = append(providerAddresses, provider.LoadAddresses()...)
+	}
+
+	addresses = append(addresses, providerAddresses...)
```
furkansenharputlu: Here, we should use set. Same address can be added more than one.

internal/addressTranslator.go
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
```
furkansenharputlu: it should be address_translator.go

internal/addressProvider.go
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
```
furkansenharputlu: address_provider.go

test/discovery/cloud_config_test.go
```diff
@@ -0,0 +1,60 @@
...
+	returnedCloudCfg := cfg.NetworkConfig().CloudConfig()
+	assert.Equalf(t, nil, true, returnedCloudCfg.IsEnabled(), "Cloud discovery should be enabled.")
+	assert.Equalf(t, nil, testDiscoveryToken, returnedCloudCfg.DiscoveryToken(), "Cloud discovery token "+
+		"should be set.")
+}
+
+func TestCloudConfigWithPropertySet(t *testing.T) {
+	cloudConfig := config.NewClientCloud()
+	cloudConfig.SetEnabled(true)
+	os.Setenv("hazelcast.client.cloud.discovery.token", "aaa")
```
furkansenharputlu: Can you give a meaningful name ? like `EXAMPLE_TOKEN`

PR #281, opened 2018-06-18
add hazelcast property

### LINKED ISSUE #280: Hazelcast Property
Currently, users can only use programmatic config. We should add a structure so that adding a property via environment variables should also be possible.

As in Java client, we should first check if a value is programmatically configured, then if environment variable is set, and lastly return the default value.

SaitTalhaNisanci:
This PR adds Hazelcast Property as explained in #280.

It also removes some of the methods from config that do not exist in Java and uses `Hazelcast Property` instead. 

fixes #280.

### DIFF
--- a/config/property/hazelcast_properties.go
+++ b/config/property/hazelcast_properties.go
@@ -1,0 +1,98 @@
+// Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License")
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package property
+
+import (
+	"os"
+
+	"strings"
+
+	"time"
+
+	"strconv"
+
+	"fmt"
+
+	"github.com/hazelcast/hazelcast-go-client/config"
+)
+
+// HazelcastProperties is a container for configured Hazelcast properties.
+// A Hazelcast property can be set via:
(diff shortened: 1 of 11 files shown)


config/property/hazelcast_property.go
```diff
@@ -0,0 +1,72 @@
...
+// limitations under the License.
+
+package property
+
+import (
+	"time"
+)
+
+// Hazelcast is a struct for client properties.
+type Hazelcast struct {
```
sancar: Isn't the name HazelcastProperty is much more suitable for this struct ?

SaitTalhaNisanci: In go since dotimport is not used, it will be imported as `property.Hazelcast` vs ` property.HazelcastPropery` do you think it looks bad ? 

This is from go package names:

```
Avoid stutter. Since client code uses the package name as a prefix when referring to the package contents, the names for those contents need not repeat the package name. The HTTP server provided by the http package is called Server, not HTTPServer. Client code refers to this type as http.Server, so there is no ambiguity.
```

config/property/hazelcast_properties.go
```diff
@@ -0,0 +1,99 @@
...
+// HazelcastProperties is a container for configured Hazelcast properties.
+// A Hazelcast property can be set via:
+//  * an environmental variable
+//  * programmatic configuration Config.SetProperty
+type HazelcastProperties struct {
+	properties config.Properties
+}
+
+// NewHazelcastPropertiesWithConfig returns HazelcastProperties with the given config's properties.
+func NewHazelcastPropertiesWithConfig(config *config.Config) *HazelcastProperties {
```
sancar: This constructor does not seem really necessary. I think caller should call `NewHazelcastPropertiesWithProperties` , and pass the properties instead of config always.

config/property/hazelcast_property.go
sancar: When I read this source code with type `Hazelcast`, I automatically expect to see methods like  `Hazelcast.newHazelcastInstance/shutdownAll` etc.. 
This is what we are used to see from Java API. 
I get that golang convention is different also. I personally prefer the `HazelcastProperty` instead. 
Lets get more opinion on this ? @asimarslan wdyt ?

PR #284, opened 2018-06-26
add client statistics

### LINKED ISSUE #277: Client Statistics
- [X] Implement Statistics
- [X] Write Tests
- [ ] Sample Code

SaitTalhaNisanci:
This pr implements client statistics. 

Some of the metrics in `os` and `runtime`  are missing. As far as I understand there is no easy way to get them that works for cross platform, there are external O/S metric packages that provide most of the missing metrics.

Fixes #277.


internal/statistics.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+var emptyStatValueFunc = func() string {
+	return emptyStatValue
+}
+
+func (s *statistics) sendStatistics() {
+	ownerConnection := s.getOwnerConnection()
+	if ownerConnection == nil {
+		log.Println("Cannot send client statistics to the server. No owner connection")
```
ihsandemir: will this be printed periodically a number of times during which the client has no cluster connection since this method is called periodically? Is this desired?

internal/statistics.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func (s *statistics) registerMetrics() {
+	var m runtime.MemStats
+	runtime.ReadMemStats(&m)
+	s.registerGauge("os.committedVirtualMemorySize",
+		func() string {
+			return strconv.Itoa(int(m.Sys))
+		})
+	s.registerGauge("os.freePhysicalMemorySize", emptyStatValueFunc)
```
ihsandemir: a lot of statistics is registered as empty stat, why? does "go" have no way to obtain os statistics? If it is empty why are you registering? Better not to register at all.

internal/statistics.go
SaitTalhaNisanci: Currently this will be printed periodically. I am not sure if we should print this periodically, what do you think?

internal/statistics.go
SaitTalhaNisanci: I added some more but I couldnt find some of the statistics.

internal/statistics.go
ihsandemir: Java impl:
            // suppress exception, do not print too many messages
            if (logger.isFinestEnabled()) {
                logger.finest("Could not send stats ", e);
            }

internal/statistics.go
ihsandemir: Do not add if no such statistics, this is misleading as if it were implemented. If the statistics do not exist do not register empty stat. I believe the there should be no registration of empty stat.

SaitTalhaNisanci: I think in Non java clients, missing statistics are added as empty stat. I am not sure what the exact reason for this is though, @asimarslan ?

internal/statistics.go
SaitTalhaNisanci: Currently there is no logging level in the go client, it will be added in `0.5`. I think we can put the lowest level for this when it is implemented. What do you think?

PR #286, opened 2018-06-27
Added PS scripts for enabling to run TS on Windows

olukas:
Hazelcast Go client is currently not supported on Windows. Following powershell scripts provide ability to execute testsuite on Windows which will be useful once Windows will start to be supported. All these scripts are based on similar bash scripts used for testing on linux.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,0 +1,43 @@
+$ClientImportPath="github.com/hazelcast/hazelcast-go-client"
+$ListOfPackages = New-Object System.Collections.ArrayList
+go list $ClientImportPath/... | Select-String -pattern ".*/test|.*/compatibility|.*/rc|.*/sample" -notmatch | %{$ListOfPackages.add($_)}
+$PackageList=$ListOfPackages -join ","
+
+#run linter
+cd $env:gopath\src\$ClientImportPath
+.\linter.ps1
+
+If ($lastExitCode -ne 0) {
+    Exit 1
+}
+
+If (Test-Path -PathType Container "$env:gopath\src\github.com\apache\thrift\") {
+    Write-Host "thrift already exists, not downloading."
+} Else {
+    go get github.com/apache/thrift/lib/go/thrift
+    cd $env:gopath\src\github.com\apache\thrift
+    git fetch --tags --quiet
+    git checkout 0.10.0    
+}
+
+cd $env:gopath\src\$ClientImportPath
+go build
+
+cd $env:gopath
+.\start-rc.ps1
+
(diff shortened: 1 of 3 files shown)


build.ps1
```diff
@@ -0,0 +1,47 @@
+$ClientImportPath="github.com/hazelcast/hazelcast-go-client"
+$ListOfPackages = New-Object System.Collections.ArrayList
+go list $ClientImportPath/... | Select-String -pattern ".*/test|.*/compatibility|.*/rc|.*/sample" -notmatch | %{$ListOfPackages.add($_)}
+$PackageList=$ListOfPackages -join ","
+
+#run linter
+cd $env:gopath\src\$ClientImportPath
+.\linter.ps1
+
+# currently comment out due to a lot of found issues
```
SaitTalhaNisanci: Why does it find a lot of issues? Can you show a few of the found issues?

build.ps1
```diff
@@ -0,0 +1,47 @@
...
+cd $env:gopath\src\$ClientImportPath
+go build
+
+cd $env:gopath
+.\start-rc.ps1
+
+Start-Sleep -s 11
+
+# Run vet tools (Compiler warning plugin)
+go vet $ClientImportPath > vet.txt
```
SaitTalhaNisanci: This is actually not necessary, we already do vet analysis with go metalinter. I will remove this from `build.sh` as well.

build.ps1
olukas: I have already written about them to Furkan. You can see those issues here - https://hazelcast-l337.ci.cloudbees.com/job/golang-client-windows-temp/1/consoleFull in time 19:16:15.

build.ps1
olukas: Ok, thanks, I will remove it

build.ps1
SaitTalhaNisanci: I think I found why it is finding issues. I will `gometalinter.json` for windows.

olukas: Great, thanks!

yuce:
Thanks @olukas for this PR. The latest code for the Go client that runs the tests work with Windows Subsystem for Linux (tested only with WSL 1). Is that a reasonable workaround?

olukas:
Hi @yuce , this PR was basically an attempt how to help Go client team with Windows testing 3 years ago. I guess this PR is currently very obsolete and can be closed (or reworked by someone).

Regarding to your question - I think it can be used as workaround but if Windows OS is supported for Go client then you should probably consider to use also real Windows for testing.

PR #293, opened 2018-07-03
add missing alive check to connection manager

### LINKED ISSUE #288: TestInvocationShouldNotHang_whenClientShutsDown
```
11:05:59 --- FAIL: TestInvocationShouldNotHang_whenClientShutsDown (60.06s)
11:05:59 	assertions.go:29: true != false : invocationNotSent failed
```

This test sometimes fails.

SaitTalhaNisanci:
Fixes #288.

### DIFF
--- a/internal/connection_manager.go
+++ b/internal/connection_manager.go
@@ -106,6 +106,9 @@ func (cm *connectionManagerImpl) onConnectionClose(connection *Connection, cause
 }
 
 func (cm *connectionManagerImpl) getOrTriggerConnect(address core.Address) (*Connection, error) {
+	if cm.isAlive.Load() == false {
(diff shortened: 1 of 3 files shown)


internal/connection_manager.go
```diff
@@ -106,6 +106,9 @@ func (cm *connectionManagerImpl) onConnectionClose(connection *Connection, cause
 }
 
 func (cm *connectionManagerImpl) getOrTriggerConnect(address core.Address) (*Connection, error) {
+	if cm.isAlive.Load() == false {
+		return nil, core.NewHazelcastIllegalStateError("Connection Manager is not active", nil)
```
sancar: I think we should use `HazelcastClientNotActiveError` here

PR #300, opened 2018-07-11
add tls support

### LINKED ISSUE #294: TLS Support with Mutual Authentication
We should have TLS support and mutual authentication for go client.

SaitTalhaNisanci:
This pr adds ssl/tls support to client. SSLConfig has embedded `tls.Config` that can be used to configuressl/ tls configuration.  When dialing server with tls, the embedded `tls.Config` will be used. Currently client does not return any error if enterprise version is not available but user is trying to use TLS, this could be misleading for users.

Also when there is an error while connecting, the error message was being swallowed, now it is returned to connection Manager. 

Missing LifecycleService `IsRunning` method is added too.

 
fixes #294.


config/config.go
```diff
@@ -336,6 +336,9 @@ type NetworkConfig struct {
 
 	// cloudConfig is the config for cloud discovery.
 	cloudConfig *ClientCloud
+
+	// tlsConfig is used for ssl/tls configuration of client.
+	tlsConfig *TLSConfig
```
sancar: Config name should be SSL.

sancar: Basically because we used name SSL in all other clients.

sancar:
Can you put a readme alongside with pem/cert files describing how they are generated or where are they from ?

SaitTalhaNisanci:
I added the readme that explains where they come from. https://github.com/hazelcast/hazelcast-go-client/pull/300/files#diff-df3de6eab8a95957501aa8f7abc0809aR1

ihsandemir:
Can you explain why no error returned to the user?: "Currently client does not return any error if enterprise version is not available but user is trying to use TLS, this could be misleading for users."

config/ssl_config.go
```diff
@@ -0,0 +1,85 @@
...
+// SetCaPath sets CA file path.
+// It returns an error if file cannot be loaded.
+func (sc *SSLConfig) SetCaPath(path string) error {
+	// load CA cert
+	caCert, err := ioutil.ReadFile(path)
+	if err != nil {
+		return err
+	}
+	caCertPool := x509.NewCertPool()
+	caCertPool.AppendCertsFromPEM(caCert)
```
ihsandemir: would this return error if file format is wrong? how do we handle this error?

config/ssl_config.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// AddClientCertAndKeyPath adds client certificate path and client key path to tls config.
+// In order to add multiple client certificate-key pairs one should call this function for each of them.
+// If certificates is empty then no certificate will be sent to
+// the server. If this is unacceptable to the server then it may abort
+// the handshake.
+// For mutual authentication at least one client certificate should be added.
+// It returns an error if any of files cannot be loaded.
+func (sc *SSLConfig) AddClientCertAndKeyPath(clientCertPath string, clientKeyPath string) error {
```
ihsandemir: is the client key public key or private? can you document it? Can you also document what is the file format expected? X.509?

SaitTalhaNisanci: I changed the parameter name as clientPrivateKeyPath also added what the format of files should be. What do you mean by `X.509` ? Are you asking to document what it is or?

config/ssl_config.go
SaitTalhaNisanci: It would return a bool to indicate if it was successfully added or not, we should check it, I will add this check.

SaitTalhaNisanci:
@ihsandemir  I added the check to return an error when it is not enterprise. However currently it is looking if the environment variable `HAZELCAST_ENTERPRISE_KEY` exists. Does this cover all the cases? If not what covers all the cases?

ihsandemir:
@SaitTalhaNisanci For client to work, you do not need the HAZELCAST_ENTERPRISE_KEY environment variable. This variable is needed by the server only, right? Do you want to provide error if the server does not support TLS? That is a different question to solve. The check that you put does not seem to be related.

SaitTalhaNisanci:
@ihsandemir You are right, I will just return what tls library returns to user as we currently have no way of knowing if the server is enterprise from the client.

test/ssl/client_ssl_authentication_test.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	config.NetworkConfig().SSLConfig().Certificates = nil
+	client, err := hazelcast.NewClientWithConfig(config)
+	defer client.Shutdown()
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestSSLOptionalMutualAuthentication_NeitherServerNotClientKnowsTheOther(t *testing.T) {
```
ihsandemir: Neither nor? wording.

test/ssl/client_ssl_authentication_test.go
```diff
@@ -0,0 +1,245 @@
...
+		t.Fatal(err)
+	}
+	_, err = hazelcast.NewClientWithConfig(config)
+	if _, ok := err.(*core.HazelcastIllegalStateError); !ok {
+		t.Error(err)
+	}
+}
+
+func TestSSLOptionalMutualAuthenticationConnect(t *testing.T) {
+	if !test.IsEnterprise() {
```
ihsandemir: Why do you need this check in the tests? I assume that you always need to use the ee jar for server during your tests, what is the exception to this?

SaitTalhaNisanci:
verify

SaitTalhaNisanci:
@ihsandemir  @sancar I added an `enterprise` tag which is used only when `ENTERPRISE_TESTS_ENABLED` exists as environment variable (I also updated the build script and jenkins configuration). So the tests that require enterprise will have this tag and by default they will not be run.

config/ssl_config.go
```diff
@@ -0,0 +1,136 @@
...
+func (sc *SSLConfig) SetCaPath(path string) error {
+	// load CA cert
+	caCert, err := ioutil.ReadFile(path)
+	if err != nil {
+		return err
+	}
+	caCertPool := x509.NewCertPool()
+	if ok := caCertPool.AppendCertsFromPEM(caCert); !ok {
+		return core.NewHazelcastIOError("error while loading the CA file, make sure the path exits "+
+			"the format is pem", nil)
```
mdumandag: small typo, there should be an and here

config/ssl_config.go
```diff
@@ -0,0 +1,136 @@
...
+// the handshake.
+// For mutual authentication at least one client certificate should be added.
+// It returns an error if any of files cannot be loaded.
+func (sc *SSLConfig) AddClientCertAndKeyPath(clientCertPath string, clientPrivateKeyPath string) error {
+	cert, err := tls.LoadX509KeyPair(clientCertPath, clientPrivateKeyPath)
+	if err != nil {
+		return err
+	}
+	if len(sc.Certificates) == 0 {
+		sc.Certificates = []tls.Certificate{cert}
```
mdumandag: is tls.Certificates a nil slice ?

test/ssl/client_ssl_authentication_test.go
```diff
@@ -0,0 +1,226 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = hazelcast.NewClientWithConfig(config)
+	if _, ok := err.(*core.HazelcastIllegalStateError); !ok {
+		t.Error(err)
+	}
+}
+
+func TestSSLMutualAuthentication_ServerDoesntKnowClientFail(t *testing.T) {
```
mdumandag: this test and the test below it are duplicate. for this test server1ca should be used

test/ssl/client_ssl_authentication_test.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+}
+
+func TestSSLMutualAuthentication_NeitherServerNorClientKnowsTheOtherFail(t *testing.T) {
+	clusterID, err := createMemberWithXML(maRequiredXML)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer remoteController.ShutdownCluster(clusterID)
+	config, err := createClientConfigWithSSLConfig("client2-cert.pem",
```
mdumandag: maybe you can change client2-cert.pem with client2Cert for consistency across the tests

test/ssl/client_ssl_authentication_test.go
```diff
@@ -0,0 +1,226 @@
...
+		client2Key, server2CA)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = hazelcast.NewClientWithConfig(config)
+	if _, ok := err.(*core.HazelcastIllegalStateError); !ok {
+		t.Error(err)
+	}
+}
```
mdumandag: there should be a test where you set mutual authentication to optional and provide no client certificates

test/ssl/client_ssl_test.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer remoteController.ShutdownCluster(clusterID)
+
+	cfg := hazelcast.NewConfig()
+	sslCfg := cfg.NetworkConfig().SSLConfig()
+	sslCfg.SetEnabled(true)
+	sslCfg.SetCaPath(server1CA)
+	sslCfg.ServerName = serverName
```
mdumandag: what happens when server name is not set and server provides a certificate with empty common name/subject alternative name fields ? this may be the case for hazelcast cloud

config/ssl_config.go
SaitTalhaNisanci: Yes

PR #314, opened 2018-08-11
enhance godocs in core

SaitTalhaNisanci:
This pr enhances the go docs in core package and fixes some namings.

Lifecycle states are moved to core package from internal so that users can see them.


hazelcast.go
```diff
@@ -87,13 +87,13 @@ type Instance interface {
...
 	Shutdown()
 
-	// GetCluster returns the Cluster this instance is part of.
-	// Cluster interface allows you to add listener for membership
+	// GetClusterService returns the ClusterService this instance is part of.
+	// ClusterService interface allows you to add listener for membership
 	// events and learn more about the cluster this Hazelcast
 	// instance is part of.
-	GetCluster() core.Cluster
+	GetClusterService() core.ClusterService
```
sancar: This naming is not consistent with hazelcast server. I think this should stay as Cluster.

SaitTalhaNisanci: Right, is there a reason why it is called `Cluster` but not `ClusterService` on server side?

sancar: I don't know. There is a ClusterService but it is in an internal package. Public API seems to be named as Cluster.

core/map.go
```diff
@@ -18,16 +18,16 @@ import (
 	"time"
 )
 
-// Map is Hazelcast Map client proxy to access the map on the cluster.
-// Concurrent, distributed, observable and queryable map.
+// Map is concurrent, distributed, observable and queryable map.
```
Serdaro: I think there is no need for the "map" at the end of sentence.

hazelcast.go
```diff
@@ -87,13 +87,13 @@ type Instance interface {
 	// Shutdown shuts down this Instance.
 	Shutdown()
 
-	// GetCluster returns the Cluster this instance is part of.
-	// Cluster interface allows you to add listener for membership
+	// GetCluster returns the ClusterService this instance is part of.
```
sancar: This looks like should not change.

sample/orgwebsite/list_sample.go
```diff
@@ -22,7 +22,7 @@ import (
 
 func listSampleRun() {
 
-	// Start the Hazelcast Client and connect to an already running Hazelcast Cluster on 127.0.0.1
+	// Start the Hazelcast Client and connect to an already running Hazelcast ClusterService on 127.0.0.1
```
sancar: Similar problem. Can you recheck all `Cluster` changes ?

core/lifecycle.go
```diff
@@ -14,13 +14,34 @@
...
+	LifecycleStateConnected = "CONNECTED"
+
+	// LifecycleStateDisconnected is fired when the client disconnected from a member.
+	LifecycleStateDisconnected = "DISCONNECTED"
+
+	// LifecycleStateShuttingDown is fired when the client is shutting down.
+	LifecycleStateShuttingDown = "SHUTTING_DOWN"
+
+	// LifecycleStateShutdown is fired when the client shutdown is completed.
+	LifecycleStateShutdown = "SHUTDOWN"
```
sancar: Is this camel case because of go convention ? 
If there is no convention from go, I think  as capital letters divided by underscore is more appropriate. 
But if this is the way from go world, following go convention makes more sense.

SaitTalhaNisanci: It is the convention in Go.

PR #315, opened 2018-08-11
remove unnecessary method from cluster service

SaitTalhaNisanci:
`GetMemberByUUID ` and `GetMember` should not be in public API. This pr removes them.


core/cluster.go
```diff
@@ -29,7 +29,4 @@ type Cluster interface {
 
 	// GetMember gets the member with the given address.
 	GetMember(address Address) Member
-
-	// GetMemberByUUID gets the member with the given uuid.
-	GetMemberByUUID(uuid string) Member
```
furkansenharputlu: It exists on Java client. See: https://github.com/hazelcast/hazelcast/blob/91604323f07e5cfe373884f41757582046dbb085/hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientClusterService.java#L53

PR #319, opened 2018-08-12
make services internal in client

SaitTalhaNisanci:
This pr makes client services private except two which are used indirectly in tests. They can also be made private after modifying the tests.

It also removes the `Get` prefix from some getters.


sancar:
What is the motivation behind making these fields private ? User can not access them anyway via `Instance` interface. To me they being private or not does not seem to be an issue. 
Was there a specific problem that you are targeting ?

SaitTalhaNisanci:
Currently with casting they can be used. If they are private they cant be used with even casting. And as it implements `Instance` interface and `HazelcastClient` struct has `LifecycleService` method so it cannot have the same field as that conflicts with the method. That is how I realized this issue.

sancar:
I think casting being illegal is clear from the fact that hazelcastClient struct is in internal. 
I am OK with making them private but if one day someone else converts them to public again, I would be OK with that too. Just looking for a strong reason to do so. 
For example, if someone introduces a package called `connection` or (`nio` as in server) and put ConnectionManager, Connection in it most of these needs to be public again. Otherwise, ConnectionManager can not access them.

SaitTalhaNisanci:
Yes that is true, what about the method name and variable name conflict? Do you think in that case we should find another variable name as method name should not change I guess?

PR #323, opened 2018-08-16
fix bug in getTimeInMillisecond

### LINKED ISSUE #322: GetTimeInMilliSeconds with time less than a millisecond
In java client when time duration is less than a millisecond this method returns 1. We return `time.Millisecond` which is `1000*1000` as `int64`. This means that instead of `1 millisecond` we return `1000000 milliseconds`.

SaitTalhaNisanci:
Fixes #322.


internal/timeutil/util.go
```diff
@@ -20,11 +20,9 @@ import (
 )
 
 func GetTimeInMilliSeconds(duration time.Duration) int64 {
-	if duration == -1 {
```
sancar: There are places that -1 is passed here. Removing this `if` does not seem correct

PR #330, opened 2018-08-28
add cloud discovery tests

SaitTalhaNisanci:
This PR adds tests for cloud discovery.

When using hazelcast cloud discovery SSL/TLS should be used, it is also added in this PR however since TLS PR hasnt yet merged, that argument is set as nil.


SaitTalhaNisanci:
verify

PR #342, opened 2018-09-06
implement reliable topic with tests

### LINKED ISSUE #335: Support for reliable topics
Go version:1.10.3
Hazelcast Go Client version: 0.3
Hazelcast server version: 3.10.0
Number of the clients: 1
Cluster size, i.e. the number of Hazelcast cluster members: 1
OS version : Linux

#### Expected behaviour
I can subscribe to a reliable topic

#### Actual behaviour
No such API exists

I've tried this as a workaround
`topic, err := client.GetDistributedObject("hz:impl:reliableTopicService", topicName)`
which fails with 
`no factory registered for service: hz:impl:reliableTopicService`

SaitTalhaNisanci:
This pr implements reliable topic which is backed by ringbuffer.

fixes #335.


internal/reliable_topic.go
```diff
@@ -0,0 +1,285 @@
...
+		msgProcessors:   make(map[string]*messageProcessor),
+		msgProcessorsMu: new(sync.Mutex),
+		proxy: &proxy{
+			client:      client,
+			serviceName: serviceName,
+			name:        name,
+		},
+	}
+	proxy.serializationService = client.SerializationService
+	// TODO:: get reliable topic config with ringbuffer id
```
sancar: This seems important for this feature. I think, It should be handled with this pr.

test/proxy/reliabletopic/reliable_topic_test.go
```diff
@@ -0,0 +1,359 @@
...
+	msg := item.(*reliabletopic.Message)
+	obj, _ := client.(*internal.HazelcastClient).SerializationService.ToObject(msg.Payload().(*serialization.Data))
+	assert.Equal(t, nil, obj, int64(10))
+}
+
+func TestReliableTopicProxy_Overwrite(t *testing.T) {
+	reliableTopic, _ := client.GetReliableTopic("overwrite")
+	topic := reliableTopic.(*internal.ReliableTopicProxy)
+	items := generateItems(client.(*internal.HazelcastClient), 10)
+	topic.Ringbuffer().AddAll(items, core.OverflowPolicyFail)
```
sancar: Why don't we use topic.publish for first 10 items ?

SaitTalhaNisanci: For this test I think it can be used but what would be the advantage?

internal/reliable_topic.go
SaitTalhaNisanci: I think I forgot to remove the TODO, because we already get the reliable config with its name which is its id.

test/proxy/reliabletopic/reliable_topic_test.go
sancar: We are testing the ReliableTopic here. We should use RingBuffer only when necessary. Otherwise it looks confusing to the reader.

config/config.go
```diff
@@ -72,6 +76,16 @@ func (cc *Config) SecurityConfig() *SecurityConfig {
 	return cc.securityConfig
 }
 
+// ReliableTopicConfig returns the reliable topic config for this client.
+func (cc *Config) ReliableTopicConfig(name string) *ReliableTopicConfig {
```
sancar: Naming is not consistent with `GetFlakeIDGeneratorConfig` 
Which one is correct ?

sancar: I think implementation also should be similar. Using the `default` when not found is missing here.

SaitTalhaNisanci: We can use `GetReliableTopicConfig` since it is not a getter. This one creates a new one with the given name when it is not found.  It is the same as Java Client.

sancar: Oh, java client also does not use `default`. But on the member config we use `default` when given name is not found. Can you open an issue for java client also to fix this ?
And can you also update this so that it behaves similar to current member ?
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/config/Config.java#L1881

internal/reliable_topic.go
```diff
@@ -0,0 +1,284 @@
...
+	} else {
+		m.onFailure(err)
+	}
+}
+
+func (m *messageProcessor) onFailure(err error) {
+	if m.cancelled.Load() == true {
+		return
+	}
+	if strings.Contains(err.Error(), "com.hazelcast.ringbuffer.StaleSequenceException") {
```
ihsandemir: string comparison? isn't there some kind of error code to compare?

internal/reliable_topic.go
```diff
@@ -0,0 +1,284 @@
...
+				strconv.Itoa(int(m.sequence)) + " " +
+				" to new sequence " + strconv.Itoa(int(headSeq))
+			log.Println(msg)
+			m.sequence = headSeq
+			go m.next()
+			return
+		}
+
+	}
+	msg := "Terminating Message Listener: " + m.id + " on topic: " + m.proxy.name + ". Reason: " + err.Error()
```
ihsandemir: Please change to the same message as in Java.

internal/reliable_topic.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+}
+
+func (m *messageProcessor) onFailure(err error) {
+	if m.cancelled.Load() == true {
+		return
+	}
+	if strings.Contains(err.Error(), "com.hazelcast.ringbuffer.StaleSequenceException") {
+		if m.listener.IsLossTolerant() {
+			headSeq, _ := m.proxy.ringBuffer.HeadSequence()
```
ihsandemir: outside if at Java, used in the log statement after the if statement.

internal/reliable_topic.go
```diff
@@ -0,0 +1,284 @@
...
+			msg := "Topic " + m.proxy.name + " ran into a stale sequence. Jumping from old sequence " +
+				strconv.Itoa(int(m.sequence)) + " " +
+				" to new sequence " + strconv.Itoa(int(headSeq))
+			log.Println(msg)
+			m.sequence = headSeq
+			go m.next()
```
ihsandemir: This is sync in Java. Inside the next method there is async readMany and andThen method, hence we better go with what is at Java.

internal/reliable_topic.go
```diff
@@ -0,0 +1,284 @@
...
+			return
+		}
+
+	}
+	msg := "Terminating Message Listener: " + m.id + " on topic: " + m.proxy.name + ". Reason: " + err.Error()
+	log.Println(msg)
+	m.proxy.msgProcessorsMu.Lock()
+	m.cancel()
+	m.proxy.msgProcessorsMu.Unlock()
+}
```
ihsandemir: Missing handling of other exceptions handled at Java such as HazelcastInstanceNotActiveException, DistributedObjectDestroyedException, HazelcastClientNotActiveException || t instanceof RejectedExecutionException.

internal/reliable_topic.go
```diff
@@ -0,0 +1,284 @@
...
+
+func (m *messageProcessor) onResponse(readResults core.ReadResultSet) {
+	// We process all messages in batch. So we don't release the thread and reschedule ourselves;
+	// but we'll process whatever was received in 1 go.
+	var i int32
+	for ; i < readResults.Size(); i++ {
+		item, err := readResults.Get(i)
+		if msg, ok := item.(*reliabletopic.Message); ok && err == nil {
```
ihsandemir: this check should go after  m.cancelled.Load() == true check.

internal/reliable_topic.go
```diff
@@ -0,0 +1,284 @@
...
+			if m.cancelled.Load() == true {
+				return
+			}
+			m.listener.StoreSequence(m.sequence)
+			m.process(msg)
```
ihsandemir: error handling missing? java side catches exception.

internal/reliable_topic.go
```diff
@@ -0,0 +1,284 @@
...
+)
+
+const (
+	topicRBPrefix  = "_hz_rb_"
+	maxBackoff     = 2000 * time.Millisecond
+	initialBackoff = 100 * time.Millisecond
+)
+
+type ReliableTopicProxy struct {
+	*proxy
```
ihsandemir: missing method boolean terminate(Throwable failure)

config/config.go
```diff
@@ -72,6 +76,22 @@ func (cc *Config) SecurityConfig() *SecurityConfig {
 	return cc.securityConfig
 }
 
+// GetReliableTopicConfig returns the reliable topic config for this client.
+func (cc *Config) GetReliableTopicConfig(name string) *ReliableTopicConfig {
+	if cfg, found := cc.reliableTopicConfigMap[name]; found {
+		return cfg
+	}
+	_, found := cc.reliableTopicConfigMap["default"]
```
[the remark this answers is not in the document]
SaitTalhaNisanci: @sancar said that this should be similar to flakeID and it should also be changed in Java.

config/reliable_topic_config.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// defaultReadBatchSize is the default read batch size.
+	defaultReadBatchSize = 10
+
+	// defaultTopicOverloadPolicy is the default slow consumer policy.
+	defaultTopicOverloadPolicy = core.TopicOverLoadPolicyBlock
+)
+
+// ReliableTopicConfig contains the ReliableTopic configuration for a client.
+type ReliableTopicConfig struct {
```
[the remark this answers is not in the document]
SaitTalhaNisanci: In go executor doesnt fit to its structure. That is why I didnt add it. If we want to get the same functionality we will need to add additional things, which is not necessary now as far as I see.

core/topic_overload_policy.go
```diff
@@ -0,0 +1,51 @@
...
+type TopicOverloadPolicy int
+
+const (
+	// TopicOverLoadPolicyDiscardOldest is a policy so that a message that has not expired can be overwritten.
+	// No matter the retention period set, the overwrite will just overwrite the item.
+	//
+	// This can be a problem for slow consumers because they were promised a certain time window to process messages. But it will
+	// benefit producers and fast consumers since they are able to continue. This policy sacrifices the slow producer in favor
+	// of fast producers/consumers.
+	TopicOverLoadPolicyDiscardOldest TopicOverloadPolicy = iota
```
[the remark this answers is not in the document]
SaitTalhaNisanci: In java the enumarator has a name `TopicOverloadPolicy` so accessing this would be `TopicOverloadPolicy.DISCARD_OLDEST`. Since it is not possible to name enum in go instead of `DiscardOldest`, `TopicOverLoadPolicyDiscardOldest ` seemed better.

internal/cluster.go
```diff
@@ -361,6 +361,9 @@ func (cs *clusterService) GetMembersWithSelector(selector core.MemberSelector) (
 }
 
 func (cs *clusterService) GetMember(address core.Address) core.Member {
+	if address == nil {
```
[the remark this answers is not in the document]
SaitTalhaNisanci: When client publishes a message, publisher address is nil.`https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientReliableTopicProxy.java#L103`

internal/reliable_topic.go
SaitTalhaNisanci:
```
        void next() {
            if (cancelled) {
                return;
            }

            ICompletableFuture<ReadResultSet<ReliableTopicMessage>> f
                    = ringbuffer.readManyAsync(sequence, 1, config.getReadBatchSize(), null);
            f.andThen(this, executor);
        }
```

isnt this async ? Doesnt the method register the callback and continue immediately ?

internal/reliable_topic.go
SaitTalhaNisanci: What can return an error here?  As far as I see it is also not necessary in Java.

internal/reliable_topic.go
SaitTalhaNisanci: We can get that information from the stack trace but currently there is no way of accessing the parts of stack trace and the whole stack trace is a single string. I will create an issue for this so that we can get individual parts of stack trace.

internal/reliable_topic.go
SaitTalhaNisanci: I guess since user code is used in processing the message they can throw an exception there. In go `OnMessage` doesnt return an error. I will update it.

config/config.go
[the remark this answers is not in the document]
ihsandemir: Note: Opened issue for Java : https://github.com/hazelcast/hazelcast/issues/13880

core/topic_overload_policy.go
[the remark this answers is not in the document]
ihsandemir: In C++, it is the same. I go around it with a class named TopicOverLoadPolicy with an unnamed enum DISCARD_OLDEST. You may do whichever way you want, just easier for the user with shorter names.

PR #344, opened 2018-09-09
use serialization as a separate service through interfaces

SaitTalhaNisanci:
This PR adds serialization interface and moves all serialization related code to serialization package. 
It changes all `*serialization.Data` structs with `serialization.Data` interface. Serialization service should not depend on any other package, and to have a separation of services principle, serialization service is only accessed through interfaces.  Registering of factories related to hazelcast internal are moved to hazelcast internal. 

Serialization's config is moved to serialization package. Now it is pretty much like an independent service. 

Codecs are also updated from `*serialization.Data` struct to `serialization.Data` interface.  

Currently the serialization service interface has only two methods:
```
// SerializationService is used to serialize user objects to Data and deserialize data to objects.
type SerializationService interface {
	// ToObject deserializes the given data to an object.
	// It can safely be called on an object that is already deserialized. In that case, that instance
	// is returned.
	// If this is called with nil, nil is returned.
	ToObject(data serialization.Data) (interface{}, error)
	// ToData serializes an object to a data.
	// It can safely be called with a Data. In that case, that instance is returned.
	// If it is called with nil, nil is returned.
	ToData(object interface{}) (serialization.Data, error)
}
```

The new package's structure is as follows :
```
serialization/
  api.go (public api)
  config.go (public config)
  internal/ (this contains the implementations and since its name is internal it cannot be imported)
  spi/ (what client uses but users shouldnt use is here.)
```


furkansenharputlu:
In the previous implementation there was no problem for the user. Everything was clear. The user could use the features he wanted to use by looking at the serialization package. In addition, this change is inversely in line with the internal-core logic. Serialization implementations that are not relevant to the user will appear to the user so I think that this change is not consistent with our approach up to now.

sancar:
We need to make it clear that `serialization` is not part of public API. 
For example, in core side, `serialization` package is under `spi`

config/config.go
```diff
@@ -167,7 +168,7 @@ func (cc *Config) SetNetworkConfig(networkConfig *NetworkConfig) {
 	cc.networkConfig = networkConfig
 }
 
-// SetSerializationConfig sets the SerializationConfig.
-func (cc *Config) SetSerializationConfig(serializationConfig *SerializationConfig) {
+// SetSerializationConfig sets the Config.
```
sancar: Be careful with the javadocs when renaming. This does not sound correct.

serialization/internal/class_definition_builder.go
```diff
@@ -1,11 +1,9 @@
-// Package classdef is used to build and register class definitions manually.
-package classdef
+package internal
```
sancar: I believe this class was part of public API and should remain so.

sancar:
`bufferUtil` was private API. Is that intentinally moved to public API. I think it should be reside in `serialization/internal`

furkansenharputlu:
I'm not in favor of making a change that will not provide extra features or convenience. Also, as I said before, it is incompatible with internal-core logic. Serialization may not be dependent on other services but it is the fact that it is specific to hazelcast so its internal part should be in the `internal` package.

ihsandemir:
@SaitTalhaNisanci Can you describe why you can not do this (or can you do it at all)?

move serialization.internal to internal.serialization 

as @furkansenharputlu suggested.

SaitTalhaNisanci:
The idea is that `serialization/internal` should not be used anywhere but in serialization. The only way to access the serialization should be through `serialization service` that is why I believe it should be in `serialization/internal` not `internal/serialization`.

PR #352, opened 2018-09-21
give server error to users in ServerError interface

### LINKED ISSUE #349: Stack Trace from server
Currently when server sends the stack trace for an error it is returned to user as a single string. Looking up information in the stack trace requires string comparison. There should be a way so that users can access parts of stack trace such as error code.

SaitTalhaNisanci:
When server returns an error client was returning it as a string. When they need to look up for a certain thing in the error message they needed to do string comparison. This pr adds `ServerError` interface, which has error `className`, `errorCode`, `stackTrace` etc.  

In order to get ServerError user will call `ServerError` Method which does the following:

```
func (e *HazelcastErrorType) ServerError() ServerError {
	if serverErr, ok := e.cause.(ServerError); ok {
		return serverErr
	}
	return nil
}
```

Fixes #349.

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -274,3 +274,47 @@ type LoadBalancer interface {
 	// It returns nil if no member is available.
 	Next() Member
 }
+
+// StackTraceElement contains stacktrace information for server side exception.
+type StackTraceElement interface {
+	// DeclaringClass returns the fully qualified name of the class containing
+	// the execution point represented by the stack trace element.
+	DeclaringClass() string
+
+	// MethodName returns the name of the method containing the execution point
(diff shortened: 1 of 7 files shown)


sancar:
How about a PrintStackTrace method on ServerError as we have on java. 
I think common use case is printing the error. Having access to stack-trace line by line will not mean much to users.  @asimarslan @SaitTalhaNisanci

SaitTalhaNisanci:
@sancar Printing error functionality already exits with `Error` method. This pr adds the functionality which enables accessing parts of the error without having to look at all the error. For example users might want to take certain action in certain errors.

PR #373, opened 2018-10-31
Reference manual

sancar:


README.md
```diff
@@ -1,144 +1,1607 @@
...
+```
+
+- `<group>`: Specifies which cluster this member belongs to. A member connects only to other members that are in the same group as
+itself. You will see `<name>` and `<password>` tags with some pre-configured values. You may give your clusters different names so that they can
+live in the same network without disturbing each other. Note that the cluster name should be the same across all members and clients that belong
+ to the same cluster. `<password>` tag is not in use since Hazelcast 3.9. It is there for backward compatibility
+purposes. You can remove or leave it as it is if you use Hazelcast 3.9 or later.
+- `<network>`
+    - `<port>`: Specifies the port number to be used by the member when it starts. Its default value is 5701 You can specify another port number, and if
+     you set `auto-increment` to `true`, than Hazelcast will try subsequent ports until it finds an available port or `port-count` is reached.
```
SaitTalhaNisanci: it should be `then` instead of `than`.

README.md
```diff
@@ -1,144 +1,1607 @@
...
+
+This section describes some network configuration settings to cover common use cases in connecting the client to a cluster. Refer to [Configuration Overview](#configuration-overview)
+and the following sections for information about detailed network configuration and/or additional features of Hazelcast Go client configuration.
+
+An easy way to configure your Hazelcast Go client is to create a `Config` object and set the appropriate options. Then you can
+supply this object to your client at the startup.
+
+**Configuration**
+
+You need to create a `ClientConfig` object and adjust its properties. Then you can pass this object to the client when starting it.
```
SaitTalhaNisanci: It is `Config` in Go. We should do replacement with `ClientConfig` -> `Config`

README.md
```diff
@@ -1,144 +1,1607 @@
...
+}
+```
+---
+
+If you run Hazelcast IMDG members in a different server than the client, you most probably have configured the members' ports and cluster
+names as explained in the previous section. If you did, then you need to make certain changes to the network settings of your client.
+
+### Group Settings
+
+```
```
SaitTalhaNisanci: It could be better to use go snippet here.

README.md
```diff
@@ -1,144 +1,1607 @@
...
+config := hazelcast.NewConfig()
+config.GroupConfig().SetName("GROUP_NAME_OF_YOUR_CLUSTER")
+```
+> **NOTE: If you have a Hazelcast release older than `3.11`, you need to provide also a group password along with the group name.**
+
+### Network Settings
+
+You need to provide the ip address and port of at least one member in your cluster so the client finds it.
+
+```
```
SaitTalhaNisanci: missing `go`.

README.md
```diff
@@ -1,144 +1,1607 @@
...
+func (*address) FactoryID() int32 {
+	return 1
+}
+
+func (*address) ClassID() int32 {
+	return 1
+}
+```
+
+IdentifiedDataSerializable uses `ClassID()` and `FactoryID()` to reconstitute the object. To complete the implementation `IdentifiedDataSerializableFactory` should also be implemented and registered into `SerializationConfig` which can be accessed from `config.SerializationConfig()`. The factory's responsibility is to return an instance of the right `IdentifiedDataSerializable` object, given the class id.
```
SaitTalhaNisanci: it is used sometimes as `ClassID()` and sometimes as `ClassID`.

SaitTalhaNisanci: class id -> `classID`

README.md
```diff
@@ -1,144 +1,1607 @@
...
+config := hazelcast.NewConfig()
+networkConfig := config.NetworkConfig()
+networkConfig.SetConnectionTimeout(6 * time.Second)
+```
+
+Its default value is `5000` milliseconds.
+
+## 5.5. Setting Connection Attempt Limit
+
+While the client is trying to connect initially to one of the members in the `ClientNetworkConfig.addresses`, that member might not be available at that moment. Instead of giving up, throwing an error and stopping the client, the client will retry as many as `ClientNetworkConfig.connectionAttemptLimit` times. This is also the case when the previously established connection between the client and that member goes down.
```
SaitTalhaNisanci: `ClientNetworkConfig` -> `NetworkConfig` in all places.

README.md
```diff
@@ -1,144 +1,1607 @@
...
+config := hazelcast.NewConfig()
+networkConfig := config.NetworkConfig()
+networkConfig.SetConnectionAttemptLimit(5)
+```
+
+Its default value is `2`.
+
+## 5.6. Setting Connection Attempt Period
+
+Connection timeout period is the duration in milliseconds between the connection attempts
```
SaitTalhaNisanci: missing period.

README.md
```diff
@@ -1,144 +1,1607 @@
...
+## 5.7. Enabling Client TLS/SSL
+
+You can use TLS/SSL to secure the connection between the clients and members. If you want TLS/SSL enabled
+for the client-cluster connection, you should set an SSL configuration. Please see [TLS/SSL section](#1-tlsssl).
+
+As explained in the [TLS/SSL section](#1-tlsssl), Hazelcast members have key stores used to identify themselves (to other members) and Hazelcast Go clients have certificate authorities used to define which members they can trust. Hazelcast has the mutual authentication feature which allows the Go clients also to have their private keys and public certificates and members to have their certificate authorities so that the members can know which clients they can trust. Please see the [Mutual Authentication section](#13-mutual-authentication).
+
+## 5.8. Enabling Hazelcast Cloud Discovery
+
+The purpose of Hazelcast Cloud Discovery is to provide clients to use IP addresses provided by `hazelcast orchestrator`. To enable Hazelcast Cloud Discovery, specify a token for the `discoveryToken` field and set the `enabled` field to `true`.
```
SaitTalhaNisanci: I think we should talk about Hazelcast Cloud here, or maybe put some links and the first sentence should be changed to something like : `The purpose of Hazelcast Cloud Discovery is to discover cloud IP addresses provided the hazelcast orchestrator`

README.md
```diff
@@ -1,144 +1,1607 @@
...
+cloudConfig.SetEnabled(true)
+```
+
+To be able to connect to the provided IP addresses, you should use secure TLS/SSL connection between the client and members. Therefore, you should set an SSL configuration as described in the previous section.
+
+# 6. Securing Client Connection
+
+This chapter describes the security features of Hazelcast Go client. These include using TLS/SSL for connections between members and between clients and members and mutual authentication. These security features require **Hazelcast IMDG Enterprise** edition.
+
+### 6.1. TLS/SSL
```
SaitTalhaNisanci: I think it should be SSL/TLS.

README.md
```diff
@@ -1,144 +1,1607 @@
...
+
+#### 6.1.2. TLS/SSL for Hazelcast Go clients
+
+Hazelcast Go clients which support TLS/SSL should have the following user supplied SSLConfig
+
+```go
+config := hazelcast.NewConfig()
+sslConfig := config.NetworkConfig().SSLConfig()
+sslConfig.SetEnabled(true)
+sslConfig.SetCaPath("yourCaPath")
```
SaitTalhaNisanci: We should also add `sslConfig.ServerName="serverName"`

README.md
```diff
@@ -1,144 +1,1607 @@
...
+```
+
+#### 6.1.3. Mutual Authentication
+
+As explained above, Hazelcast members have key stores used to identify themselves (to other members) and Hazelcast clients have trust stores used to define which members they can trust.
+
+Using mutual authentication, the clients also have their key stores and members have their trust stores so that the members can know which clients they can trust.
```
SaitTalhaNisanci: `they can trust to.`

README.md
```diff
@@ -1,144 +1,1607 @@
...
+
+In the unisocket client mode, the client will only connect to one of the configured addresses. This single member will behave as a gateway to the other members. For any operation requested from the client, it will redirect the request to the relevant member and return the response back to the client returned from this member.
+
+## 7.3. Handling Failures
+
+There are two main failure cases you should be aware of, and configurations you can perform to achieve proper behavior.
+
+### 7.3.1. Handling Client Connection Failure
+
+While the client is trying to connect initially to one of the members in the `ClientNetworkConfig.SetAddresses`, all the members might be not available. Instead of giving up, throwing an error and stopping the client, the client will retry as many as `connectionAttemptLimit` times.
```
SaitTalhaNisanci: `ClientNetworkConfig.SetAddresses` -> `NetworkConfig.Addresses`

SaitTalhaNisanci: returning an error

SaitTalhaNisanci: `as many times as connectionAttemptLimit`

README.md
```diff
@@ -1,144 +1,1607 @@
...
+
+You can configure `connectionAttemptLimit` for the number of times you want the client to retry connecting. Please see [Setting Connection Attempt Limit](#5-setting-connection-attempt-limit).
+
+The client executes each operation through the already established connection to the cluster. If this connection(s) disconnects or drops, the client will try to reconnect as configured.
+
+### 7.3.2. Handling Retry-able Operation Failure
+
+While sending the requests to related members, operations can fail due to various reasons. Read-only operations are retried by default. If you want to enable retry for the other operations, you can set the `redoOperation` to `true`. Please see [Enabling Redo Operation](#3-enabling-redo-operation).
+
+You can set a timeout for retrying the operations sent to a member. This can be provided by using the property `hazelcast.client.invocation.timeout.seconds` in `config.SetProperty`. The client will retry an operation within this given period, of course, if it is a read-only operation or you enabled the `redoOperation` as stated in the above paragraph. This timeout value is important when there is a failure resulted by either of the following causes:
```
SaitTalhaNisanci: `as stated before` seems better instead of `as stated in the above paragraph`

README.md
```diff
@@ -1,144 +1,1607 @@
...
+
+- Member throws an exception.
+
+- Connection between the client and member is closed.
+
+- Client’s heartbeat requests are timed out.
+
+When a connection problem occurs, an operation is retried if it is certain that it has not run on the member yet or if it is idempotent such as a read-only operation, i.e., retrying does not have a side effect. If it is not certain whether the operation has run on the member, then the non-idempotent operations are not retried. However, as explained in the first paragraph of this section, you can force all client operations to be retried (`redoOperation`) when there is a connection failure between the client and member. But in this case, you should know that some operations may run multiple times causing conflicts. For example, assume that your client sent a `queue.offer` operation to the member, and then the connection is lost. Since there will be no response for this operation, you will not now whether it has run on the member or not. If you enabled `redoOperation`, it means this operation may run again, which may cause two instances of the same object in the queue.
```
SaitTalhaNisanci: `you will not now` -> `you will not know whether`

README.md
```diff
@@ -1,144 +1,1607 @@
...
+
+### 7.5.2. Distributed Data Structure Events
+
+You can add event listeners to the Distributed Data Structures.
+
+#### 7.5.2.1. Listening for Map Events
+
+You can listen to map-wide or entry-based events. To listen to these events, you need to implement the relevant interfaces.
+
+An entry-based  event is fired after the operations that affect a specific entry. For example, `IMap.put()`, `IMap.remove()` or `IMap.evict()`. An `EntryEvent` object is passed to the listener function.
```
SaitTalhaNisanci: All `Imap` should be replaced with `Map`.

PR #382, opened 2018-11-12
try to connect cluster even if the error is authentication error

SaitTalhaNisanci:
When client gets an authentication error while trying to connect to cluster
it stops trying connecting. This behavior does not exist in Java and it should
keep trying connecting.

### DIFF
--- a/test/cluster_test.go
+++ b/test/cluster_test.go
@@ -183,9 +183,7 @@ func TestAuthenticationWithWrongCredentials(t *testing.T) {
 	config.GroupConfig().SetName("wrongName")
 	config.GroupConfig().SetPassword("wrongPassword")
 	client, err := hazelcast.NewClientWithConfig(config)
-	if _, ok := err.(*core.HazelcastAuthenticationError); !ok {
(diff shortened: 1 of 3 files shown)


test/cluster_test.go
```diff
@@ -241,6 +239,22 @@ func TestReconnectToNewNodeViaLastMemberList(t *testing.T) {
...
 
+func TestClusterScaleDown(t *testing.T) {
+	cluster, _ = remoteController.CreateCluster("", DefaultServerConfig)
+	member1, _ := remoteController.StartMember(cluster.ID)
+	remoteController.StartMember(cluster.ID)
+	client, _ := hazelcast.NewClient()
+	assert.Len(t, client.Cluster().GetMembers(), 2)
+
+	remoteController.ShutdownMember(cluster.ID, member1.UUID)
+	time.Sleep(3 * time.Second)
```
sancar: This test is time sensitive.
Instead of sleep, we should either wait the shutdown event via a listener
or we should wait for next assert to be True eventually.

PR #393, opened 2018-11-21
Add logging levels

### LINKED ISSUE #392: Log Levels
In Go's builtin log there is no log levels. We should have log levels and if possible we should support custom and third party loggers. 

- [ ] Log Levels
- [ ] Default Logger
- [ ] Custom Logger
- [ ] No Logger
- [ ] Documentation
- [ ] Sample

SaitTalhaNisanci:
This PR adds:

- Log levels (debug, info, warn etc)
- Default Logger implementation
- Custom Logger support

Fixes #392.


README.md
```diff
@@ -2048,6 +2049,214 @@ config.SetProperty(property.StatisticsPeriodSeconds.Name(), "4")
...
+l := logger.New()
+l.SetPrefix("myPrefix ")
+config := hazelcast.NewConfig()
+config.LoggerConfig().SetLogger(l)
+``` 
+
+The same log message will now be as follows:
+
+```
+myPrefix 2018/11/30 17:55:40 github.com/hazelcast/hazelcast-go-client/internal.(*lifecycleService).fireLifecycleEvent
```
ihsandemir: I thought that this prefix will go before the actual message "New State :  CONNECTED", don't you think so?

README.md
```diff
@@ -2048,6 +2049,214 @@ config.SetProperty(property.StatisticsPeriodSeconds.Name(), "4")
...
+	*zap.Logger
+}
+
+func (z *zapLogger) Debug(args ...interface{}) {
+	message := fmt.Sprintln(args)
+	z.Logger.Debug(message)
+}
+
+func (z *zapLogger) Trace(args ...interface{}) {
+	// NO OP
```
ihsandemir: ? if debug can be printed why suppress trace messages? you can do `Debug(args);` call.

README.md
```diff
@@ -2048,6 +2049,214 @@ config.SetProperty(property.StatisticsPeriodSeconds.Name(), "4")
...
+
+```go
+opt := zap.AddCallerSkip(1)
+logger, _ := zap.NewProduction(opt)
+zapLogger := &zapLogger{logger}
+config := hazelcast.NewConfig()
+config.LoggerConfig().SetLogger(zapLogger)
+```
+
+Note that we call `zap.AddCallerSkip(1)` to skip our wrapper method.
```
ihsandemir: can you explain the reason for this?

README.md
```diff
@@ -2048,6 +2049,214 @@ config.SetProperty(property.StatisticsPeriodSeconds.Name(), "4")
...
+
+
+#### Glog
+
+```go
+type gLogger struct {
+}
+
+func (*gLogger) Debug(args ...interface{}) {
+	// NO OP
```
ihsandemir: ? supressing?

core/logger/logger_test.go
```diff
@@ -0,0 +1,52 @@
...
+package logger
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsValidLogLevel(t *testing.T) {
+	logLevel := "debug"
```
ihsandemir: if testing for debug, you better test for all the other levels as well.

internal/client.go
```diff
@@ -171,16 +173,40 @@ func (c *HazelcastClient) LifecycleService() core.LifecycleService {
...
+}
+
+func (c *HazelcastClient) initLogger() {
+	setLogger := c.Config.LoggerConfig().Logger()
+	logLevel := c.getLogLevel()
+	versionMessage := c.name + " [" + c.Config.GroupConfig().Name() + "]" + " [" + ClientVersion + "] "
+	if setLogger == nil {
+		l := logger.New()
+		l.Level = logLevel
+		l.SetVersionMessage(versionMessage)
```
ihsandemir: SetVersionMessage may be named better since it has mor information than the version.

internal/client.go
```diff
@@ -171,16 +173,40 @@ func (c *HazelcastClient) LifecycleService() core.LifecycleService {
...
 	addressTranslator, err := c.createAddressTranslator()
 	if err != nil {
 		return err
 	}
-	c.id = c.nextClientID()
-	c.initClientName()
-
 	c.credentials = c.initCredentials(c.Config)
-	c.lifecycleService = newLifecycleService(c.Config)
+	c.lifecycleService = newLifecycleService(c.Config, c)
```
ihsandemir: why pass both c and c.Config?

internal/connection.go
```diff
@@ -211,7 +219,7 @@ func (c *Connection) close(err error) {
 	if !atomic.CompareAndSwapInt32(&c.status, 0, 1) {
 		return
 	}
-	//TODO log close message
+	c.logger.Debug("Connection :", c, " closed, err: ", err)
```
ihsandemir: you are logging different messages from what Java logs (both content and number of messages), we need to log at least for info level and above the ones that Java client logs.

internal/lifecycle.go
```diff
@@ -83,7 +86,7 @@ func (ls *lifecycleService) fireLifecycleEvent(newState string) {
 			listener.(core.LifecycleListener).LifecycleStateChanged(newState)
 		}
 	}
-	log.Println("New State : ", newState)
+	ls.logger.Info("New State : ", newState)
```
ihsandemir: See the same method in Java and look at the content of the message, this content is different. It is better all clients logs the same information so that it is easier for the user to trace what is going on.

test/logger/custom_logger_test.go
```diff
@@ -0,0 +1,56 @@
...
+	c.logged = true
+}
+
+func (c *customLogger) Warn(args ...interface{}) {
+}
+
+func (c *customLogger) Error(args ...interface{}) {
+}
+
+func TestCustomLogger(t *testing.T) {
```
ihsandemir: custom logger prints nothing, but did you verify it? Also more tests can be added!!!

test/logger/logger_test.go
```diff
@@ -0,0 +1,90 @@
...
+	remoteController.StartMember(cluster.ID)
+	l := logger.New()
+	buf := new(bytes.Buffer)
+	l.SetOutput(buf)
+	config := hazelcast.NewConfig()
+	config.LoggerConfig().SetLogger(l)
+	client, err := hazelcast.NewClientWithConfig(config)
+	assert.NoError(t, err)
+	client.Shutdown()
+	assert.NotZero(t, buf.Len())
```
ihsandemir: also verify the string content is in expected format.

test/logger/logger_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"bytes"
+
+	"github.com/hazelcast/hazelcast-go-client"
+	"github.com/hazelcast/hazelcast-go-client/config/property"
+	"github.com/hazelcast/hazelcast-go-client/core/logger"
+	"github.com/hazelcast/hazelcast-go-client/rc"
+	"github.com/hazelcast/hazelcast-go-client/test"
+	"github.com/stretchr/testify/assert"
+)
+
```
ihsandemir: missing tests: Verify that when a level is chosen all levels above are printed and below are not.

README.md
SaitTalhaNisanci: This is the default behavior of built in logger. The method `SetPrefix` belongs to the built in logger. So I think this method should not be changed as users expect it to add the prefix to the beginning of each line. If we want we can add a new method that adds a prefix just before the actual message.

README.md
SaitTalhaNisanci: Let's say we have a method x:

```
func x()  {
    logger.Info(..)
}
```

If we expand `logger.Info()` in `x()`, we have:

```
func (z *zapLogger) Info(args ...interface{}) {
 	message := fmt.Sprintln(args)
 	z.Logger.Info(message)
 }
```

So `z.Logger.Info(message)` will be called from our wrapper method, which means that the caller of `z.Logger.Info(message)` is: `func (z *zapLogger) Info(args ...interface{})`. However we want it to be `x`, so we will skip one caller, which is `func (z *zapLogger) Info(args ...interface{})`, and the caller will be `x`.

README.md
SaitTalhaNisanci: This is just an example, users can put anything here. In this case leaving them empty means they dont want trace or debug messages to be logged. As far as I remember `glog` does not have a debug level, so users can print them at info level if they want to.

internal/lifecycle.go
SaitTalhaNisanci: I have changed it as `ls.logger.Info("HazelcastClient", ClientVersion, "is", newState)`. We dont have buildInfo in Go client like Java. Is that okay?

test/logger/custom_logger_test.go
SaitTalhaNisanci: I just verified that `Info` is called with a boolean flag. I will add more tests.

internal/client.go
```diff
@@ -171,16 +173,40 @@ func (c *HazelcastClient) LifecycleService() core.LifecycleService {
...
+func (c *HazelcastClient) initLogger() {
+	setLogger := c.Config.LoggerConfig().Logger()
+	logLevel := c.getLogLevel()
+	versionMessage := c.name + " [" + c.Config.GroupConfig().Name() + "]" + " [" + ClientVersion + "] "
+	if setLogger == nil {
+		l := logger.New()
+		l.Level = logLevel
+		l.SetVersionMessage(versionMessage)
+		setLogger = l
+	} else if setLoggerImpl, ok := setLogger.(*logger.DefaultLogger); ok {
```
sancar: Does this mean of the logger is a custom one, we don't print the `versionMessage`?

SaitTalhaNisanci: Yes currently if a custom logger is used, like logrus, we dont print the `versionMessage`.

sancar: I think the version message is part of our log message. And it should be always there.

internal/connection.go
SaitTalhaNisanci: There are some parts like that but to be able to get the same message as Java it requires some more work. So I think they can be added in another PR.

PR #394, opened 2018-11-23
Add data structure listener part for Readme

SaitTalhaNisanci:


README.md
```diff
@@ -1455,11 +1458,13 @@ Lifecycle Event >>>  SHUTDOWN
...
 
-#### 7.5.2.1. Listening for Map Events
+#### 7.5.2.1. Map Listener
+
+The Map Listener is used by the Hazelcast `Map`.
 
 You can listen to map-wide or entry-based events. To listen to these events, you need to implement the relevant interfaces.
 
-An entry-based  event is fired after the operations that affect a specific entry. For example, `Map.put()`, `Map.remove()` or `Map.evict()`. An `EntryEvent` object is passed to the listener function.
+An entry-based  event is fired after the operations that affect a specific entry. For example, `Map.Put()`, `Map.Remove()` or `Map.Evict()`. An `EntryEvent` object is passed to the listener function.
```
furkansenharputlu: You may add here an explanation for the following list. 

`You can use the following listeners to listen to entry-based events.`

README.md
```diff
@@ -1487,7 +1492,7 @@ m.Put("1", "Furkan")
 ```
 
 
-A map-wide event is fired as a result of a map-wide operation. For example, `Map.clear()` or `Map.evictAll()`. A `MapEvent` object is passed to the listener function.
+A map-wide event is fired as a result of a map-wide operation. For example, `Map.Clear()` or `Map.EvictAll()`. A `MapEvent` object is passed to the listener function.
```
furkansenharputlu: Same here like above.

README.md
```diff
@@ -1512,6 +1517,113 @@ m.Put("3", "Furkan")
 
 m.Clear()
 ```
+As you see, there is a parameter in the `AddEntryListener` function: `includeValue`. It is a boolean parameter, and if it is `true`, the map event contains the entry value.
+
+#### 7.5.2.2. Entry Listener
+
+The Entry Listener is used by the Hazelcast `MultiMap` and `ReplicatedMap`.
```
furkansenharputlu: Replicated Map*

PR #396, opened 2018-11-27
Refactor cluster service 

SaitTalhaNisanci:
This PR refactors cluster service to increase readability and maintainability. No functionality is changed.

### DIFF
--- a/internal/cluster.go
+++ b/internal/cluster.go
@@ -41,9 +41,8 @@ type clusterService struct {
 	ownerUUID              atomic.Value
 	uuid                   atomic.Value
 	ownerConnectionAddress atomic.Value
-	listeners              atomic.Value
+	listeners              sync.Map
 	addressProviders       []AddressProvider
-	mu                     sync.Mutex
 	memberWg               *sync.WaitGroup
 	reconnectChan          chan struct{}
 	cancelChan             chan struct{}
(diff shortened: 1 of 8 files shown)


internal/proto/core.go
```diff
@@ -88,6 +88,10 @@ func (m *Member) String() string {
 	return memberInfo
 }
 
+func (m *Member) HasEqualAddress(member *Member) bool {
```
ihsandemir: renaming: HasSameAddress

test/client_test.go
```diff
@@ -134,7 +134,7 @@ func TestClientUniqueNames(t *testing.T) {
 
 func TestOpenedClientConnectionCount_WhenMultipleMembers(t *testing.T) {
 	cluster, _ = remoteController.CreateCluster("", DefaultServerConfig)
-	for i := 0; i < 5; i++ {
+	for i := 0; i < 2; i++ {
```
ihsandemir: why decrease?

SaitTalhaNisanci: opening 5 members and closing them that a long time. It was taking 40 seconds. That is why I decreased it.

internal/cluster.go
```diff
@@ -223,31 +257,13 @@ func (cs *clusterService) logMembers() {
...
-	copyListeners := make(map[string]interface{}, len(listeners)-1)
-	for k, v := range listeners {
-		copyListeners[k] = v
-	}
-	_, found := copyListeners[registrationID]
-	if found {
-		delete(copyListeners, registrationID)
-	}
-	cs.listeners.Store(copyListeners)
+	_, found := cs.listeners.Load(registrationID)
```
sancar: this change is not preserving behaviour. Two concurrent call to `RemoveMembershipListener` can return true, eventough only one of them makes the actual delete.  I have checked the sync.Map interface but could not find a fitting solution on it.

SaitTalhaNisanci: I have used a mutex instead of defining a new data structure that extends sync.Map as this probably wont be needed at different places of the codebase. If we need to use this a lot in the future, we can define a new data structure this way.

PR #400, opened 2018-12-03
Refactor connection manager

SaitTalhaNisanci:
This PR refactors connection manager for better readability and maintainability.

It also fixes a defect in `nextConnectionID`, which wouldnt increment the next connection ID atomically.

### DIFF
--- a/internal/connection_manager.go
+++ b/internal/connection_manager.go
@@ -226,8 +226,7 @@ func (cm *connectionManagerImpl) shouldAuthenticateConn(asOwner bool, conn *Conn
 }
 
 func (cm *connectionManagerImpl) NextConnectionID() int64 {
-	cm.nextConnectionID = cm.nextConnectionID + 1
-	return cm.nextConnectionID
+	return atomic.AddInt64(&cm.nextConnectionID, 1)
 }
 
 func (cm *connectionManagerImpl) encodeAuthenticationRequest(asOwner bool) *proto.ClientMessage {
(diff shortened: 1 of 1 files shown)


internal/connection_manager.go
```diff
@@ -191,16 +193,10 @@ func (cm *connectionManagerImpl) getConnection(address core.Address, asOwner boo
...
 		return nil
 	}
-
-	if !asOwner {
-		return conn
-	}
-
-	if conn.isOwnerConnection {
-		return conn
+	if cm.shouldAuthenticateConn(asOwner, conn) {
```
sancar: The method name does not correctly explain the intention here. getConnection does not authenticate.
This is more like a check to see if the connection meets the owner requirement if asked. 
I would prefer not to extract function here. The code is not really complicated, and will probably explain itself better. 
Same for the other usage of this method. There, the connection is already authenticated. It is to see if it is authenticated as the owner if asked that way. Again, I would prefer the leave it as is.

sancar:
> It also fixes a defect in nextConnectionID, which wouldnt increment the next connection ID atomically.
It is not a defect since the increment is always done under connectionsMutex lock. The related change does not do any harm but I would suggest fixing the commit comment. It can give us a false idea in the future.

internal/connection_manager.go
SaitTalhaNisanci: So this check is to see if we are trying to get connection as owner, then the connection is indeed the owner connection right?

In the other usage, if the check is true then we authenticate the connection. If it is already authenticated then why do we authenticate it once more?

The same code is used twice in different methods and the checks are the same checks, so I would prefer to have a method for that.

PR #403, opened 2018-12-10
Update os and runtime metrics before sending them to server

### LINKED ISSUE #402: Client statistics metrics update
Currently registration of runtime and os metrics are done only once and they are not updated. They should be updated every client period seconds.

SaitTalhaNisanci:
Fixes #402.

### DIFF
--- a/internal/statistics.go
+++ b/internal/statistics.go
@@ -48,7 +48,7 @@ type statistics struct {
 	client       *HazelcastClient
 	enabled      bool
 	ownerAddress atomic.Value
-	gauges       map[string]func() string
+	gauges       map[string]string
 	cancel       chan struct{}
 }
 
@@ -58,7 +58,7 @@ func newStatistics(client *HazelcastClient) *statistics {
 		cancel: make(chan struct{}),
 	}
 	stats.enabled = client.properties.GetBoolean(property.StatisticsEnabled)
-	stats.gauges = make(map[string]func() string)
+	stats.gauges = make(map[string]string)
 	stats.ownerAddress.Store(&proto.Address{}) // initialize
 	return &stats
 }
@@ -68,7 +68,6 @@ func (s *statistics) start() {
 	if !s.enabled {
 		return
 	}
-	s.registerMetrics()
 	period := s.client.properties.GetDuration(property.StatisticsPeriodSeconds)
 	if period <= 0 {
 		// This will return the default value since we have a non-positive value now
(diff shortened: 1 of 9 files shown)


internal/statistics.go
```diff
@@ -140,38 +139,30 @@ func (s *statistics) sendStatsToOwner(stats *bytes.Buffer) {
...
 
-func (s *statistics) registerMetrics() {
+func (s *statistics) collectMetrics() {
+	s.collectOSMetrics()
+	s.collectRuntimeMetrics()
+}
+
+func (s *statistics) collectRuntimeMetrics() {
 	var m runtime.MemStats
 	runtime.ReadMemStats(&m)
```
ihsandemir: does this call have an error code return possibility?

internal/statistics.go
```diff
@@ -140,38 +139,30 @@ func (s *statistics) sendStatsToOwner(stats *bytes.Buffer) {
...
+	s.registerGauge("runtime.availableProcessors", strconv.Itoa(runtime.NumCPU()))
+	s.registerGauge("runtime.freeMemory", strconv.Itoa(int(m.HeapIdle)))
+	s.registerGauge("runtime.totalMemory", strconv.Itoa(int(m.HeapSys)))
+	s.registerGauge("runtime.usedMemory", strconv.Itoa(int(m.HeapInuse)))
```
ihsandemir: what is the plan for the missing metrics?

test/statistics/client_statistics_test.go
```diff
@@ -229,6 +231,58 @@ func TestClientStatisticsContent(t *testing.T) {
...
+	defer remoteController.ShutdownCluster(cluster.ID)
+
+	config := hazelcast.NewConfig()
+	config.SetProperty(property.StatisticsEnabled.Name(), "true")
+	config.SetProperty(property.StatisticsPeriodSeconds.Name(), "1")
+
+	client, _ := hazelcast.NewClientWithConfig(config)
+	defer client.Shutdown()
+	var stats string
+	test.AssertEventually(t, func() bool {
```
ihsandemir: is the method name supposed to be `AssertTrueEventually`?

test/statistics/client_statistics_test.go
```diff
@@ -229,6 +231,58 @@ func TestClientStatisticsContent(t *testing.T) {
...
+		}
+		nextRuntimeMetrics := getRuntimeMetrics(nextStats)
+		return isRuntimeMetricsDifferent(runTimeMetrics, nextRuntimeMetrics)
+	})
+
+}
+
+func isRuntimeMetricsDifferent(runtimeMetrics []string, nextRuntimeMetrics []string) bool {
+	isDifferent := false
+	for index, nextRuntimeMetric := range nextRuntimeMetrics {
```
ihsandemir: Should you first check the number of stats in the metrics returned before one-by-one comparison?

internal/statistics.go
```diff
@@ -140,38 +139,30 @@ func (s *statistics) sendStatsToOwner(stats *bytes.Buffer) {
...
-func (s *statistics) registerMetrics() {
+func (s *statistics) collectMetrics() {
+	s.collectOSMetrics()
+	s.collectRuntimeMetrics()
+}
+
+func (s *statistics) collectRuntimeMetrics() {
 	var m runtime.MemStats
 	runtime.ReadMemStats(&m)
+	s.registerGauge("runtime.availableProcessors", strconv.Itoa(runtime.NumCPU()))
```
ihsandemir: I do not think that you need to design with gauges, it was used in Java since the monitor also uses it but for non-java i think it is just extra overhead. you can just pass around the string buffer stream to be filled instead.

internal/statistics.go
SaitTalhaNisanci: No this function does not return an error.

internal/statistics.go
SaitTalhaNisanci: I dont think there is any plan for it that I know of.

We will probably end up including some source code of a third party package in vendor folder.

test/statistics/client_statistics_test.go
```diff
@@ -229,6 +231,58 @@ func TestClientStatisticsContent(t *testing.T) {
...
+		if runtimeMetrics[index] != nextRuntimeMetric {
+			isDifferent = true
```
ihsandemir: you can break the loop here by just returning true at this point.

PR #409, opened 2018-12-26
Implement eventually consistent near cache

### LINKED ISSUE #410: Cluster with only lite members [API-1033]
When the cluster is formed only with lite members, which means they dont store any data, partition table is returned as empty however client still tries to get a nonempty partition table which is not possible.

### LINKED ISSUE #411: Proxy Manager shutdown [API-1259]
When client shutdowns, proxy manager should send shutdown signal to all proxies. Currently there isny anything done when shutdown happens.

SaitTalhaNisanci:
This PR implements eventually consistent near cache. The layout of nearcache is very similar to Java client.

In java eviction is done with sampling map. When eviction is required only one entry is evicted. In csharp when eviction is done %20 of the entries are evicted. In this implementation, the csharp implementation is used so %20 of the entries are evicted when eviction is required as there is no easy way for sampling map.

Fixes #410.
Fixes #411.


SaitTalhaNisanci:
verify

SaitTalhaNisanci:
verify

bpereto:
bump. can we merge this?

PR #425, opened 2019-01-21
Use error in datainput for ease of use

### LINKED ISSUE #426: Serialization Error
When there is an error in read or write methods of serialization, it is expected to return immediately. This results in a lot of code chunks as follows:
```
_, err = in.readInt32()
if err != nil {
   return err
}
_, err = in.readInt64()
if err != nil {
   return err
}
_, err = in.readBool()
if err != nil {
   return err
}
_, err = in.readInt32()
if err != nil {
   return err
}
```

The goal here is to return the first error. If we had an error field in internal, for example in `DataInput`, and when any read or write method is called we could check the error field, if it is set just return that. This way it would be possible to write:

```
in.readInt32()
in.readInt64()
in.readBool()
in.readInt32()
return in.Error() // the first error 
```

and the returned error will be the first one, so the functionality is the same.

The structure of implementation:

```
	if i.err != nil {  // if this is set it means a previous method returned an error
		return nil
	}
	var ret (type)
	ret, i.err = i.read(type)() // readData, readInt32 etc...
	return ret
```

See: https://blog.golang.org/errors-are-values

SaitTalhaNisanci:
Fixes #426.


SaitTalhaNisanci:
verify

asimarslan:
verify

serialization/api.go
```diff
@@ -214,72 +214,101 @@ type PositionalDataOutput interface {
...
-	// ReadUTF returns string read and error.
-	ReadUTF() (string, error)
+	// ReadUTF returns string read.
+	// It returns the zero value if an error is set previously.
+	ReadUTF() string
 
-	// ReadObject returns object read and error.
-	ReadObject() (interface{}, error)
+	// ReadObject returns object read.
+	// It returns the zero value if an error is set previously.
```
sancar: returns `nil` if an error is set previously. And there are similar problems in other methods. Can recheck them?

SaitTalhaNisanci: I used `the zero value` as in a general phrase. For example for `int` it is 0, for `string` it is "", for object it is `nil`, do you think it is unclear?

sancar: Yes, I think we need to be precise.

PR #451, opened 2019-02-25
Fix partition service when cluster has only lite members

### LINKED ISSUE #410: Cluster with only lite members [API-1033]

SaitTalhaNisanci:
Partition service gets stuck when cluster is formed only by lite members. This PR fixes that.

Fixes #410.

### DIFF
--- a/internal/partition.go
+++ b/internal/partition.go
@@ -91,6 +91,16 @@ func (ps *partitionService) GetPartitionIDWithKey(key interface{}) (int32, error
 	return ps.GetPartitionID(data), nil
 }
 
+func (ps *partitionService) isClusterFormedByOnlyLiteMembers() bool {
+	members := ps.client.ClusterService.GetMembers()
+	for _, member := range members {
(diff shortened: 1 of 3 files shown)


internal/partition.go
```diff
@@ -126,6 +136,10 @@ func (ps *partitionService) shutdown() {
 
 func (ps *partitionService) waitForPartitionsFetchedOnce() {
 	for len(ps.mp.Load().(map[int32]*proto.Address)) == 0 && ps.client.ConnectionManager.IsAlive() {
+		if ps.isClusterFormedByOnlyLiteMembers() {
+			ps.logger.Warn("Partitions cannot be assigned since all nodes in the cluster are lite members")
```
sancar: We need to return error from here. And it should be propagated to the user.

Issue #458, opened 2019-03-11
gometalinter: has been depricated


7erry:
Go version: go version go1.11.4 darwin/amd64
Hazelcast Go Client version: building from snapshot/latest
Hazelcast server version:3.12-SNAPSHOT
Number of the clients:1
Cluster size, i.e. the number of Hazelcast cluster members:2
OS version : Windows/Linux/OSX 10.14.3

#### Expected behaviour
./build.sh to build the go client

#### Actual behaviour
7erry:hazelcast-go-client twal7ers$ ./build.sh 
./build.sh: line 13: pushd: /src/github.com/hazelcast/hazelcast-go-client: No such file or directory
downloading gometalinter
./linter.sh: line 6: gometalinter: command not found
error when installing gometalinter

#### Steps to reproduce the behaviour
git clone https://github.com/hazelcast/hazelcast-go-client.git
cd hazelcast-go-client
./build.sh

PR #459, opened 2019-03-13
Migrate from gometalinter to golangci

sancar:
fixes https://github.com/hazelcast/hazelcast-go-client/issues/458

See https://github.com/alecthomas/gometalinter/issues/590.


sancar:
run-lab-run

sancar:
run-lab-run

SaitTalhaNisanci:
verify

linter.sh
```diff
@@ -1,28 +1,17 @@
...
-        echo "error when installing gometalinter"
-    exit 1
-    fi
-fi
-
-gometalinter ./... 2>&1 > gometalinter-err.out
-
+if ! [ -d $GOPATH/src/github.com/golangci/golangci-lint/  ]; then
+    echo "downloading golangci-lint"
+    go get github.com/golangci/golangci-lint/cmd/golangci-lint
```
SaitTalhaNisanci: I think we should use:
```
curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.16.0
```

Since it says this is the recommended way here to have a stable version: https://github.com/golangci/golangci-lint#ci-installation.

And in that case the above if check would be: `if ! command -v golangci-lint > /dev/null; then`

.golangci.yml
```diff
@@ -0,0 +1,62 @@
...
+    - deadcode
+    - goimports
+    - golint
+    - ineffassign
+    - interfacer
+    - lll
+    - misspell
+    - unconvert
+    - unparam
+    - vet
```
SaitTalhaNisanci: It seems like this cant be run. When I look at the logs I see this:

`[33mWARN[0m [runner] Can't run linter govet: asmdecl: analysis skipped due to errors in package `

I got this locally too.

sancar: Excluded govet for now from linters

.golangci.yml
```diff
@@ -0,0 +1,61 @@
+# This file contains all available configuration options
```
mdumandag: Hostname for this test https://github.com/hazelcast/hazelcast-go-client/blob/master/test/ssl/client_ssl_authentication_test.go#L256 should be `foobar.hazelcast.com` due to new comodo certificate

PR #460, opened 2019-03-14
Json API update

sancar:
class name is aligned with other clients.
Constructors to create json from string and object is added.
ToString method is added.

Added nil check for hazelcastJSONValue 
Create methods will return nil actual check will be done in
put,get, ... methods.

This is to simplify API. Otherwise, user needs to seperate API
calls to two line as follows:
``
value , _ := CreateHazelcastJSONValue(person)
map.put("key", value)
``
isNil method also implemented on proxy to correctly check nil type over
interface{}

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -19,6 +19,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"time"
+
+	"github.com/hazelcast/hazelcast-go-client/internal/proto/bufutil"
+	"github.com/hazelcast/hazelcast-go-client/internal/util/nilutil"
 )
 
 // Address represents an address of a member in the cluster.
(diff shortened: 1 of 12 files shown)


core/api.go
```diff
@@ -304,10 +304,29 @@ type LoadBalancer interface {
 	Next() Member
 }
 
-// HazelcastJSON is a wrapper for json byte array.
+// HazelcastJSONValue is a wrapper for json byte array.
 // JsonString should be a valid json string.
-type HazelcastJSON struct {
-	JSONString []byte
+type HazelcastJSONValue struct {
```
asimarslan: Can you improve the API doc including the below func's, just copy the java version content?

core/api.go
```diff
@@ -304,10 +304,29 @@ type LoadBalancer interface {
...
+	jsonString []byte
+}
+
+func CreateHazelcastJSONValueFromString(jsonString string) *HazelcastJSONValue {
```
SaitTalhaNisanci: this belongs to the API, godoc should be added. Normally linter should fail saying something like public `CreateHazelcastJSONValueFromString ` should have a comment..., there is a bug with the linter or its configuration I think.

core/api.go
```diff
@@ -304,10 +304,29 @@ type LoadBalancer interface {
...
+	return &HazelcastJSONValue{[]byte(jsonString)}
+}
+func CreateHazelcastJSONValue(object interface{}) *HazelcastJSONValue {
+	if object == nil {
+		return nil
+	}
+	byteArray, _ := json.Marshal(object)
```
SaitTalhaNisanci: what if this method returns an error? I think this function should return an error in that case.

core/api.go
```diff
@@ -304,10 +304,29 @@ type LoadBalancer interface {
...
+	return &HazelcastJSONValue{byteArray}
+}
+
+func (h *HazelcastJSONValue) Unmarshal(v interface{}) {
+	json.Unmarshal(h.jsonString, &v)
```
SaitTalhaNisanci: This method also returns an error, why dont we return it to the user?

core/api.go
sancar: This slipped my attention. It seems only time marshal returns error is when an object is `function` `complex` or  `channel`

core/api.go
sancar: An assumption here was that server will return valid JSON always. When I reconsider, there is nothing preventing the user from putting invalid JSON, therefore it makes more sense to return an error from Unmarshal. @asimarslan  wdyt ?

asimarslan: yes it should just return `json.Unmarshal` all of the results. no need to hide it.

core/api.go
```diff
@@ -304,10 +307,47 @@ type LoadBalancer interface {
...
-type HazelcastJSON struct {
-	JSONString []byte
+// HazelcastJSONValue is a wrapper for Json formatted strings. It is
+// preferred to store HazelcastJsonValue instead of string/[]byte for
+// Json. Users can run predicates/aggregations and use indexes on
+// the attributes of the underlying Json objects.
+//
+// HazelcastJSONValue is queried using Hazelcast's querying language.
+//
+// HazelcastJSONValue keeps given string as it is. Strings are not
```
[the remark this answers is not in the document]
sancar: This is copied from java so I am leaving it as is.

core/api.go
```diff
@@ -304,10 +307,47 @@ type LoadBalancer interface {
 	Next() Member
 }
 
-// HazelcastJSON is a wrapper for json byte array.
-// JsonString should be a valid json string.
-type HazelcastJSON struct {
-	JSONString []byte
+// HazelcastJSONValue is a wrapper for Json formatted strings. It is
```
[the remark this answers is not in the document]
sancar: I will convert all to JSON since it is that way in the class name.

PR #470, opened 2019-11-22
Entry loaded listener

neilstevenson:
Completes support for the new "LOADED" event type


neilstevenson:
There doesn't seem to be any test coverage for specific event types

PR #492, opened 2020-08-06
Update Hazelcast Cloud documentation and code sample

### LINKED ISSUE #490: Confusing chapter about Cloud Discovery
Original report: https://github.com/hazelcast/hazelcast-python-client/issues/207

I looked at the README, the problem described in the issue above also exists in this client. 

Also, we need to update the code samples since they use default system certificates. However, Hazelcast Cloud uses private certificate pair now. See https://github.com/hazelcast/hazelcast-go-client/blob/master/sample/hzclouddiscovery/cloud_discovery_example.go#L29

gulcesirvanci:
Does not use secure communication by default. One have to enable encryption before starting the cluster to use secure communication. 
Does not use certificates signed by trusted CA. It generates a private certificate pair.
This commit update hazelcast cloud documentation and the code sample accordingly.

fixes #490


gulcesirvanci:
verify

srknzl:
Related to #538, could you check that as well. Thanks

PR #497, opened 2020-11-17
Pass isSmart option as localOnly flag in AddEntryListenerWithPredicate

### LINKED ISSUE #495: EntryAdded listener method invoked three times
Go version: 1.15.3
Hazelcast Go Client version: 0.6.0
Hazelcast server version: 3.12
Number of the clients: 1
Cluster size, i.e. the number of Hazelcast cluster members: 3
OS version : Linux  

I set up an entry listener with predicate using _AddEntryListenerWithPredicate_ method. After that I put one new entry into the hazelcast map. 

Note: When I use _AddEntryListener_ (without a predicate) everything works fine.

#### Expected behaviour
EntryAdded listener method invoked only once. With the new map entry I've put. 
 
#### Actual behaviour
EntryAdded listener method invoked three times (probably equals cluster size). 

1. Set entry listener with predicate using _AddEntryListenerWithPredicate_ method.
2. Put an entry to the map

Maybe I do something wrong. I am ready to give more infomation to invistigate the reason of problem.

JorgenPo:
It seems like the client ignores the _isSmart_ option when executing _AddEntryListenerWithPredicate_ request. 
This change fixes #495.

### DIFF
--- a/internal/map.go
+++ b/internal/map.go
@@ -516,7 +516,7 @@ func (mp *mapProxy) AddEntryListenerWithPredicate(listener interface{}, predicat
 	if err != nil {
 		return "", err
 	}
-	request = proto.MapAddEntryListenerWithPredicateEncodeRequest(mp.name, predicateData, includeValue, listenerFlags, false)
+	request = proto.MapAddEntryListenerWithPredicateEncodeRequest(mp.name, predicateData, includeValue, listenerFlags, mp.isSmart())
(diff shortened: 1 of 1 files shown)


yuce:
Thanks @JorgenPo for your contribution. This should be fixed in the latest v1 preview. Note that that version requires Hazelcast 4.x

PR #539, opened 2021-02-07
README correction: scripts are in bash and cannot be run using sh

srknzl:
Building and testing part:
The scripts are in bash and cannot be run using sh(hashbang indicates that). The README should be changed to reflect that.


yuce:
Thanks @srknzl, this should be fixed in the latest master.

PR #541, opened 2021-02-16
[DON'T MERGE] Enabled go module support for v3

yuce:
- Enabled go mod support for v3
- Duplicated v3 code in its own directory to prepare for v4


yuce:
The tests pass on local but fails on jenkins, since I need to update CI config. Let's hold merging the PR until CI changes are done.

PR #552, opened 2021-03-17
Initial V4 Implementation [API-261, API-263, API-137, API-345, API-399, API-400, API-405]

yuce:


victiondev:
Any idea when this is being merged? Need it for a project of mine which depends on a microservice using this.

yuce:
Hi @victiondev, we don't have a public release date yet, but you can watch this repo or join us at https://hazelcastcommunity.slack.com/channels/go-client for the announcement.

client.go
```diff
@@ -0,0 +1,410 @@
...
+		return nil, ErrClientNotReady
+	}
+	if ctx == nil {
+		return nil, ErrContextIsNil
+	}
+	return c.proxyManager.getReplicatedMapWithContext(ctx, name)
+}
+
+// Start connects the client to the cluster.
+func (c *Client) Start() error {
```
sancar, line 196: I think we should start with API compatible with our other clients. 
When we introduce `NewClient` and `Start` there are more questions to answer like:
- what happens if user starts a shutdown client ? 
- what is the relation with this API and the blue/green feature that we are going to have ?

I think before we discuss and decide what the behaviour should be we should not introduce an API.

types/index.go
```diff
@@ -0,0 +1,43 @@
...
+
+const (
+	UniqueKeyTransformationObject = int32(0)
+	UniqueKeyTransformationLong   = int32(1)
+	UniqueKeyTransformationRaw    = int32(2)
+)
+
+type IndexConfig struct {
+	Name               string
+	Type               int32
```
sancar, line 35: Should this int32 be `IndexType` instead ?

types/index.go
```diff
@@ -0,0 +1,43 @@
...
+	Attributes         []string
+	BitmapIndexOptions BitmapIndexOptions
+}
+
+type BitmapIndexOptions struct {
+	UniqueKey               string
+	UniqueKeyTransformation int32
```
sancar, line 42: Should this int32 be `UniqueKeyTransformationType` instead to avoid user to pass random ints here?

proxy_map.go
```diff
@@ -0,0 +1,1023 @@
...
+	if subscriptionIDInt, err := strconv.Atoi(subscriptionID); err != nil {
+		return fmt.Errorf("invalid subscription ID: %s", subscriptionID)
+	} else {
+		m.userEventDispatcher.Unsubscribe(EventEntryNotified, subscriptionIDInt)
+		return m.listenerBinder.Remove(m.name, subscriptionIDInt)
+	}
+}
+
+func (m *Map) addIndex(indexConfig types.IndexConfig) error {
+	request := codec.EncodeMapAddIndexRequest(m.name, indexConfig)
```
sancar, line 782: IndexConfig validation is missing. See https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/proxy/ClientMapProxy.java#L1326

internal/cluster/cluster_service.go
```diff
@@ -0,0 +1,329 @@
...
+}
+
+func (s *ServiceImpl) sendMemberListViewRequest(conn *Connection) {
+	request := codec.EncodeClientAddClusterViewListenerRequest()
+	inv := s.invocationFactory.NewConnectionBoundInvocation(request, -1, nil, conn, s.config.InvocationTimeout, func(response *proto.ClientMessage) {
+		codec.HandleClientAddClusterViewListener(response, func(version int32, memberInfos []pubcluster.MemberInfo) {
+			s.logger.Debug(func() string { return "members updated" })
+			s.eventDispatcher.Publish(NewMembersUpdated(memberInfos, version))
+		}, func(version int32, partitions []proto.Pair) {
+			s.eventDispatcher.Publish(NewPartitionsUpdated(partitions, version))
```
sancar, line 167: Why are we not making a call to directly to "partition_service.handlePartitionsUpdated" ?
Is there a reason for this to go over an event dispatcher ?

internal/cluster/partition_service.go
```diff
@@ -0,0 +1,168 @@
...
+type partitionTable struct {
+	partitionStateVersion int32
+	partitions            map[int32]internal.UUID
+	mu                    *sync.RWMutex
+}
+
+func (p *partitionTable) Update(pairs []proto.Pair, version int32) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	if version > p.partitionStateVersion {
```
sancar, line 131: We should also check the last connection we got the event. 
It could be the case that the new connection has smaller partition state version. 
And new events will not be applied eventough it should.
see https://github.com/hazelcast/hazelcast/blob/b1629303830a0394b86fb919d2c88d5178b05846/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/ClientPartitionServiceImpl.java#L93

internal/cluster/partition_service.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func (s *PartitionService) GetPartitionID(keyData pubserialization.Data) int32 {
+	if count := s.PartitionCount(); count == 0 {
+		// Partition count can not be zero for the sync mode.
+		// On the sync mode, we are waiting for the first connection to be established.
+		// We are initializing the partition count with the value coming from the server with authentication.
+		// This exception is used only for async mode client.
+		// TODO: panic
+		//core.NewHazelcastClientOfflineException()
```
sancar, line 95: Why are we not returning error here ?

internal/cluster/partition_service.go
```diff
@@ -0,0 +1,168 @@
...
+func (p *partitionTable) GetOwnerUUID(partitionID int32) internal.UUID {
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+	if uuid, ok := p.partitions[partitionID]; ok {
+		return uuid
+	}
+	return nil
+}
+
+func (p *partitionTable) PartitionCount() int {
```
sancar, line 156: This method looks unused. I think we can delete it. Partition count is already a different variable.

internal/cluster/partition_service.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/hazelcast/hazelcast-go-client/internal/util/murmur"
+	pubserialization "github.com/hazelcast/hazelcast-go-client/serialization"
+)
+
+type PartitionServiceCreationBundle struct {
+	SerializationService *iserialization.Service
+	EventDispatcher      *event.DispatchService
+	startCh              chan struct{}
+	startChAtom          int32
+	Logger               ilogger.Logger
```
sancar, line 37: These two fields startCh and StartChAtom looks unused. 
Is that intentional ?

internal/cluster/partition_service.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (s *PartitionService) handlePartitionsUpdated(event event.Event) {
+	if partitionsUpdatedEvent, ok := event.(*PartitionsUpdated); ok {
+		s.partitionTable.Update(partitionsUpdatedEvent.Partitions, partitionsUpdatedEvent.Version)
+		s.eventDispatcher.Publish(NewPartitionsLoaded())
+		s.logger.Debug(func() string { return "partitions loaded" })
+	}
+}
+
+func (s *PartitionService) checkAndSetPartitionCount(newPartitionCount int32) {
```
sancar, line 118: Even blue/green feature is not implemented, we should check if partition count same. 
If users closes and reopens a cluster ,our clients can connect back to cluster. We don't want our clients to connect to a cluster with different partition count.

internal/cluster/connection_invocation_handler.go
```diff
@@ -0,0 +1,143 @@
...
+		connectionManager: bundle.ConnectionManager,
+		clusterService:    bundle.ClusterService,
+		logger:            bundle.Logger,
+		cb:                circuitBreaker,
+	}
+}
+
+func (h *ConnectionInvocationHandler) Invoke(inv invocation.Invocation) error {
+	_, err := h.cb.Try(func(ctx context.Context) (interface{}, error) {
+		if h.clusterService.SmartRoutingEnabled() {
```
sancar, line 76: It looks like just to be able to check this we are adding a dispatcher to partition service and setting a boolean and removing the dispatcher back. 
Do we really need all these ceremony ?
If the operation is partition based invokeSmart will fail after seeing that partition owner is nil. And this is not something expensive. 

Also note that for address based operations this `SmartRoutingEnabled` gives to wrong result. 
Address based operations can go to an address directly even-though partition table is not arrived yet.

client.go
```diff
@@ -0,0 +1,410 @@
...
+	if !c.canStart() {
```
puzpuzpuz, line 197: This check against running `client.Start()` in multiple goroutines in parallel. Such code would lead to a race with undefined behavior, so we should protect this method with a mutex. The same consideration applies to `Shutdown()` method.

proxy_map.go
```diff
@@ -0,0 +1,1023 @@
...
+	Key                interface{}
+	NotifyEntryAdded   bool
+	NotifyEntryRemoved bool
+	NotifyEntryUpdated bool
+}
+
+type Map struct {
+	*proxy
+	refIDGenerator *iproxy.ReferenceIDGenerator
+	ctx            context.Context
```
puzpuzpuz, line 49: Context shouldn't be stored at the proxy level since the context should be provided at the operation level. Do we expect users to create separate proxies for each operation/chain of operations?

The same question applies to other proxies.

proxy.go
```diff
@@ -0,0 +1,288 @@
...
+	request := codec.EncodeClientCreateProxyRequest(p.name, p.serviceName)
+	inv := p.invocationFactory.NewInvocationOnRandomTarget(request, nil)
+	p.requestCh <- inv
+	if _, err := inv.Get(); err != nil {
+		return fmt.Errorf("error creating proxy: %w", err)
+	}
+	return nil
+}
+
+func (p *proxy) destroy() error {
```
puzpuzpuz, line 138: Why is this method kept private? How users are supposed to destroy cluster-side data structures in v4.0?

proxy_manager.go
```diff
@@ -0,0 +1,103 @@
...
+	obj, ok := m.proxies[name]
+	m.mu.RUnlock()
+	if ok {
+		return obj, nil
+	}
+	if p, err := m.createProxy(serviceName, objectName); err != nil {
+		return nil, err
+	} else {
+		m.mu.Lock()
+		m.proxies[name] = p
```
puzpuzpuz, line 91: It's better to check `m.proxies[name]`'s contents before this line in order to prevent races between `Get<ds_name>` calls leading to different proxy structs returned from those calls.

puzpuzpuz (commented):
Leaving some intermediate comments.

internal/invocation/invocation.go
```diff
@@ -0,0 +1,137 @@
...
+func (i *Impl) EventHandler() proto.ClientMessageHandler {
+	return i.eventHandler
+}
+
+func (i *Impl) Get() (*proto.ClientMessage, error) {
+	response := <-i.response
+	return i.unwrapResponse(response)
+}
+
+func (i *Impl) GetContext(ctx context.Context) (*proto.ClientMessage, error) {
```
sancar, line 87: This name sounds a lot like a simple getter.
Can we rename it to GetWithContext ?

proxy_map.go
```diff
@@ -0,0 +1,1023 @@
...
+			partitionToKeys[partitionKey] = append(arr, keyData)
+		}
+	}
+	result := make([]types.Entry, 0, len(keys))
+	// create futures
+	f := func(partitionID int32, keys []pubserialization.Data) cb.Future {
+		return m.circuitBreaker.TryContext(m.ctx, func(ctx context.Context) (interface{}, error) {
+			request := codec.EncodeMapGetAllRequest(m.name, keys)
+			inv := m.invokeOnPartitionAsync(request, partitionID)
+			return inv.GetWithTimeout(1 * time.Second)
```
sancar, line 241: Is this one second a leftover ? 
Same goes for putAll as well?

internal/invocation/invocation.go
```diff
@@ -0,0 +1,137 @@
...
+type Impl struct {
+	request         *proto.ClientMessage
+	response        chan *proto.ClientMessage
+	completed       int32
+	boundConnection *Impl
+	address         *pubcluster.AddressImpl
+	partitionID     int32
+	//sentConnection  atomic.Value
+	eventHandler func(clientMessage *proto.ClientMessage)
+	deadline     time.Time
```
sancar, line 55: I could not find a usage of `deadline`. Is it missing?

internal/cb/circuitbreaker.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (cb *CircuitBreaker) try(ctx context.Context, resultCh chan interface{}, tryHandler TryHandler) {
+	var result interface{}
+	var err error
+loop:
+	for trial := 0; trial <= cb.MaxRetries; trial++ {
+		select {
+		case <-ctx.Done():
+			err = ctx.Err()
+			break loop
```
sancar, line 89: There should be a place where a client message needs to be converted to error. I traced this errors only to local ones. Are we checking if remove response exception or not anywhere.

Depending on error type of local/remote errors, configuration and type of the request, we should retry it or not. 
I am assuming it is the circuit breakers job, but I could not find the related logic.

To make it more clear, I am talking about this logic.
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/ClientInvocation.java#L314

yuce: Error responses are decoded at https://github.com/yuce/hazelcast-go-client/blob/01a76f647faf7b7c7bfa8ca2b793f845b98fc164/internal/cluster/connection.go#L186 and handled at: https://github.com/yuce/hazelcast-go-client/blob/01a76f647faf7b7c7bfa8ca2b793f845b98fc164/internal/invocation/invocation_service.go#L120

circuit breaker retries all errors, I'll fix that.

proxy_map.go
yuce: `invokeOnPartitionAsync` just creates the invocation, it doesn't wait it to be resolved. But we don't need that anymore, so I've replaced it with invokeOnPartition.

internal/cluster/cluster_service.go
```diff
@@ -0,0 +1,329 @@
...
+	addrToMemberUUID map[string]string
+	membersMu        *sync.RWMutex
+	version          int32
+}
+
+func newMembersMap() membersMap {
+	return membersMap{
+		members:          map[string]*Member{},
+		addrToMemberUUID: map[string]string{},
+		membersMu:        &sync.RWMutex{},
```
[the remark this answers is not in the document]
yuce: Mutexes should not be copied. Making it a value type here may cause it to be copied. https://eli.thegreenplace.net/2018/beware-of-copying-mutexes-in-go/

client.go
yuce: I've removed NewClient/Start

proxy_map.go
yuce: Thanks, added that.

internal/cluster/partition_service.go
yuce: Thanks, updated.

internal/cluster/partition_service.go
yuce: Updated.

internal/cluster/partition_service.go
yuce: This method is used in the GetPartitionID function. We cannot unexport PartitionCount, since there is also `partitionCount`. This code is package private, so we can keep it as is for now.

internal/cluster/partition_service.go
yuce: Removed.

internal/cluster/partition_service.go
yuce: Updated.

internal/cluster/cluster_service.go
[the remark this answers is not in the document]
puzpuzpuz: It's simple to avoid copying mutexes. Values for mutexes is a common approach and it can be seen in many popular open source projects written in Go and personally I don't see any problems with it.

client.go
yuce: Thanks, I've unexported start and added a race test to shutdown

proxy_map.go
yuce: `GetMapContext` doesn't create a proxy if it already exists. Current context implementation is experimental and used mainly for lock support. This is not the final design.

proxy.go
yuce: updated

client.go
```diff
@@ -0,0 +1,394 @@
...
+func (c *Client) makeCredentials(config *Config) *security.UsernamePasswordCredentials {
+	securityConfig := config.ClusterConfig.SecurityConfig
+	return security.NewUsernamePasswordCredentials(securityConfig.Username, securityConfig.Password)
+}
+
+func (c *Client) createComponents(config *Config) {
+	credentials := c.makeCredentials(config)
+	serializationService, err := serialization.NewService(&config.SerializationConfig)
+	if err != nil {
+		panic(fmt.Errorf("error creating client: %w", err))
```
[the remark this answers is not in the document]
yuce: Thanks, moved serialization service creation to newClient.

internal/cb/circuitbreaker.go
```diff
@@ -0,0 +1,138 @@
...
+	future := NewFutureImpl()
+	go cb.try(ctx, future.resultCh, tryHandler)
+	return future
+}
+
+func (cb *CircuitBreaker) try(ctx context.Context, resultCh chan interface{}, tryHandler TryHandler) {
+	var result interface{}
+	var err error
+loop:
+	for trial := 0; trial <= cb.MaxRetries; trial++ {
```
puzpuzpuz, line 85: Arbitrary invocation can't be retried safely. `ClientMessage.IsRetryable()` and redo operation config option should be considered.

See: https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/ClientInvocation.java#L402

proxy_map.go
```diff
@@ -0,0 +1,1023 @@
...
+		request := codec.EncodeMapUnlockRequest(m.name, keyData, m.lockID, refID)
+		_, err = m.invokeOnKey(m.ctx, request, keyData)
+		return err
+	}
+}
+
+// ListenEntryNotification adds a continuous entry listener to this map.
+func (m *Map) ListenEntryNotification(config MapEntryListenerConfig, handler EntryNotifiedHandler) (string, error) {
+	flags := makeListenerFlags(&config)
+	subscriptionID := int(m.refIDGenerator.NextID())
```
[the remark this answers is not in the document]
yuce: I hesitate to use the UUID type in the public API, since that would force the user to depend on our internal UUID implementation without much benefit. I've changed event subscription ID type to int64

internal/cluster/connection.go
```diff
@@ -0,0 +1,256 @@
...
+		c.closedTime.Store(time.Time{})
+		c.startTime = timeutil.GetCurrentTimeInMilliSeconds()
+		c.lastRead.Store(time.Now())
+		if err := c.sendProtocolStarter(); err != nil {
+			c.socket.Close()
+			c.socket = nil
+			return err
+		}
+		go c.socketReadLoop()
+		go c.socketWriteLoop()
```
puzpuzpuz, line 89: Having a single event loop per reads and writes hurts performance since only one write on a single connection can make progress simultaneously. A better approach might be to have a number (say, `GOMAXPROCS`) goroutines responsible for reads and writes for active connections.

Another thing that could improve performance significantly is accumulation of queued client messages into a single buffer before writing them into the socket.

internal/cluster/connection.go
```diff
@@ -0,0 +1,256 @@
...
+			if c.isTimeoutError(err) {
+				continue
+			}
+			c.logger.Errorf("read error: %w", err)
+			break
+		}
+		if n == 0 {
+			continue
+		}
+		clientMessageReader.Append(buf[:n])
```
puzpuzpuz, line 176: nit: memory copying could be avoided here in situations when the input buffer (byte slice) contains all parts of a response message. In that case, we could reslice the buffer into frames while reading it.

puzpuzpuz (commented):
Leaving some intermediate comments.

proxy_map.go
yuce: I  put back invokeOnPartitionAsync.

internal/cb/circuitbreaker.go
```diff
@@ -0,0 +1,138 @@
...
+				break loop
+			}
+			if trial < cb.MaxRetries {
+				time.Sleep(cb.RetryPolicyFunc(trial))
+			}
+		}
+	}
+	if err != nil {
+		// failed
+		cb.notifyFailed()
```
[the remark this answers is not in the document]
yuce: I got the inspiration from Vert.x's circuit breaker for this implementation: https://vertx.io/docs/vertx-circuit-breaker/java/ The main purpose of it is avoiding stressing a part of the system with retries when most probably those retries will fail.

internal/cluster/invocation_factory.go
```diff
@@ -0,0 +1,74 @@
...
+
+func NewConnectionInvocationFactory(invocationTimeout time.Duration) *ConnectionInvocationFactory {
+	return &ConnectionInvocationFactory{
+		invocationTimeout: invocationTimeout,
+	}
+}
+
+func (f *ConnectionInvocationFactory) NewInvocationOnPartitionOwner(message *proto.ClientMessage, partitionID int32) invocation.Invocation {
+	// TODO: remove this when message pool is implemented
+	message = message.Copy()
```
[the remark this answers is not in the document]
yuce: I've implemented a byte array pool in the client. Although that decreased the allocations, the performance was affected negatively so I held back committing those changes. I will try the message pool (as you said not trivial) and frame pool once the client stabilized a bit more.

internal/cb/circuitbreaker.go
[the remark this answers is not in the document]
sancar: Our clients should behave the same. I did not review this part in detail because I am expecting this to be configured with `ConnectionRetryConfig` and also as stated in this comment https://github.com/hazelcast/hazelcast-go-client/pull/552#discussion_r620008875 , we did not port the invocation retry logic yet. Once that is ported, I will also review again to be sure it behaves same as java client. 

In short, I would expect this kind of central logics to work same across all the clients. 

If there is missing something even on others(including java), we should identify the problem together and improve that on all(at least open an issue to other languages to be addressed later).

internal/cluster/connection.go
yuce: I've tried having more than a single write loop, but that doesn't seem to help dramatically. Having multiple read loops doesn't seem to be feasible, unless we share the message reader between loops.

I didn't try batching, that will probably affect the performance quite dramatically.

puzpuzpuz: Sorry, I misread the code. Please ignore the above message.

nit: event loops are run on per connection basis and the problem is a bit different. When it comes to writes, the concurrency between connections is limited by the `processIncoming()` loop run by the invocation service. Since it reads from the `requestCh` channel in the loop and then writes the client message into a channel of size 1 on one of connections, a "slow" connection or a write of a large payload on one of connections may lead to a situation when `requestCh` is full and no operations are able to make progress. A straight-forward solution to this problem may be to allow caller goroutines to write into connections directly, rather than through a set of event loops connected with channels.

internal/cluster/connection.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (c *Connection) isAlive() bool {
+	return atomic.LoadInt32(&c.status) == open
+}
+
+func (c *Connection) socketWriteLoop() {
+	for {
+		select {
+		case request, ok := <-c.pending:
```
puzpuzpuz, line 136: nit: it should be more or less trivial to implement autopipelining (Node.js client already [has it](https://github.com/hazelcast/hazelcast-nodejs-client/pull/481)) with current design, so I'm leaving some hints here.

`pending` channel's size should be increase from the current size of 1 message to something more significant (say, 64). Then when the `socketWriteLoop()` loop reads a message from the channel, it should append it to an intermediate array and then schedule a short-living timer (say, 1ms), if it's not yet scheduled. Once the timer is triggered, all accumulated messages should be written into an intermediate, reusable buffer and only then written into the socket.

Autopipelining should improve client's throughput (in Node.js client it was something 20-30% improvement in a number of benchmarks) at the cost of slightly increased latency of individual messages. For high-priority messages, like heartbeats, we could introduce a field on `ClientMessage` struct that would instruct the `socketWriteLoop()` loop to immediately write any accumulated messages.

proxy_map.go
puzpuzpuz: The thing is that `GetMapContext` method is root level, so it looks a lot like `GetMap` and thus many users may use it as a way to obtain a proxy instance and cache it in their modules. The purpose of the method would be more obvious if it would be available on the `Map` struct itself (see `withContext` method).

proxy_map.go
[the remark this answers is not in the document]
puzpuzpuz: A string may be returned back to the user instead of the UUID type. Also, all Hz v4.0 clients need to support UUID serialization, so a dependency or a in-house implementation will have to be added into the Go client anyway.

internal/cluster/connection_invocation_handler.go
```diff
@@ -0,0 +1,143 @@
...
+			h.logger.Trace(func() string {
+				return fmt.Sprintf("sending invocation to %s failed, address not found", addr.String())
+			})
+			return fmt.Errorf("address not found: %s", addr.String())
+		}
+	}
+	return h.sendToConnection(inv, conn)
+}
+
+func (h *ConnectionInvocationHandler) sendToOwnerAddress(inv invocation.Invocation) error {
```
puzpuzpuz, line 136: There is no definition of owner connection in v4.0. Also, smart invocations should go to a random connection based on the load balancer implementation.

See:
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/ClientInvocation.java#L198
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/connection/tcp/TcpClientConnectionManager.java#L800

proxy_map.go
```diff
@@ -0,0 +1,1067 @@
...
+	return &Map{
+		proxy:          p,
+		refIDGenerator: iproxy.NewReferenceIDGenerator(),
+		ctx:            ctx,
+		lockID:         lockID,
+	}
+}
+
+func (m *Map) withContext(ctx context.Context) *Map {
+	lockID := int64(reflect.ValueOf(ctx).Pointer())
```
puzpuzpuz, line 69: nit: this method seems to be unused. Also, it seems weird and potentially unsafe to use context pointer's address as the lock identity. While Go has a non-moving GC, new objects may be allocated at the same address as the old, collected objects.

internal/event/util.go
```diff
@@ -0,0 +1,39 @@
...
+	"strconv"
+)
+
+// MakeSubscriptionID creates subscription ID from a function
+func MakeSubscriptionID(fun interface{}) int64 {
+	value := reflect.ValueOf(fun)
+	if value.Kind() != reflect.Func {
+		panic("not a func")
+	}
+	return int64(value.Pointer())
```
puzpuzpuz, line 30: Godoc for [`Value.Pointer()`](https://golang.org/pkg/reflect/#Value.Pointer) says the following:
```
If v's Kind is Func, the returned pointer is an underlying code pointer, but not necessarily enough to identify a single function uniquely. The only guarantee is that the result is zero if and only if v is a nil func Value.
```

Considering this, relying on function pointer's address for subscription ids may be a bad idea. Not speaking of potential collisions with ids assigned to user subscriptions. WDYT?

puzpuzpuz (commented):
Leaving intermediate comments.

proxy_map.go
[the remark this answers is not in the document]
puzpuzpuz: Another problem with the current approach is collision of subscription ids between different `Map`s. Since each `Map` gets a new `ReferenceIDGenerator`, assigned subscription ids in all maps start with 1 and overwrite each other when `m.userEventDispatcher.Subscribe()` call happens. This could be mitigated by using UUIDs (since they provide good level of entropy) or by using a global, per-client `ReferenceIDGenerator` for all user subscriptions.

internal/serialization/object_data.go
```diff
@@ -121,10 +125,9 @@ func (o *ObjectDataOutput) WriteFloat64(v float64) {
 	o.position += bufutil.Float64SizeInBytes
 }
 
-func (o *ObjectDataOutput) WriteUTF(v string) {
+func (o *ObjectDataOutput) WriteString(v string) {
```
puzpuzpuz, line 128: Client protocol v3 string ser/deser is not compatible with v4. The changes are: first 4 byte now contain byte length, not the number of chars; the serialization now supports UTF-8 standard, incl. some 4-byte chars. The current implementation is also slow (at least, it shows in the `pprof` profile) and could be replaced with `[]bytes(v)`.

In general, the client should include binary compatibility tests in order to make sure that it plays nicely with other Hz clients.

See:
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/nio/serialization/compatibility/BinaryCompatibilityTest.java
https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/test/unit/serialization/BinaryCompatibilityTest.js

internal/cluster/connection.go
puzpuzpuz: We should definitely do something with the current `socketReadLoop()` (and underlying `clientMessageReader.resetBuffer()`) behavior, since it shows as the Top 1 source of allocations in `pprof` memory profile.

Screenshot from 2021-04-26 16-19-33

internal/cluster/connection_manager.go
```diff
@@ -0,0 +1,655 @@
...
+	}()
+	return m.doneCh
+}
+
+func (m *ConnectionManager) NextConnectionID() int64 {
+	return atomic.AddInt64(&m.nextConnID, 1)
+}
+
+func (m *ConnectionManager) GetConnectionForAddress(addr *pubcluster.AddressImpl) *Connection {
+	return m.connMap.GetConnectionForAddr(addr.String())
```
puzpuzpuz, line 217: nit: both `AddressImpl.String()` and `UUID.String()` used in this and other methods of the connection manager, generate litter on each operation. That could be mitigated by caching the string in the struct and returning it from the `String()` method.

internal/event/util.go
yuce: This function is only used internally. Note that there's a problem only with functions attached to structs, since that is just syntactic sugar that Go provides:
```go
type Foo struct {
}
func (*f Foo) Bar() bool {
 return false
}

foo1 := &Foo{}
foo2 := &Foo{}
// reflect.ValueOf(foo1.Bar).Pointer() == reflect.ValueOf(foo2.Bar)
```

Unless we use the the same struct function as a handler ID more than once in the same client, there's no problem.

yuce: User event dispatcher is separate from the internal event dispatcher, so there's no subscription ID collision between them.

internal/serialization/object_data.go
yuce: Thanks for the pointers, I didn't yet deep dive in the serialization code. I'll fix that.

internal/serialization/serialization.go
```diff
@@ -1,87 +1,89 @@
...
 // If it is called with nil, nil is returned.
-func (s *Service) ToData(object interface{}) (serialization.Data, error) {
-	if _, ok := object.(*Data); ok {
-		return object.(*Data), nil
+func (s *Service) ToData(object interface{}) (pubserialization.Data, error) {
+	if _, ok := object.(*SerializationData); ok {
+		return object.(*SerializationData), nil
 	}
-	dataOutput := NewPositionalObjectDataOutput(1, s, s.serializationConfig.IsBigEndian())
+	dataOutput := NewPositionalObjectDataOutput(1, s, s.serializationConfig.BigEndian)
```
puzpuzpuz, line 55: nit: we could avoid some allocations done by `ObjectDataOutput.EnsureAvailable()` if we start with a larger size. Java client uses 4096 bytes as the start size for `ByteArrayObjectDataOutput`, but that's quite aggressive and Java client also uses a thread-local pool to reuse `ByteArrayObjectDataOutput` instances. So, as an alternative option, Go client could use a free-list or `sync.Pool` based pool.

yuce: I've tried a sync.Pool based byte array pool for the serializer. It decreases allocations / used memory quite a bit but makes the overall performance worse. I'll try again once I fix the broken stuff pointed in this review.

internal/cluster/connection_manager.go
yuce: This seems like something Go's GC can easily handle. I think the biggest benefit will be in the client message side as you pointed out.

internal/event/util.go
puzpuzpuz: > User event dispatcher is separate from the internal event dispatcher, so there's no subscription ID collision between them.

That's not obvious from the code and nothing enforces this invariant. This may become a source of subtle bugs in the future. I'd better go with a registry of constants for internal listeners and additional validation of absence of the listener id during the registration.

internal/cluster/connection_manager.go
puzpuzpuz: GC can handle a lot of litter structs for sure. But users will appreciate less stress put on the GC by the library. We're talking of low-hanging fruits here and, since Hazelcast client libraries claim to provide high performance, such optimizations should be included. At least, that's how I see.

puzpuzpuz:
@yuce could you post commit hashes when you're resolving issues? It's very hard to track updates on such big PRs especially when there are multiple reviews in progress.

puzpuzpuz:
> @yuce could you post commit hashes when you're resolving issues? It's very hard to track updates on such big PRs especially when there are multiple reviews in progress.

puzpuzpuz:
Sorry for mistakenly closing the PR. It's my touchpad (and me) to blame 😅

internal/serialization/object_data.go
yuce: Updated at https://github.com/yuce/hazelcast-go-client/commit/33bc62f5a109f751db1af984138d5fa320da3730 and https://github.com/yuce/hazelcast-go-client/commit/17b753d61dc46622c9baad39e079a014db781f8f

internal/cluster/cluster_service.go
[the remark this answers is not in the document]
yuce: I would argue that making the mutex a pointer and initializing it is easier than remembering to not copy it. I'll mark this as resolved, but please feel free to reopen in case we should discuss that further.

internal/invocation/invocation.go
yuce: Updated at: https://github.com/yuce/hazelcast-go-client/commit/925534a086ef8ae094679c0ee9afcd3ee8d51e13

internal/cluster/cluster_service.go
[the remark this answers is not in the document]
puzpuzpuz: No need to reopen since this comment was a `nit` (`n`ot `i`mportant `t`hought), hence it's subjective and optional by the definition.

yuce:
@puzpuzpuz No problem at all. Adding commit hashes...

start-rc.sh
```diff
@@ -1,17 +1,43 @@
 #!/bin/bash
```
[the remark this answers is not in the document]
yuce: Updated at: https://github.com/yuce/hazelcast-go-client/commit/5c7daf7cd3ae981e5862717d78436f1bb448d977

internal/invocation/invocation.go
yuce: Implemented at: https://github.com/yuce/hazelcast-go-client/commit/65e58171e8b415d9f55907da4de1ed46abcb0238

internal/cb/circuitbreaker.go
yuce: Non-retryable errors implemented at: https://github.com/yuce/hazelcast-go-client/commit/e56053c7635cabdc5b9baffabbc8a8dfe0140e59

internal/serialization/object_data.go
```diff
@@ -619,27 +614,18 @@ func (i *ObjectDataInput) readFloat64WithPosition(pos int32) (float64, error) {
...
+		i.err = err
+		return ""
 	}
-	return string(ret), nil
+	s := string(i.buffer[i.position : i.position+size])
+	i.position += size
+	return s
 }
 
 func (i *ObjectDataInput) ReadUTFWithPosition(pos int32) string {
```
puzpuzpuz, line 631: nit: to make naming consistent with `ReadString`, it makes sense to rename this one to `ReadStringWithPosition`.

internal/serialization/object_data.go
puzpuzpuz: Resolving this one since the string serialization is now fixed. Still, it makes sense to plan adding binary compatibility tests for the client.

internal/serialization/serialization.go
```diff
@@ -1,87 +1,89 @@
...
 	serializer, err := s.FindSerializerFor(object)
 	if err != nil {
 		return nil, err
 	}
 	dataOutput.WriteInt32(0) // partition
 	dataOutput.WriteInt32(serializer.ID())
 	err = serializer.Write(dataOutput, object)
-	return &Data{dataOutput.buffer}, err
+	return &SerializationData{dataOutput.buffer[:dataOutput.position]}, err
```
puzpuzpuz, line 63: nit: access to private fields on `dataOutput` seems to be fragile. It makes sense to wrap it in a method on `ObjectDataOutput`:
```go
func (o *ObjectDataOutput) AsBufferSlice() []byte { // exact name doesn't matter
	return o.buffer[:o.position]
}
```

internal/invocation/invocation.go
puzpuzpuz: I'm not sure if changes in 65e58171e8b415d9f55907da4de1ed46abcb0238 are enough for a proper implementation. The thing is that invocation has to be deregistered from the invocation service once the deadline is reached. Java client has `InvocationMonitor` responsible for scanning through pending invocations, detecting deadlines for them (incl. backup timeouts, but that's for boomerang backup feature) and de-registering them. Go client may need to implement something similar, otherwise stale invocations will lead to mem leaks.

See: https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java#L304

puzpuzpuz: Sorry for pointing you at member-side invocation code. The proper link should be this connection closed handler: https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/ClientInvocationServiceImpl.java#L217

internal/cb/circuitbreaker.go
yuce: Updated at: https://github.com/yuce/hazelcast-go-client/commit/ca130a5cf876968824a8c0375daf46e2cab85fd9 https://github.com/yuce/hazelcast-go-client/commit/769d9de62ef62ecd08a2b14eb4c28cd6f14d0364  (older commits with relevant changes: https://github.com/yuce/hazelcast-go-client/commit/9fa5987332f0cf77c5940c8dbd59f3f355e6c143 https://github.com/yuce/hazelcast-go-client/commit/65e58171e8b415d9f55907da4de1ed46abcb0238)

internal/cluster/connection_invocation_handler.go
yuce: Updated at: https://github.com/yuce/hazelcast-go-client/commit/fac635414bfaea419438a7bb8614fb1b2955e45b Currently there's no load balancer apart from random routing.

internal/cb/circuitbreaker.go
```diff
@@ -0,0 +1,138 @@
...
+func (cb *CircuitBreaker) Try(tryHandler TryHandler) Future {
+	return cb.TryContext(context.Background(), tryHandler)
+}
+
+func (cb *CircuitBreaker) TryContext(ctx context.Context, tryHandler TryHandler) Future {
+	if state := atomic.LoadInt32(&cb.State); state == StateOpen {
+		return NewFailedFuture(ErrCircuitOpen)
+	}
+	future := NewFutureImpl()
+	go cb.try(ctx, future.resultCh, tryHandler)
```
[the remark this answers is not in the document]
yuce: I've made async optional for the circuit breaker: https://github.com/yuce/hazelcast-go-client/commit/14f9a0a943a35149f8e0912ef9834b443f0da859

internal/cluster/connection_manager.go
```diff
@@ -0,0 +1,620 @@
...
+		return addr
+	}
+	return nil
+}
+
+func (m *connectionMap) RandomConn() *Connection {
+	m.connectionsMu.RLock()
+	defer m.connectionsMu.RUnlock()
+	for _, conn := range m.addrToConn {
+		// Go randomizes maps, this is random enough for now.
```
puzpuzpuz, line 568: Maps in Go are based on hash tables, so the order on a stable map (i.e. when the are no mutations happening) will be the same among iterations. Hence, on for a stable cluster no randomization will take place. Not sure if that's what this comment implies, but other clients implement load balancers in order to "randomize" the selected connection.

See https://github.com/golang/go/blob/master/src/runtime/map.go#L23 for more details on how maps are implemented internally

internal/cluster/connection_invocation_handler.go
puzpuzpuz: It seems to me that current routing implementation is not random: https://github.com/hazelcast/hazelcast-go-client/pull/552/files#r622204484

internal/cluster/partition_service.go
sancar: @yuce Can you recheck this one ? It seems the method is not still returning an error.

client.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+	c.eventDispatcher.Publish(newLifecycleStateChanged(LifecycleStateShuttingDown))
+	c.invocationService.Stop()
+	c.clusterService.Stop()
+	c.partitionService.Stop()
+	<-c.connectionManager.Stop()
+	atomic.StoreInt32(&c.state, stopped)
+	c.eventDispatcher.Publish(newLifecycleStateChanged(LifecycleStateShutDown))
+	// wait for the shut down ebent to be dispatched
+	time.Sleep(10 * time.Millisecond)
```
sancar, line 217: We should not need this sleep. 
anything pushed to a channel will be consumed before close anyway.

internal/cluster/partition_service.go
yuce: @sancar Sorry, this should fix that: https://github.com/yuce/hazelcast-go-client/commit/faafeaa735e6867faeee4a22c4cd3c03384426d9

internal/cluster/connection_invocation_handler.go
yuce: Thanks, simplified smart routing and fixed invoke at: https://github.com/yuce/hazelcast-go-client/commit/7fae2fb2c319ead787a4968cb6047e6f53e87d20 and https://github.com/yuce/hazelcast-go-client/commit/002631f06968e580970d0adb7966ea7d45d59495

client.go
```diff
@@ -0,0 +1,394 @@
...
+}
+
+// Start connects the client to the cluster.
+func (c *Client) Start() error {
+	if !atomic.CompareAndSwapInt32(&c.state, created, starting) {
+		return ErrClientCannotStart
+	}
+	// TODO: Recover from panics and return as error
+	c.eventDispatcher.Publish(newLifecycleStateChanged(LifecycleStateStarting))
+	clusterServiceStartCh := c.clusterService.Start(c.clusterConfig.SmartRouting)
```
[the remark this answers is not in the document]
yuce: The channel is internal to the client, so it shouldn't be a problem to use. But it is not needed anymore and I've removed it at: https://github.com/yuce/hazelcast-go-client/commit/7fae2fb2c319ead787a4968cb6047e6f53e87d20 and https://github.com/yuce/hazelcast-go-client/commit/9aead68af3e94a97b713104a430aa6d1228b058a

internal/cluster/connection_manager.go
yuce: > the order on a stable map (i.e. when the are no mutations happening) will be the same among iterations. 
That's not correct. See: https://stackoverflow.com/questions/55925822/in-golang-why-are-iterations-over-maps-random

Some sample code:
```go
var m = map[string]int {
	"foo": 1,
	"bar": 2,
	"qoox": 3,
}

func random(m map[string]int) string {
	for k := range m {
		return k
	}
	return ""
}

func main() {
	for i := 0; i < 100; i++ {
		fmt.Print(random(m), " ")
	}
}
```
The output is not deterministic even though there is no mutation on the map.

The relevant code in the client is there until the load balancer is properly implemented.

internal/cluster/connection_invocation_handler.go
yuce: See: https://github.com/hazelcast/hazelcast-go-client/pull/552#discussion_r622753422

client.go
sancar: I just wrote a test noticed that it is not guaranteed that all messages will be consumed. Still we should have a better way than the sleep. This can not guarantee that even will be dispathced

sancar: The following statement is true for channels
`anything pushed to a channel will be consumed before close`
I think what we should do is to consume the eventCh when doneCh is closed, until eventCh starts to return false.  
So adding something the following range loop to where we wait for doneCh should do the trick if I am not mistaken
```
case <-s.doneCh:
			for event := range s.eventCh {
				s.dispatch(event)
			}		
			return
```

internal/cluster/cluster_service.go
yuce: Updated at: https://github.com/yuce/hazelcast-go-client/commit/15136dbdd26d8f9e1acdf9eba6c5ef62be6392d4

client.go
```diff
@@ -0,0 +1,394 @@
...
+
+// Shutdown disconnects the client from the cluster.
+func (c *Client) Shutdown() error {
+	if !atomic.CompareAndSwapInt32(&c.state, ready, stopping) {
+		return ErrClientNotReady
+	}
+	c.eventDispatcher.Publish(newLifecycleStateChanged(LifecycleStateShuttingDown))
+	c.clusterService.Stop()
+	c.partitionService.Stop()
+	<-c.connectionManager.Stop()
```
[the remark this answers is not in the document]
yuce: Updated at: https://github.com/yuce/hazelcast-go-client/commit/7fae2fb2c319ead787a4968cb6047e6f53e87d20

proxy_map.go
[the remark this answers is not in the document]
yuce: Refactored listeners to use UUIDs: https://github.com/yuce/hazelcast-go-client/commit/15136dbdd26d8f9e1acdf9eba6c5ef62be6392d4

internal/event/util.go
yuce: Refactored listener stuff for DSSs: https://github.com/yuce/hazelcast-go-client/commit/15136dbdd26d8f9e1acdf9eba6c5ef62be6392d4

PR #556, opened 2021-05-12
Add List support

puzpuzpuz:
* Adds `List` support
* Introduces common `convertToObjects` method for proxies
* Fixes error handling in `Queue.Add`

Client protocol changes for the new codecs are ad-hoc for now. They will be added into the protocol repo with https://github.com/hazelcast/hazelcast-client-protocol/pull/375 or in a subsequent PR.

### DIFF
--- a/proxy_list.go
+++ b/proxy_list.go
@@ -323,10 +323,10 @@ func (l *List) SubList(start int32, end int32) ([]interface{}, error) {
 
 // Iterator returns all of the items in this list in proper sequence.
 func (l *List) Iterator() ([]interface{}, error) {
-	request := codec.EncodeListGetAllRequest(l.name)
+	request := codec.EncodeListIteratorRequest(l.name)
 	response, err := l.invokeOnPartition(context.TODO(), request, l.partitionID)
 	if err != nil {
 		return nil, err
 	}
-	return l.convertToObjects(codec.DecodeListGetAllResponse(response))
+	return l.convertToObjects(codec.DecodeListIteratorResponse(response))
 }
--- a/proxy_list.go
+++ b/proxy_list.go
@@ -1,0 +1,332 @@
+/*
+ * Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package hazelcast
+
+import (
+	"context"
+
+	"github.com/hazelcast/hazelcast-go-client/internal/proto"
+	"github.com/hazelcast/hazelcast-go-client/internal/proto/codec"
+	"github.com/hazelcast/hazelcast-go-client/serialization"
+	"github.com/hazelcast/hazelcast-go-client/types"
+)
+
+/*
+List is a concurrent, distributed, ordered collection. The user of this
+data structure has precise control over where in the list each element is
+inserted. The user can access elements by their integer index (position in the list),
+and search for elements in the list.
+
+List is not a partitioned Hazelcast data structure. So all the contents of the List are stored in a single
+machine (and in the backup). So, a single List will not scale by adding more members in the cluster.
+*/
+type List struct {
+	*proxy
+	partitionID int32
+}
+
+func newList(p *proxy) (*List, error) {
+	if partitionID, err := p.stringToPartitionID(p.name); err != nil {
+		return nil, err
+	} else {
+		return &List{proxy: p, partitionID: partitionID}, nil
+	}
+}
+
+// AddListener adds an item listener for this list.
+// The listener will be invoked whenever an item is added to or removed from this list.
+// Returns subscription ID of the listener.
+func (l *List) AddListener(handler ListItemNotifiedHandler) (types.UUID, error) {
+	return l.addListener(false, handler)
+}
+
+// AddListener adds an item listener for this list.
+// The listener will be invoked whenever an item is added to or removed from this list.
+// Received events include the updated item.
+// Returns subscription ID of the listener.
+func (l *List) AddListenerIncludeValue(handler ListItemNotifiedHandler) (types.UUID, error) {
+	return l.addListener(true, handler)
+}
+
+func (l *List) addListener(includeValue bool, handler ListItemNotifiedHandler) (types.UUID, error) {
+	subscriptionID := types.NewUUID()
+	addRequest := codec.EncodeListAddListenerRequest(l.name, includeValue, l.config.ClusterConfig.SmartRouting)
+	removeRequest := codec.EncodeListRemoveListenerRequest(l.name, subscriptionID)
+	listenerHandler := func(msg *proto.ClientMessage) {
+		codec.HandleListAddListener(msg, func(itemData serialization.Data, uuid types.UUID, eventType int32) {
+			item, err := l.convertToObject(itemData)
+			if err != nil {
+				l.logger.Warnf("cannot convert data to Go value: %v", err)
+				return
+			}
+			member := l.clusterService.GetMemberByUUID(uuid.String())
+			handler(newListItemNotified(l.name, item, member, eventType))
+		})
+	}
+	err := l.listenerBinder.Add(subscriptionID, addRequest, removeRequest, listenerHandler)
+	return subscriptionID, err
+}
+
+// RemoveListener removes the item listener with the given subscription ID.
+func (l *List) RemoveListener(subscriptionID types.UUID) error {
+	return l.listenerBinder.Remove(subscriptionID)
+}
+
+// Add appends the specified element to the end of this list.
+// Returns true if the list has changed as a result of this operation, false otherwise.
+func (l *List) Add(element interface{}) (bool, error) {
+	elementData, err := l.validateAndSerialize(element)
+	if err != nil {
+		return false, err
+	}
+	request := codec.EncodeListAddRequest(l.name, elementData)
+	response, err := l.invokeOnPartition(context.TODO(), request, l.partitionID)
+	if err != nil {
+		return false, err
+	}
+	return codec.DecodeListAddResponse(response), nil
+}
+
+// AddAt inserts the specified element at the specified index.
+// Shifts the subsequent elements to the right.
+func (l *List) AddAt(index int32, element interface{}) error {
+	elementData, err := l.validateAndSerialize(element)
+	if err != nil {
+		return err
+	}
+	request := codec.EncodeListAddWithIndexRequest(l.name, index, elementData)
+	_, err = l.invokeOnPartition(context.TODO(), request, l.partitionID)
+	return err
+}
+
+// AddAll appends all elements in the specified slice to the end of this list.
+// Returns true if the list has changed as a result of this operation, false otherwise.
+func (l *List) AddAll(elements ...interface{}) (bool, error) {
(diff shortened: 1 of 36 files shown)


proxy_list.go
```diff
@@ -0,0 +1,332 @@
...
+	response, err := l.invokeOnPartition(context.TODO(), request, l.partitionID)
+	if err != nil {
+		return false, err
+	}
+	return codec.DecodeListAddResponse(response), nil
+}
+
+// AddAt inserts the specified element at the specified index.
+// Shifts the subsequent elements to the right.
+func (l *List) AddAt(index int32, element interface{}) error {
```
yuce, line 106: I think `index` should be `int` instead of `int32`. Although the API takes an `int32`, an `int` is more appropriate to denote an index, size, etc.

proxy_list.go
```diff
@@ -0,0 +1,332 @@
...
+	if err != nil {
+		return false, err
+	}
+	return codec.DecodeListAddAllResponse(response), nil
+}
+
+// AddAllAt inserts all elements in the specified slice at specified index, keeping the order of the slice.
+// Shifts the subsequent elements to the right.
+// Returns true if the list has changed as a result of this operation, false otherwise.
+func (l *List) AddAllAt(index int32, elements ...interface{}) (bool, error) {
```
yuce, line 134: The same comment about `int32` vs `int` as the `index`...

proxy_list.go
```diff
@@ -0,0 +1,332 @@
...
+	request := codec.EncodeListContainsAllRequest(l.name, elementsData)
+	response, err := l.invokeOnPartition(context.TODO(), request, l.partitionID)
+	if err != nil {
+		return false, err
+	}
+	return codec.DecodeListContainsAllResponse(response), nil
+}
+
+// Get retrieves the element at given index.
+func (l *List) Get(index int32) (interface{}, error) {
```
yuce, line 185: The same comment about `int32` vs `int` as the `index`...

proxy_list.go
```diff
@@ -0,0 +1,332 @@
...
+	request := codec.EncodeListGetRequest(l.name, index)
+	response, err := l.invokeOnPartition(context.TODO(), request, l.partitionID)
+	if err != nil {
+		return nil, err
+	}
+	return l.convertToObject(codec.DecodeListGetResponse(response))
+}
+
+// IndexOf returns the index of the first occurrence of the given element in this list.
+func (l *List) IndexOf(element interface{}) (int32, error) {
```
yuce, line 195: The same comment about `int32` vs `int` as the `index`...

proxy_list.go
```diff
@@ -0,0 +1,332 @@
...
+	request := codec.EncodeListIsEmptyRequest(l.name)
+	response, err := l.invokeOnPartition(context.TODO(), request, l.partitionID)
+	if err != nil {
+		return false, err
+	}
+	return codec.DecodeListIsEmptyResponse(response), nil
+}
+
+// LastIndexOf returns the index of the last occurrence of the given element in this list.
+func (l *List) LastIndexOf(element interface{}) (int32, error) {
```
yuce, line 219: The same comment about `int32` vs `int` as the `index`...

proxy_list.go
```diff
@@ -0,0 +1,332 @@
...
+	response, err := l.invokeOnPartition(context.TODO(), request, l.partitionID)
+	if err != nil {
+		return false, err
+	}
+	return codec.DecodeListRemoveResponse(response), nil
+}
+
+// RemoveAt removes the element at the given index.
+// Returns the removed element.
+func (l *List) RemoveAt(index int32) (interface{}, error) {
```
yuce, line 249: The same comment about `int32` vs `int` as the `index`...

proxy_list.go
```diff
@@ -0,0 +1,332 @@
...
+	response, err := l.invokeOnPartition(context.TODO(), request, l.partitionID)
+	if err != nil {
+		return false, err
+	}
+	return codec.DecodeListCompareAndRetainAllResponse(response), nil
+}
+
+// Set replaces the element at the specified index in this list with the specified element.
+// Returns the previous element from the list.
+func (l *List) Set(index int32, element interface{}) (interface{}, error) {
```
yuce, line 290: The same comment about `int32` vs `int` as the `index`...

proxy_list.go
```diff
@@ -0,0 +1,332 @@
...
+	request := codec.EncodeListSetRequest(l.name, index, elementData)
+	response, err := l.invokeOnPartition(context.TODO(), request, l.partitionID)
+	if err != nil {
+		return nil, err
+	}
+	return l.convertToObject(codec.DecodeListSetResponse(response))
+}
+
+// Size returns the number of elements in this list.
+func (l *List) Size() (int32, error) {
```
yuce, line 304: The same comment about `int32` vs `int` as the `index`, but for the return value...

proxy_list.go
```diff
@@ -0,0 +1,332 @@
...
+	response, err := l.invokeOnPartition(context.TODO(), request, l.partitionID)
+	if err != nil {
+		return 0, err
+	}
+	return codec.DecodeListSizeResponse(response), nil
+}
+
+// SubList returns a view of this list that contains elements between index numbers
+// from start (inclusive) to end (exclusive).
+func (l *List) SubList(start int32, end int32) ([]interface{}, error) {
```
yuce, line 315: The same comment about `int32` vs `int` as the `index`...

proxy_list.go
```diff
@@ -0,0 +1,332 @@
...
+	request := codec.EncodeListSubRequest(l.name, start, end)
+	response, err := l.invokeOnPartition(context.TODO(), request, l.partitionID)
+	if err != nil {
+		return nil, err
+	}
+	return l.convertToObjects(codec.DecodeListSubResponse(response))
+}
+
+// ToSlice returns a slice that contains all elements of this list in proper sequence.
+func (l *List) ToSlice() ([]interface{}, error) {
```
yuce, line 325: Could you rename this function to `GetAll` to keep the name-compatibility within the client and with other clients?

yuce (changes requested):
Thanks for the PR. It looks good except:
* Some missing error handling in tests,
* `int32` vs `int` for indices and sizes,
* `getAll` and `iterator` seems to be doing the same thing in other clients but since they are using different codecs, I think we should have both of them.
* The function that corresponds to `ListListIterator` is missing. I think `IteratorAt` could be a good name for it (other clients call it `listIterator`, etc.). What do you think?

puzpuzpuz:
> Thanks for the PR.

My pleasure.

> * Some missing error handling in tests,

Added those in 15ea36d69509b8016cb3afa50153ac2c3102ec5f

> * `int32` vs `int` for indices and sizes,

Will reply inline in the corresponding comments.

> * `getAll` and `iterator` seems to be doing the same thing in other clients but since they are using different codecs, I think we should have both of them.

The PR is almost one to one port of the old code base, so that's why the List API is kept as it used to be in v0.x.

> * The function that corresponds to `ListListIterator` is missing. I think `IteratorAt` could be a good name for it (other clients call it `listIterator`, etc.). What do you think?

v0.x client didn't have it and I don't think iterators are idiomatic in Go. I can see that there is `Queue.Iterator()` method, but it does the same thing as `List.ToSlice()`. I've renamed the method into `List.Iterator()` for the sake of consistency (although, I find the name to be misleading). See 02d893e68d91cc695dad90b583c56810eb551ef3

As for having two separate methods for the same thing, I don't think it makes sense.

puzpuzpuz:
BTW is Jenkins CI build meant to be broken? Also, CI build configuration seems to be missing a linter run (say, golint + go vet).

proxy_list.go
puzpuzpuz: Renamed into `Iterator()` for the sake of consistency with `Queue.Iterator()`.

proxy_list.go
puzpuzpuz: Java's `j.u.List` supports signed 32-bit integer indexes, so in case if we accept `int` users may face overflow issues. I think, it's better to be precise about the expected number format to avoid such issues.

proxy_list.go
puzpuzpuz: For the returned indexes we may use `int`, but I'm not sure if it's worth it. Having the same type for in and out values has the should be more obvious for the user due to the symmetry.

yuce:
> 
> 
> BTW is Jenkins CI build meant to be broken? Also, CI build configuration seems to be missing a linter run (say, golint + go vet).

We don't use Jenkins anymore, but I couldn't find a way to remove it in the Github UI. Last time I ran golint, it threw quite a few not so-useful warnings, so it is not enabled yet. I use https://staticcheck.io before which works OK with out code base. I will integrate it in the optimization PR or a separate one.

puzpuzpuz:
> We don't use Jenkins anymore, but I couldn't find a way to remove it in the Github UI.

Probably, it makes sense to ask someone with admin permissions for some help. Those false positive failed run notifications are distracting.

yuce:
The function that corresponds to `ListListIterator` takes an optional index parameter and seem to work slightly different from `GetAll`/`Iterator`.  As far as I see, it doesn't exist in the PR, I can add it in case you don't have the time right now.

`GetAll` and `Iterator` seem to do the same thing, but their codecs have different names. If in the future they do different things on the server side, then we would get stuck with `list.Iterator` calling the `GetAll` codec and `list.GetAll` (?) calling the `Iterator` codec :)

yuce:
re: `int32` vs `int`. I think the best way to handle that is using `int` for all indexes (and sizes) but check them against overflow. It's OK to keep them as is for now.

puzpuzpuz:
> The function that corresponds to `ListListIterator` takes an optional index parameter and seem to work slightly different from `GetAll`/`Iterator`. As far as I see, it doesn't exist in the PR, I can add it in case you don't have the time right now.

That codec is dictated by Java's [`j.u.List` interface](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#listIterator-int-) and we don't use it in all non-Java clients. It makes sense to reconsider the API for each client based on the language and standard library. For instance, [Node.js client](https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/src/proxy/ListProxy.ts) doesn't have `iterator()`/`iterator(int)` methods. Also, `SubList()` method allows doing quite the same thing, as long as the list is stable and you know its size.

But if there is a consensus in the clients team that this method has to be present in all clients, I'll add it into this PR.

> If in the future they do different things on the server side, then we would get stuck with `list.Iterator` calling the `GetAll` codec and `list.GetAll` (?) calling the `Iterator` codec :)

First of all, that sounds like a breaking change. :)

Then, I'd better keep the `ToSlice()` name since Go doesn't have the notion of iterators and what `Queue/List.Iterator()` returns is not an iterator in any sense. Anyhow, I've changed the underlying codec to the `*Iterator` one in e8785af834e8b85cdd97ea0096f93f3d32dc9e87

> OK to keep them as is for now.

👍

puzpuzpuz:
As for `int32` -> `int` I can apply the suggestion with input overflow validation to both `List` and `Queue` and include it into this PR. WDYT?

yuce:
I agree with making the API as idiomatic Go as possible and removing duplicate functions. In the Queue implementation there is `Offer` and `Alias` which is almost an alias for `Offer` so we decided to use `Add` (since what `Add` does is more clear to non-Java folks). But in this case `GetAll` and `Iterator` has different codecs (not sure about the history of that decision) and `ListIterator` is also very similar in the default case.

I think it's OK to keep it as `ToSlice` for this PR. We can decide on whether to rename that (and also `Queue.Iterator`) next week (after the holiday here).

> As for int32 -> int I can apply the suggestion with input overflow validation to both List and Queue and include it into this PR.

Thanks, that would be great!

puzpuzpuz:
Reverted `ToSlice()` method rename in c17aa9418b5faf6679ad0ae25d6d0d48dc4f6c84

Converted args from `int32` to `int` in `List` and `Queue` methods in the following commits respectively: 695a5bacaa3e976a589544df1ea2c0b20f231468, e34e438c52b7c043a9a34ef080fa7da71dcbfb88

yuce (approved):
Thanks, looks great!

puzpuzpuz:
@yuce many thanks for the review!

yuce:
@puzpuzpuz Thanks a lot for the PR!

PR #557, opened 2021-05-13
Purge proxies from local cache on Destroy()

puzpuzpuz:
* Fixes the issue with proxies not being purged from the local proxy manager cache on Destroy() which led to a memory leak
* Also fixes some code issues, like unused consts, reported by [staticcheck](https://staticcheck.io/) linter

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -31,7 +31,7 @@ func TestClient_NewConfigBuilder(t *testing.T) {
 	if err != nil {
 		t.Error(err)
 	}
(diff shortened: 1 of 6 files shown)


puzpuzpuz:
@yuce could you take a look at this one when you return from holiday and have some time? It's not urgent at all and can wait.

yuce (approved):
Thanks, looks good!

puzpuzpuz:
@yuce thanks for the quick review!

PR #558, opened 2021-05-14
Improve test stability

puzpuzpuz:
* Increases sleep time for TTL/idleTime tests. TTL/idleTime is handled with second granularity on cluster side, so previous timeouts weren't enough under certain conditions.
  - I've faced a test failure in one of those tests when running test locally.
* Adds rand seed initialization for tests, so that generated data structure names used in test runners have better entropy.
* Adds destroy step for data structures used in test runners in order to save some memory in remote controller's JVM.


yuce (approved):
Thanks!

puzpuzpuz:
@yuce thanks for the review!

PR #559, opened 2021-05-17
Fix flaky test for Map.GetKeySetWithPredicate()

puzpuzpuz:
* Fixes flaky `Map.GetKeySetWithPredicate()` test which assumes the same order in the expected and actual key sets

One of failures: https://github.com/hazelcast/hazelcast-go-client/runs/2598042429

### DIFF
--- a/map_it_test.go
+++ b/map_it_test.go
@@ -332,8 +332,8 @@ func TestMap_GetKeySetWithPredicate(t *testing.T) {
 		it.Must(m.Set(serialization.JSON(`{"a": 15}`), "v3"))
 		if keys, err := m.GetKeySetWithPredicate(predicate.GreaterOrEqual("__key.a", 10)); err != nil {
 			t.Fatal(err)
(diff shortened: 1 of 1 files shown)


yuce:
@puzpuzpuz Thanks for the PR. It seems there's an easier way using testify, fixed this one at: https://github.com/hazelcast/hazelcast-go-client/pull/555/commits/e76bf9208f007c2b3c52fc6eadb3337726edf268

puzpuzpuz:
> @puzpuzpuz Thanks for the PR. It seems there's an easier way using testify, fixed this one at: [e76bf92](https://github.com/hazelcast/hazelcast-go-client/commit/e76bf9208f007c2b3c52fc6eadb3337726edf268)

I'm aware of testify's utilities, but I intentionally kept the change consistent with the rest of the test code base where in-house utilities are favored instead of testify ones.

I'm closing this PR then. But I don't like the idea of adding more commits into #555 which already contains tons of stuff unrelated with optimizations making it harder to review.

PR #563, opened 2021-05-20
Updates error codes for Hz 4

yuce:
- Unexports error codes
- Fixes a few tests


hzerrors/codes.go
```diff
@@ -16,100 +16,107 @@
...
-	ErrorCodeStaleTaskID                      ErrorCode = 80
-	ErrorCodeDuplicateTask                    ErrorCode = 81
-	ErrorCodeStaleTask                        ErrorCode = 82
-	ErrorCodeLocalMemberReset                 ErrorCode = 83
-	ErrorCodeIndeterminateOperationState      ErrorCode = 84
-	ErrorCodeFlakeIDNodeIDOutOfRangeException ErrorCode = 85
-	ErrorCodeTargetNotReplicaException        ErrorCode = 86
-	ErrorCodeMutationDisallowedException      ErrorCode = 87
-	ErrorCodeConsistencyLostException         ErrorCode = 88
+	errorCodeUndefined                        errorCode = 0
```
mdumandag, line 22: Are these error codes the same as the ones used int https://github.com/hazelcast/hazelcast-go-client/blob/master/hzerrors/server_error.go#L20 ?

yuce: The `ServerError`s error code is coming from the decoder, then it is matched at https://github.com/hazelcast/hazelcast-go-client/blob/a68b383a9367e604696bfe736f4bba8ebf504f2d/hzerrors/error.go#L291

mdumandag: But this is also a public API, right? I mean the this method https://github.com/hazelcast/hazelcast-go-client/blob/master/hzerrors/server_error.go#L45 

If we are making this public, I think we should also make the error codes public. In the previous PR, I just wanted to say that this information is private in the Java client

yuce: `ErrorCode` is public, since it has to be used in `socketReadLoop` in `internal/cluster`. The error stuff should be refactored to use plain Go errors, instead of structs, but I'll do that in a separate PR.

mdumandag (approved):

PR #565, opened 2021-05-21
Added Checkers, Getting Ready for P2 Release

yuce:
- Fixed staticcheck warnings
- Re-enabled heartbeats


internal/serialization/object_data_test.go
```diff
@@ -214,9 +203,9 @@ func TestObjectDataInput_ReadUTF2(t *testing.T) {
 	o := NewObjectDataOutput(0, nil, false)
 	o.WriteString("Furkan Şenharputlu")
 	o.WriteString("Jack")
-	o.WriteString("⚐中💦2😭‍🙆😔")
```
puzpuzpuz, line 217: nit: what is the reason for this change? Having multiple UTF-8 code point chars is better than a single one.

puzpuzpuz (approved):

internal/serialization/object_data_test.go
yuce: staticcheck complains about using unicode code instead of some non-printable char. I'll add a few more chars.

PR #569, opened 2021-05-24
Refactored class definition

yuce:
- Makes `ClassDefinition` and `FieldDefinition` public and `struct`.
- Exports All fields of `FieldDefinition`,
- Removes `ClassDefinitionBuilder`


serialization/class_definition.go
```diff
@@ -0,0 +1,177 @@
...
+func NewClassDefinition(factoryID int32, classID int32, version int32) *ClassDefinition {
+	return &ClassDefinition{
+		FactoryID: factoryID,
+		ClassID:   classID,
+		Version:   version,
+		Fields:    make(map[string]FieldDefinition),
+	}
+}
+
+func (cd *ClassDefinition) AddField(definition FieldDefinition) {
```
mdumandag, line 38: Just to note, this API is private on Java

mdumandag: Also, I feel like exposing this to the user brings a burden to him. One needs to maintain correct indexes on FieldDefinitions, put factory id, class id, version, etc. to every definition ...

Do you see any advantage of this API over the simple `AddXxxField` methods?

serialization/class_definition.go
```diff
@@ -0,0 +1,177 @@
...
+func (cd *ClassDefinition) AddPortableField(fieldName string, def *ClassDefinition) error {
+	if def.ClassID == 0 {
+		return hzerrors.NewHazelcastIllegalArgumentError("Portable class ID cannot be zero", nil)
+	}
+	cd.AddField(NewFieldDefinition(int32(len(cd.Fields)), fieldName, TypePortable, def.FactoryID, def.ClassID, cd.Version))
+	return nil
+}
+
+func (cd *ClassDefinition) addNewFieldDefinition(fieldName string, fieldType int32) {
+	cd.AddField(NewFieldDefinition(int32(len(cd.Fields)), fieldName, fieldType, 0, 0, cd.Version))
```
mdumandag, line 87: In Java we do not permit to use of duplicate field names

serialization/class_definition.go
```diff
@@ -0,0 +1,177 @@
...
+type FieldDefinition struct {
+	Name      string
+	Index     int32
+	Type      int32
+	FactoryID int32
+	ClassID   int32
+	Version   int32
+}
+
+func NewFieldDefinition(index int32, fieldName string, fieldType int32, factoryID int32,
```
mdumandag, line 167: I think users shouldn't be able to create field definitions themselves

serialization/class_definition.go
```diff
@@ -0,0 +1,177 @@
...
+	TypeInt32Array
+	TypeInt64Array
+	TypeFloat32Array
+	TypeFloat64Array
+	TypeStringArray
+)
+
+// FieldDefinition defines name, type, index of a field.
+type FieldDefinition struct {
+	Name      string
```
mdumandag, line 159: is there a way to make the struct fields immutable, without sacrificing their simplicity to access in Go?

serialization/class_definition.go
```diff
@@ -0,0 +1,177 @@
...
+
+func (cd *ClassDefinition) AddFloat64Field(fieldName string) {
+	cd.addNewFieldDefinition(fieldName, TypeFloat64)
+}
+
+func (cd *ClassDefinition) AddStringField(fieldName string) {
+	cd.addNewFieldDefinition(fieldName, TypeString)
+}
+
+func (cd *ClassDefinition) AddPortableField(fieldName string, def *ClassDefinition) error {
```
mdumandag, line 78: It seems weird to me that this function returns an error but other similar ones do not

serialization/class_definition.go
```diff
@@ -0,0 +1,177 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package serialization
+
+import "github.com/hazelcast/hazelcast-go-client/hzerrors"
+
+// ClassDefinition defines a class schema for Portable structs.
+type ClassDefinition struct {
```
mdumandag, line 22: One thing we lost with this change is that, with the builder approach, we know that the class definition is immutable and correct. We might lose parts of these guarantees in exchange for a simpler API. It might be OK to do so, but I just wanted to write this

serialization/class_definition.go
```diff
@@ -0,0 +1,177 @@
...
+	TypeFloat32Array
+	TypeFloat64Array
+	TypeStringArray
+)
+
+// FieldDefinition defines name, type, index of a field.
+type FieldDefinition struct {
+	Name      string
+	Index     int32
+	Type      int32
```
mdumandag, line 161: What do you think about creating a type alias for int32 that is called `FieldType` and using it for this?

My point is that `Type` is not an arbitrary int32, there are some fixed values of it and I want to see this information reflected in API

examples/serialization/portable/main.go
```diff
@@ -0,0 +1,120 @@
...
+func (e EmployeePortable) ClassID() int32 {
+	return employeeClassID
+}
+
+func (e EmployeePortable) WritePortable(writer serialization.PortableWriter) {
+	writer.WriteString("name", e.Name)
+	writer.WriteInt32("age", e.Age)
+}
+
+func (e *EmployeePortable) ReadPortable(reader serialization.PortableReader) {
```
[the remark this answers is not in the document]
yuce: Pointer receiver allows the the function to change the struct they are attached to. Since `ReadPortable` mutates the fields of `EmployeePortable` it should take a pointer receiver.

examples/serialization/portable/main.go
```diff
@@ -0,0 +1,120 @@
...
+		return &EmployeePortable{}
+	}
+	return nil
+}
+
+func (m MyPortableFactory) FactoryID() int32 {
+	return factoryID
+}
+
+type BaseObject struct {
```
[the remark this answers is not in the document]
yuce: Yeah, it's the same example I used for the documentation snippets for the main Hazelcast documentation, I've updated the sample to use `BaseObject` at: https://github.com/hazelcast/hazelcast-go-client/pull/569/commits/f39d376439a26e23e6fc6ef356c49f45f8be4821

serialization/class_definition.go
yuce: It's being used in an internal package, so it should be exported. It's possible to make it an internal standalone function, but quite a few tests depend on it. I think it would be better to create a new PR when we do that.

serialization/class_definition.go
yuce: I've unexported `NewFieldDefinition` at: https://github.com/hazelcast/hazelcast-go-client/pull/569/commits/886c827b7a16d536cd96df45676e377f730bc0e0

serialization/class_definition.go
yuce: The same here, the added field ends up at: `cd.Fields[definition.Name] = definition`.

serialization/class_definition.go
yuce: There's nothing like read-only fields. But there's no need to make `FieldDefinition` immutable / limit access, since it gets copied before use. User modifications after passing the configuration aren't visible to the internals.

serialization/class_definition.go
yuce: Thanks, `ClassDefinition` is copied before actually used internally, so user modifications don't matter.

serialization/class_definition.go
yuce: Added that at: https://github.com/hazelcast/hazelcast-go-client/pull/569/commits/8febb814f4129f77c0f6fad7b69108ff40beeba5

serialization/class_definition.go
yuce: Other `AddXXX` functions (except `AddPortableArrayField`) don't produce errors.

examples/serialization/portable/main.go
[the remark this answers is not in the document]
mdumandag: Forgive my Golang ignorance but I tried the following example https://play.golang.org/p/ZKLaWi7mf3u
Basically, I defined an interface with two methods and a struct. I have implemented one interface method with a pointer receiver and one with a struct receiver. The interface check says that my struct does not implement the interface.

So, my question is, if we do that, does the `EmployeePortable` passes this check https://github.com/hazelcast/hazelcast-go-client/blob/18e5440f55cdc69832bdaa856acfd4008e4d4278/internal/serialization/serialization.go#L285 ?

serialization/class_definition.go
mdumandag: But it is a bit different here. So, in Java if someone does this, it will fail.
```java
new ClassDefinitionBuilder(1, 1,1)
              .addBooleanField("a")
              .addIntField("a")
              .build();
```

If I am not missing something, with this API, the following is possible in Go
```go

cd := NewClassDefinition(1, 1, 1)
cd.AddBooleanField("a")
cd.AddInt32Field("a")
```

serialization/class_definition.go
mdumandag: I am okay with doing that later, maybe we can create an issue to not forget to make this private before the stable release

serialization/class_definition.go
mdumandag: One more newby-Go-reviewer question :smile: 

Can I use the function defined on the internal package here and remove the `newFieldDefinition` from `serialization/class_definition.go` to avoid code duplication?

examples/serialization/portable/main.go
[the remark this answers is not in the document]
yuce: In your example, `baz` takes a pointer, so the type assertation requires a pointer-type value. If you change your code to use `&F{}` instead of `F{}` then it will print true: https://play.golang.org/p/JUyNLkyvqqR

In the sample code, the pointer to the actual value is passed to `Set` so there's no problem:
```
employee := &EmployeePortable{Name: "Jane Doe", Age: 30}
```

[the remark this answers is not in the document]
mdumandag: Oh, I see. Thanks for the clarification

examples/serialization/portable/main.go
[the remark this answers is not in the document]
yuce: Btw, I've done some trivial cleanup at: https://github.com/hazelcast/hazelcast-go-client/pull/569/commits/e0678370fa6d58a46f78a2448c43398072559181

serialization/class_definition.go
yuce: Np at all, thanks a lot for the review! :)

Internal `serialization` package depends on the public one, so it's not possible to put `NewFieldDefinition` into `internal/serialization`. It's possible to move it to another package, or a new internal package, but I think that would remove it from the package it is used, so I don't like that idea much.

serialization/class_definition.go
yuce: OK, I've added a check for that and return `error`s from AddXXX fields :(( https://github.com/hazelcast/hazelcast-go-client/pull/569/commits/3c294702ef9bc90212e7d5e65fe10896a80e757d

mdumandag (approved):

PR #570, opened 2021-05-25
Rename ErrClientNotReady to ErrClientNotActive

yuce:


sancar:
There is a separate `HazelcastClientNotActiveError which we should not. There should be only one. 
https://github.com/hazelcast/hazelcast-go-client/blob/cea917a1ef9cbeb5dce69682c0dc1a50b2209afc/hzerrors/error.go#L97

It looks like there two types of errors we have. 
One is in the hzerror which is mostly coming from remote. They implement `HazelcastError` 
They are named as *Error. To check if it is `HazelcastClientNotActiveError` user needs to do a type check.

And we have the ones created in other files like in erros.go.  
```
ErrClientOffline             = errors.New("client offline")	
ErrClientNotAllowedInCluster = errors.New("client not allowed in cluster")	
ErrAddressNotFound           = errors.New("address not found")
```
or in the clients.go
```
ErrClientCannotStart = errors.New("client cannot start")
ErrClientNotActive   = errors.New("client not active")
ErrContextIsNil      = errors.New("context is nil")
```
These are values. To check if something is `ErrClientOffline`, user needs to do a value equality check. 

Can't we go with single strategy here ? I think, to user the distinction is really not clear, and hard to understand. Since we want to deliver different messages/information for the same type in some cases, we can not make all values. So I think we can convert all to `HazelcastError`, where user will always do a type check. 

Also the `func NewHazelcastError(err *ServerError) HazelcastError {` func on hzerrors/error.go does not look complete. It should check for all the types. 
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientProtocolErrorCodes.java

yuce:
I've added another commit which removes `HazelcastClientNotActiveError`. Simple errors are preferred in Go, and they should be checked with `errors.Is` or `errors.As`. I didn't yet refactor the error stuff, but having a different type for each error does not make sense. Ideally we should have a single error type, `HazelcastError` for errors originating from the server side.

sancar:
Keep in mind that, for the erros originating from the server, we should be able to check the type, because depending on the type we take different actions. It does not have to be via different type, but we can check the errorcode as well if that is more suitable for GO. But we should not loose that information.

sancar (approved):

yuce:
We can definitely have "classes" of errors. I'll take another when refactoring the error stuff.

PR #579, opened 2021-06-15
PN Counter [API-481]

yuce:
- Adds PN Counter support


internal/proxy/vector_clock.go
```diff
@@ -0,0 +1,96 @@
...
+				return true
+			}
+			continue
+		}
+		return false
+	}
+	return len(other) < len(vc)
+}
+
+func (vc VectorClock) FactoryID() int32 {
```
asimarslan, line 82: we don't need Vector clock serialization support from client.

yuce: Removed that at: https://github.com/hazelcast/hazelcast-go-client/pull/579/commits/45d165ca618615701bd1d8215b3b24b40c93bd4c

proxy_pn_counter.go
```diff
@@ -0,0 +1,187 @@
...
+
+func (pn *PNCounter) clockEntrySet() []proto.Pair {
+	pn.mu.Lock()
+	entries := pn.clock.EntrySet()
+	pn.mu.Unlock()
+	return entries
+}
+
+func (pn *PNCounter) crdtOperationTarget() (*cluster.MemberInfo, error) {
+	if pn.target == nil {
```
sancar, line 154: I think we need a synchronization here. 
Two concurrent operations on the same pn counter should not be able to select different targets. 

Why don't we follow what we have in java. There are more logic there regarding excluded addresses and how do we choose a new address. It is battle tested and lots of fixes. 
Now I checked, even 3.x go client has those fixes. 

Can't we just copy old client proxies and modify them to compile with new structure, instead of reimplementing every single feature. This will be faster and safer.

yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/579/commits/031b2fc863f9c4a1d1c0f0e4fd78465c9ab6ea8b

The loop and retry logic in the Java client fits very well with the tryInvoke logic I've used with other DSSs, so refactored that part of the code to make use of excluded addresses.

proxy_pn_counter.go
```diff
@@ -0,0 +1,227 @@
...
+
+func (pn *PNCounter) clockEntrySet() []proto.Pair {
+	pn.mu.Lock()
+	entries := pn.clock.EntrySet()
+	pn.mu.Unlock()
+	return entries
+}
+
+func (pn *PNCounter) crdtOperationTarget(excluded map[cluster.Address]struct{}) (*cluster.MemberInfo, error) {
+	target := pn.target.Load()
```
sancar, line 154: It looks like two concurrent callers of this method my get two different members as return value. I think we need more synchronization here.

internal/cluster/cluster_service.go
```diff
@@ -139,7 +147,7 @@ func (s *Service) handleMembersUpdated(conn *Connection, version int32, memberIn
 
 func (s *Service) sendMemberListViewRequest(ctx context.Context, conn *Connection) error {
 	request := codec.EncodeClientAddClusterViewListenerRequest()
-	inv := s.invocationFactory.NewConnectionBoundInvocation(request, -1, nil, conn, func(response *proto.ClientMessage) {
+	inv := s.invocationFactory.NewConnectionBoundInvocation(request, -1, "", conn, func(response *proto.ClientMessage) {
```
sancar, line 150: It feels like this `NewConnectionBoundInvocation` should not take `partition id` and `address` as a parameter in the first place. 
It is not legal to set those to a connection bound invocation.

yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/579/commits/5801b4149d02b4f7937c5e3d37095d3eda39c742

proxy_pn_counter.go
yuce: Fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/579/commits/34cf0b6e9351b5b899f65d27f596ce2a060aa5a5

sancar (approved):

PR #580, opened 2021-06-15
Load Balancer Implementation [API-499]

yuce:
- Adds load balancer
- Moves errors in `hazelcast` package to `hzerrors`.


cluster/load_balancer.go
```diff
@@ -0,0 +1,72 @@
...
+	return &index
+}
+
+// OneOf selects the next address in order from the given address list.
+func (r *RoundRobinLoadBalancer) OneOf(addrs []Address) Address {
+	index := int(*r)
+	if len(addrs) <= index {
+		// the client lost some of the connections
+		// reset it to the last index
+		index = len(addrs) - 1
```
srknzl, line 49: I think if we reset the index to 0, the load balancer will be more fair

Quick demo:

let numbers represent visit order, `o` represent not visited

assume there are 8 addresses at the beginning and first 5 of them are visited

1 2 3 4 5 o o o

then addresses drop to 4

1 2 3 4

Now, If load balancer visits 1 first, it would be fairer since it is visited earliest.

cluster/load_balancer.go
```diff
@@ -0,0 +1,72 @@
...
+func NewRandomLoadBalancer() *RandomLoadBalancer {
+	r := rand.New(rand.NewSource(time.Now().Unix()))
+	lb := RandomLoadBalancer(*r)
+	return &lb
+}
+
+// OneOf selects a random address from the given list and returns it
+func (lb *RandomLoadBalancer) OneOf(addrs []Address) Address {
+	r := rand.Rand(*lb)
+	index := (&r).Intn(len(addrs))
```
srknzl, line 70: is this (&r) needed instead of just r?

internal/cluster/connection_manager.go
```diff
@@ -520,89 +524,106 @@ func (m *ConnectionManager) detectFixBrokenConnections() {
...
+	if _, ok := m.connToAddr[conn.connectionID]; ok {
+		return
+	}
 	m.addrToConn[addr] = conn
 	m.connToAddr[conn.connectionID] = addr
-	m.connectionsMu.Unlock()
+	m.addrs = append(m.addrs, addr)
 }
 
 // RemoveConnection removes a connection and returns true if there are no more addrToConn left.
```
srknzl, line 557: I think `returns true if there are no more addrToConn left.` is leftover. Should be `returns number of connections left`

internal/cluster/connection_manager.go
```diff
@@ -520,89 +524,106 @@ func (m *ConnectionManager) detectFixBrokenConnections() {
...
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	if len(m.addrs) == 0 {
+		return nil
+	}
+	addr := m.lb.OneOf(m.addrs)
+	conn := m.addrToConn[addr]
+	if conn != nil {
+		return conn
+	}
```
srknzl, line 606: Should we need to check here, if client is smart?

internal/cluster/connection_manager.go
```diff
@@ -613,32 +634,42 @@ func (m *connectionMap) FindAddedAddrs(members []pubcluster.MemberInfo, cs *Serv
...
+	m.mu.RUnlock()
 	return l
 }
+
+func (m *connectionMap) removeAddr(addr pubcluster.Address) {
+	for i, a := range m.addrs {
+		if a.Equal(addr) {
+			// note that this changes the order of addresses
+			m.addrs[len(m.addrs)-1] = m.addrs[i]
+			m.addrs = m.addrs[:len(m.addrs)-1]
```
srknzl, line 672: does not this always remove the last connection? You set the last connection to be m.addrs[i] then you discard the last element.

cluster/load_balancer_test.go
```diff
@@ -0,0 +1,66 @@
...
+	assert.Equal(t, cluster.Address("c:5701"), addr)
+}
+
+func TestNewRandomLoadBalancer(t *testing.T) {
+	var lb cluster.LoadBalancer = cluster.NewRandomLoadBalancer()
+	addr := lb.OneOf([]cluster.Address{"a:5071", "b:5701", "c:5701"})
+	if addr.Equal("") {
+		t.Fatalf("one of the addresses must be returned")
+	}
+}
```
srknzl, line 52: You may write

```go
func TestNewRandomLoadBalancer(t *testing.T) {
	var lb cluster.LoadBalancer = cluster.NewRandomLoadBalancer()
	targetAddrs := []cluster.Address{"a:5071", "b:5701", "c:5701"}
	addr := lb.OneOf(targetAddrs)
	assert.Contains(t, targetAddrs, addr)
}
```

cluster/load_balancer_test.go
```diff
@@ -0,0 +1,66 @@
...
+	r := rand.New(rand.NewSource(5))
+	rlb := cluster.RandomLoadBalancer(*r)
+	lb = &rlb
+	targetAddrs := []cluster.Address{"a:5071", "b:5701", "b:5701", "b:5701", "a:5071"}
+	var addrs []cluster.Address
+	for i := 0; i < 5; i++ {
+		addrs = append(addrs, lb.OneOf([]cluster.Address{"a:5071", "b:5701", "c:5701"}))
+	}
+	assert.Equal(t, targetAddrs, addrs)
+}
```
srknzl, line 66:
```suggestion
func TestRandomLoadBalancer_OneOf(t *testing.T) {
	// seed the random load balancer with a fixed int
	r := rand.New(rand.NewSource(5))
	lb := cluster.RandomLoadBalancer(*r)
	targetAddrs := []cluster.Address{"a:5071", "b:5701", "b:5701", "b:5701", "a:5071"}
	var addrs []cluster.Address
	for i := 0; i < 5; i++ {
		addrs = append(addrs, lb.OneOf([]cluster.Address{"a:5071", "b:5701", "c:5701"}))
	}
	assert.Equal(t, targetAddrs, addrs)
}
```

internal/cluster/connection_manager.go
yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/580/commits/8ff2f4a9dc5ac6228040a9b24a7045c40fa06bcd

internal/cluster/connection_manager.go
yuce: When the client is non-smart, there is only a single connection in connection map, so `OneOf` should return the only address in the map. But I've made that more explicit at: https://github.com/hazelcast/hazelcast-go-client/pull/580/commits/8ff2f4a9dc5ac6228040a9b24a7045c40fa06bcd

internal/cluster/connection_manager.go
yuce: Good catch, fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/580/commits/8ff2f4a9dc5ac6228040a9b24a7045c40fa06bcd

cluster/load_balancer_test.go
yuce: That's possible, but I prefer the original code, since it is more explicit.

cluster/load_balancer_test.go
yuce:
```
var lb cluster.LoadBalancer
rlb := cluster.RandomLoadBalancer(*r)
`lb = &rlb
```
seems unnecessary, but that's there to make sure that `RandomLoadBalancer` implements the `LoadBalancer` interface.

cluster/load_balancer.go
yuce: Just `r` is enough, updated at: https://github.com/hazelcast/hazelcast-go-client/pull/580/commits/26f96a67efca245e3882150a7f60999476782a75

cluster/load_balancer_test.go
srknzl: Ok, but I wonder when a load balancer would return an empty string.

Is not my suggestion more specific?  For example, if I change the random load balancer implementation to this, your test will pass:
```go
// OneOf selects a random address from the given list and returns it
func (lb *RandomLoadBalancer) OneOf(addrs []Address) Address {
	r := rand.Rand(*lb)
	index := (&r).Intn(len(addrs))
	return Adress()
}

```

Anyway, I leave the final decision to you. Just wanted to comment

srknzl (approved):
LGTM thanks

yuce:
@srknzl Thanks for the review!

PR #581, opened 2021-06-18
Serialization Improvements [API-32]

yuce:
- Adds more builtin types
- Refactors default serialization to use type switches instead of reflection


internal/serialization/serialization.go
```diff
@@ -83,16 +86,20 @@ func (s *Service) ToObject(data *Data) (r interface{}, err error) {
...
 		}
 	}()
+	var ok bool
 	if data == nil {
 		return nil, nil
 	}
 	typeID := data.Type()
 	if typeID == 0 {
 		return data, nil
 	}
```
mdumandag, line 96: Is this correct? If the `typeId == 0` (I would also prefer using some constant there instead of magic number of 0), then the data contains no payload, hence represents `nil`. But, we are not returning `nil` here but return the data instead?

internal/serialization/serialization.go
```diff
@@ -83,16 +86,20 @@ func (s *Service) ToObject(data *Data) (r interface{}, err error) {
...
-	serializer, ok := s.registry[typeID]
-	if !ok {
-		return nil, hzerrors.NewHazelcastSerializationError(fmt.Sprintf("there is no suitable de-serializer for type %d", typeID), nil)
+	serializer := s.lookupBuiltinDeserializer(typeID)
```
mdumandag, line 97: What is the performance impact of this change? We now have to do a bunch of comparisons before reading/writing (especially portable and identified classes), do we have some before and after numbers?

internal/serialization/serialization.go
```diff
@@ -109,18 +116,15 @@ func (s *Service) WriteObject(output pubserialization.DataOutput, object interfa
 
 func (s *Service) ReadObject(input pubserialization.DataInput) interface{} {
 	serializerID := input.ReadInt32()
-	serializer := s.registry[serializerID]
+	serializer := s.lookupBuiltinDeserializer(serializerID)
+	if serializer == nil {
+		serializer = s.registry[serializerID]
```
mdumandag, line 121: Can this be `nil`?

internal/serialization/default_serializer.go
```diff
@@ -110,70 +113,80 @@ func (*BoolSerializer) Write(output serialization.DataOutput, i interface{}) {
...
 	output.WriteUInt16(i.(uint16))
 }
 
-type Integer16Serializer struct{}
+type IntSerializer struct{}
 
-func (*Integer16Serializer) ID() int32 {
-	return ConstantTypeInteger16
+func (*IntSerializer) ID() int32 {
+	return TypeInt64
```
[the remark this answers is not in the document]
yuce: Unfortunately there's no way around it. Go's `int` type is an alias to `int32` on 32-bit platforms but it's `int64` on 64-bit platforms. Since the platform used for serialization and deserialization may be different, and we don't have an integer type which can be either 32 or 64 bits in the serialization system, we have to choose between `int32` and `int64`. If we choose the former, we should panic if the user tries to set a value > `MaxInt32`. Using `int64` seems like the easiest and safest way.

internal/serialization/default_serializer.go
```diff
@@ -348,10 +361,114 @@ func (*StringArraySerializer) Write(output serialization.DataOutput, i interface
...
+
+func (*UUIDSerializer) Write(output serialization.DataOutput, i interface{}) {
+	uuid := i.(types.UUID)
+	output.WriteInt64(int64(uuid.LeastSignificantBits()))
+	output.WriteInt64(int64(uuid.MostSignificantBits()))
+}
+
+type JavaDateSerializer struct{}
+
+func (JavaDateSerializer) ID() int32 {
```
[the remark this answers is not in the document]
yuce: Pointer receivers are required only to mutate the corresponding value. It is not needed in this case but also does not matter whether to use pointer non-pointer receivers. But I've changed all receivers to be non-pointers in order to be consistent: https://github.com/hazelcast/hazelcast-go-client/pull/581/commits/173bfd382d8de457c1bf1f3cd13b7dfb9d9dcb45

internal/serialization/serialization.go
yuce: That doesn't seem correct to me, removed that check: https://github.com/hazelcast/hazelcast-go-client/pull/581/commits/173bfd382d8de457c1bf1f3cd13b7dfb9d9dcb45

internal/serialization/serialization.go
yuce: There's no corresponding `lookupBuiltinDeserializer` for the `master` branch, so I've added a benchmark for `LookupDefaultSerializer` at: https://github.com/hazelcast/hazelcast-go-client/pull/581/commits/60fb263bd9fec8709430e76a7e26d3cde84db031

You can run it using: `go test -bench BenchmarkService_LookUpDefaultSerializer ./internal/serialization`

According to what I measured, serializer lookup in this PR is faster than the one before:
```
yuce@unreal|13:45:58|521$ go test -bench BenchmarkService_LookUpDefaultSerializer ./internal/serialization
goos: linux
goarch: amd64
pkg: github.com/hazelcast/hazelcast-go-client/internal/serialization
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkService_LookUpDefaultSerializer-12     25852930                43.15 ns/op
PASS
ok      github.com/hazelcast/hazelcast-go-client/internal/serialization 1.185s
/mnt/d/Work/hazelcast/hazelcast-go-client.git (master)
yuce@unreal|13:46:02|521$ go test -bench BenchmarkService_LookUpDefaultSerializer ./internal/serialization
goos: linux
goarch: amd64
pkg: github.com/hazelcast/hazelcast-go-client/internal/serialization
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkService_LookUpDefaultSerializer-12     25516767                48.56 ns/op
PASS
ok      github.com/hazelcast/hazelcast-go-client/internal/serialization 1.310s
/mnt/d/Work/hazelcast/hazelcast-go-client.git (master)
yuce@unreal|13:46:07|521$ go test -bench BenchmarkService_LookUpDefaultSerializer ./internal/serialization
goos: linux
goarch: amd64
pkg: github.com/hazelcast/hazelcast-go-client/internal/serialization
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkService_LookUpDefaultSerializer-12     26558836                43.12 ns/op
PASS
ok      github.com/hazelcast/hazelcast-go-client/internal/serialization 1.215s
/mnt/d/Work/hazelcast/hazelcast-go-client.git (master)
yuce@unreal|13:46:09|521$ go test -bench BenchmarkService_LookUpDefaultSerializer ./internal/serialization
goos: linux
goarch: amd64
pkg: github.com/hazelcast/hazelcast-go-client/internal/serialization
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkService_LookUpDefaultSerializer-12     24733088                43.92 ns/op
PASS
ok      github.com/hazelcast/hazelcast-go-client/internal/serialization 1.160s
/mnt/d/Work/hazelcast/hazelcast-go-client.git (serialization-improvements)
yuce@unreal|13:46:23|521$ go test -bench BenchmarkService_LookUpDefaultSerializer ./internal/serialization
goos: linux
goarch: amd64
pkg: github.com/hazelcast/hazelcast-go-client/internal/serialization
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkService_LookUpDefaultSerializer-12     65511451                17.85 ns/op
PASS
ok      github.com/hazelcast/hazelcast-go-client/internal/serialization 2.180s
/mnt/d/Work/hazelcast/hazelcast-go-client.git (serialization-improvements)
yuce@unreal|13:46:27|521$ go test -bench BenchmarkService_LookUpDefaultSerializer ./internal/serialization
goos: linux
goarch: amd64
pkg: github.com/hazelcast/hazelcast-go-client/internal/serialization
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkService_LookUpDefaultSerializer-12     56741076                18.18 ns/op
PASS
ok      github.com/hazelcast/hazelcast-go-client/internal/serialization 1.079s
```

internal/serialization/serialization.go
yuce: Yes, it can. A panic is thrown but is caught and converted to an error before it reaches user code. I've updated that to explicitly panic with a more descriptive message at: https://github.com/hazelcast/hazelcast-go-client/pull/581/commits/709d98b5690a4f8fbe20e86cec31017edf566068

internal/serialization/serialization.go
```diff
@@ -74,25 +77,24 @@ func (s *Service) ToData(object interface{}) (r *Data, err error) {
...
-// is returned.
-// If this is called with nil, nil is returned.
+// nil is returned if called with nil.
 func (s *Service) ToObject(data *Data) (r interface{}, err error) {
 	defer func() {
 		if rec := recover(); rec != nil {
 			err = makeError(rec)
 		}
 	}()
+	var ok bool
```
mdumandag, line 87: Is there a reason we define this at the top, even if we might not need it?

internal/serialization/serialization.go
```diff
@@ -108,19 +110,19 @@ func (s *Service) WriteObject(output pubserialization.DataOutput, object interfa
...
 
 func (s *Service) ReadObject(input pubserialization.DataInput) interface{} {
-	serializerID := input.ReadInt32()
-	serializer := s.registry[serializerID]
+	typeID := input.ReadInt32()
+	serializer := s.lookupBuiltinDeserializer(typeID)
+	if serializer == nil {
+		serializer = s.registry[typeID]
+	}
+	if serializer == nil {
```
mdumandag, line 118: This is a micro-optimization but would it be ok to move this if check into the if check above so that we don't perform two null checks for the happy scenario, where we find built-in serializer?

internal/serialization/serialization.go
```diff
@@ -136,42 +138,76 @@ func (s *Service) FindSerializerFor(obj interface{}) (pubserialization.Serialize
...
-		s.nameToID[ser.l] = ser.i
+func (s *Service) LookUpDefaultSerializer(obj interface{}) pubserialization.Serializer {
+	serializer := s.lookupBuiltinSerializer(obj)
+	if serializer != nil {
+		return serializer
 	}
-	err := s.registerIdentifiedFactories()
-	if err != nil {
+	if _, ok := obj.(pubserialization.IdentifiedDataSerializable); ok {
+		return s.registry[s.nameToID["identified"]]
```
mdumandag, line 147: Instead of lookup, can we store the identified serializer in a field and return it? The same applies to portable

internal/serialization/serialization.go
```diff
@@ -182,14 +218,19 @@ func (s *Service) registerDefaultSerializers() error {
 	if err = s.registerSerializer(portableSerializer); err != nil {
 		return err
 	}
-	s.nameToID["!portable"] = ConstantTypePortable
+	s.nameToID["!portable"] = TypePortable
```
mdumandag, line 221: I think we can get rid of the `nameToID` altogether by storing the few serializers we put it in the service object

internal/serialization/serialization.go
```diff
@@ -201,16 +242,19 @@ func (s *Service) registerSerializer(serializer pubserialization.Serializer) err
...
+		}
 	}
 }
 
 func (s *Service) registerGlobalSerializer(globalSerializer pubserialization.Serializer) {
 	if globalSerializer != nil {
-		s.registerSerializer(globalSerializer)
+		if err := s.registerSerializer(globalSerializer); err != nil {
+			panic(err)
+		}
```
mdumandag, line 257: Not related to this, but I wasn't able to put comments there. We don't need `getIDByObject` anymore

internal/serialization/serialization.go
```diff
@@ -136,42 +138,76 @@ func (s *Service) FindSerializerFor(obj interface{}) (pubserialization.Serialize
...
+		return s.registry[s.nameToID["!portable"]]
+	}
+	return nil
+}
+
+func (s *Service) lookupBuiltinDeserializer(typeID int32) pubserialization.Serializer {
+	switch typeID {
+	case TypeNil:
+		return nilSerializer
+	case TypeBool:
```
mdumandag, line 159: Can't we put portable and identified serializer(maybe even gob and json) ids here, instead of putting them to s.registry and doing lookup there? 
Then, I believe, the only things we put to registry would be the custom serializer and we can rename it as such

internal/serialization/serialization.go
```diff
@@ -273,7 +301,63 @@ func (s *Service) registerIdentifiedFactories() error {
...
+		return int64ArraySerializer
+	case []float32:
+		return float32ArraySerializer
+	case []float64:
+		return float64ArraySerializer
+	case types.UUID:
+		return uuidSerializer
+	case time.Time:
+		return javaDateSerializer
+	case *big.Int:
```
mdumandag, line 356: Why this is `*big.Int` but we have `time.Time` above?

internal/serialization/serialization.go
yuce: Yes, panics can only be caught by deferred functions.

mdumandag: Oh, I meant the `var ok bool` definition

internal/serialization/serialization.go
yuce: `time.Now` etc. time constructors return `time.Time` but `big.NewInt` constructor returns `*big.Int`.

internal/serialization/default_portable_writer.go
```diff
@@ -189,7 +189,7 @@ func (pw *DefaultPortableWriter) setPosition(fieldName string, fieldType int32)
 	pos := pw.output.Position()
 	pw.output.PWriteInt32(pw.offset+field.Index*Int32SizeInBytes, pos)
 	pw.output.WriteInt16(int16(len(fieldName)))
-	pw.output.WriteBytes(fieldName)
+	pw.output.WriteStringBytes(fieldName)
```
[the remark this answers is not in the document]
yuce: I've just changed the name of the function here, the string is iterated as if it is a byte array.

internal/serialization/serialization.go
yuce: The compiler would probably optimize that.

internal/serialization/serialization.go
yuce: OK, I've refactored that part of the code at: https://github.com/hazelcast/hazelcast-go-client/pull/581/commits/646f1893fc76711d3ffecc2c6d09d0253044d316

internal/serialization/serialization.go
yuce: It seems we can put identified serializable in a global. I've also put portable in a field as you suggested: https://github.com/hazelcast/hazelcast-go-client/pull/581/commits/646f1893fc76711d3ffecc2c6d09d0253044d316

internal/serialization/serialization.go
yuce: Removed it at: https://github.com/hazelcast/hazelcast-go-client/pull/581/commits/646f1893fc76711d3ffecc2c6d09d0253044d316

internal/serialization/serialization.go
yuce: Removed at: https://github.com/hazelcast/hazelcast-go-client/pull/581/commits/646f1893fc76711d3ffecc2c6d09d0253044d316

internal/serialization/serialization.go
yuce: Yes, removed portable and identified serializers there: https://github.com/hazelcast/hazelcast-go-client/pull/581/commits/646f1893fc76711d3ffecc2c6d09d0253044d316

internal/serialization/default_portable_writer.go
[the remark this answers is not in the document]
mdumandag: I tried to run a code like this before writing this comment and before and after produced different byte arrays for out-of-extended-ASCII range strings

https://play.golang.org/p/VCSEn2Rq_VO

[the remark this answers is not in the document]
yuce: I've added a test that checks for non-ASCII portable field names, It seems the prior code was wrong as well, since it wrote the length different than expected: https://github.com/hazelcast/hazelcast-go-client/pull/581/commits/fa87563ed814cbacaeec43b083ffc9427545e32b and https://github.com/hazelcast/hazelcast-go-client/pull/581/commits/41fae0df5b507d28b103ecd24bf2002ea023dde1

mdumandag (approved):
Thanks a lot for the changes. LGTM

yuce:
@mdumandag Thanks a lot for the review!

PR #583, opened 2021-06-21
Soak Tester [API-539]

yuce:
- Added the soak tester.
- Updated some config to be more JSON compatible.

Running the soak tester:
- `cd internal/soak_tests`
- `bash ./run-soak-test.sh`

### DIFF
--- a/cluster/ssl_config.go
+++ b/cluster/ssl_config.go
@@ -32,7 +32,7 @@ import (
 // of tls config as they wish. SSL config also has some helpers such as SetCaPath, AddClientCertAndKeyPath to
 // make configuration easier for users.
 type SSLConfig struct {
-	TLSConfig *tls.Config
+	TLSConfig *tls.Config `json:"-"`
 	Enabled   bool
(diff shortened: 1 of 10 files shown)


alisengul53:
It would be nice to run the soak with remote cluster. Now, as i understand we can only start server on the same mahine. For example, we can create a remote cluster, we can set the server members' addresses as a parameter and the soak will run on the cluster we set the ips'. As default, it is good to keep it as it is.
Totally could be another PR or we can handle it on this PR.

yuce:
Actually specifying a JSON configuration file is supported, which includes member addresses. I've added one in https://github.com/hazelcast/hazelcast-go-client/pull/583/commits/1e54b82ba171e2c27a22fefac31b785723524b83 . Configuring the client using JSON works but needs more improvements to make it convenient to use.

You can use the sample config by passing it it to `./main` like:
```
./main -c ./map/config.json
```

alisengul53 (approved):

PR #585, opened 2021-06-28
Configuration Improvements [API-569]

yuce:
* Simplified configuration names
* Default `hazelcast.Config` contains the defaults and can be directly fed to `StartNewClient`.
* Better JSON compatibility.


cluster/cluster_config.go
```diff
@@ -21,89 +21,83 @@ import (
...
+	} else {
+		for _, addr := range c.Address {
+			if err := checkAddress(addr); err != nil {
+				return fmt.Errorf("invalid address %s: %w", addr, err)
+			}
 		}
 	}
-	if c.ConnectionTimeout < 0 {
-		return ErrConfigInvalidConnectionTimeout
+	if c.ConnectionTimeout <= 0 {
```
mdumandag, line 79: Just curious, is the zero value of connection timeout a special value in Go?

cluster/security_config.go
```diff
@@ -1,12 +1,8 @@
 package cluster
 
 type SecurityConfig struct {
-	Username string
-	Password string
-}
-
-func NewSecurityConfig() SecurityConfig {
-	return SecurityConfig{}
+	Username string `json:",omitempty"`
```
mdumandag, line 4: How will this `SecurityConfig` look like when we add more security options in the future? (like `KerberosIdentityConfig` from Java)

cluster/ssl_config.go
```diff
@@ -28,38 +28,39 @@ import (
...
-// make configuration easier for users.
+// SSLConfig has tls.Config embedded in it so that users can set any field of tls config as they wish.
 type SSLConfig struct {
-	TLSConfig *tls.Config `json:"-"`
-	Enabled   bool
-}
-
-func NewSSLConfig() SSLConfig {
-	return SSLConfig{TLSConfig: &tls.Config{}}
+	tlsConfig *tls.Config
```
mdumandag, line 33: Will the users are able to set TLS config via JSON somehow?

cluster/cluster_config.go
```diff
@@ -21,89 +21,83 @@ import (
...
-		ConnectionTimeout:    5 * time.Second,
-		HeartbeatInterval:    5 * time.Second,
-		HeartbeatTimeout:     60 * time.Second,
-		InvocationTimeout:    120 * time.Second,
-		SecurityConfig:       NewSecurityConfig(),
-		SSLConfig:            NewSSLConfig(),
-		HazelcastCloudConfig: NewHazelcastCloudConfig(),
-		DiscoveryConfig:      NewDiscoveryConfig(),
-	}
+type Config struct {
```
mdumandag, line 32: Why this is called `Config` while others are called `XxxConfig`

config.go
```diff
@@ -82,61 +69,68 @@ func (c *Config) SetLabels(labels ...string) {
...
+	if c.membershipListeners == nil {
+		c.membershipListeners = map[types.UUID]cluster.MembershipStateChangeHandler{}
+	}
+}
+
 // StatsConfig contains configuration for Management Center.
 type StatsConfig struct {
 	// Enabled enables collecting statistics.
-	Enabled bool
+	Enabled bool `json:",omitempty"`
```
mdumandag, line 122: I am not 100% sure if we should do the same here but in the Java client, the statistics are enabled by default and have the default period of 5 seconds

cluster/hazelcast_cloud_config.go
```diff
@@ -17,12 +17,8 @@
 package cluster
 
 type HazelcastCloudConfig struct {
```
mdumandag, line 19: Would it make sense to change it to just `CloudConfig` to match with other clients?

I don't have strong opinions about this though, because `Cloud` might be too broad

logger/logger_config.go
```diff
@@ -37,21 +39,48 @@ const (
...
-type Config struct {
-	Level Level
+func (l Level) String() string {
+	return string(l)
 }
 
-func NewConfig() Config {
-	return Config{
-		Level: InfoLevel,
+func (l *Level) UnmarshalText(b []byte) error {
```
mdumandag, line 46: Are these (`UnmarshalText` / `MarshalText`) necessary to implement JSON serialization/deserialization?

serialization/serialization_config.go
```diff
@@ -23,75 +23,122 @@ import (
...
-	// PortableFactories is a map of factory IDs and corresponding Portable factories.
-	PortableFactories []PortableFactory `json:"-"`
-	// ClassDefinitions contains ClassDefinitions for portable structs.
-	ClassDefinitions []*ClassDefinition `json:"-"`
-	PortableVersion  int32
+	globalSerializer                    Serializer
+	customSerializers                   map[reflect.Type]Serializer
+	identifiedDataSerializableFactories []IdentifiedDataSerializableFactory
+	portableFactories                   []PortableFactory
+	classDefinitions                    []*ClassDefinition
```
mdumandag, line 30: We won't be able to specify these in JSON, right?

serialization/serialization_config.go
```diff
@@ -23,75 +23,122 @@ import (
...
+	b.ensureCustomSerializers()
+	sers := map[reflect.Type]Serializer{}
+	for k, v := range b.customSerializers {
+		sers[k] = v
+	}
+	return sers
+}
+
+// SetClassDefinitions adds zore or more class definitions for portable factories.
+func (b *Config) SetClassDefinitions(definitions ...*ClassDefinition) {
```
mdumandag, line 116: What is the reasoning behind taking references here (`*ClassDefinition`) but struct object above (`factories ...PortableFactory`)

types/duration.go
```diff
@@ -0,0 +1,24 @@
+package types
+
+import (
+	"time"
+)
+
+type Duration time.Duration
```
mdumandag, line 7: I guess we did this just to make it usable in JSON, right? Will this have an extra code required when used with programmatic approach (building the config in go)?

cluster/cluster_config.go
```diff
@@ -21,89 +21,83 @@ import (
...
+		SSL:               c.SSL.Clone(),
+		HazelcastCloud:    c.HazelcastCloud.Clone(),
+		Discovery:         c.Discovery.Clone(),
 	}
 }
 
 func (c *Config) Validate() error {
 	if c.Name == "" {
-		return ErrConfigInvalidClusterName
+		c.Name = defaultName
```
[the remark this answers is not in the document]
yuce: That makes it possible to use `hazelcast.Config{}` without initializing first, which is also very helpful for loading the config from JSON. Do you have suggestion for a new name for `Valudate`?

cluster/cluster_config.go
yuce: Zero value of integers is `0`.

cluster/security_config.go
yuce: I didn't think thoroughly about that but we would probably put it under `SecurityConfig`. I guess putting `Username` and `Password` in something like `SecurityConfig.CredentialsConfig` would make more sense in that case 🤔 🤔 🤔

cluster/hazelcast_cloud_config.go
yuce: Yeah, I wanted to keep the HazelcastCloud brand there but we can change that to match other clients. The user would access it using `config.Cluster.Cloud` in that case.

logger/logger_config.go
yuce: Those are not required for JSON serialization. I've put them to support YAML serialization and other serializers which don't realze `logger.Level` is in fact a `string` but that doesn't work at least for the YAML serializer I've used. So removed those functions at: https://github.com/hazelcast/hazelcast-go-client/pull/585/commits/cf7f17de479a76cb0795726025e896681cb0b265

serialization/serialization_config.go
yuce: No, it's not possible to specify arbitrary objects in JSON. I've unexported those fields because of that.

types/duration.go
yuce: Right. Otherwise the user has to specify durations in nanoseconds instead of easy to use strings like `"10s"`. When using the programmatic way, the users has to cast `time.Duration` to `types.Duration`, like: `types.Duration(10 * time.Second)`.

serialization/serialization_config.go
yuce: `PortableFactory` is an interface, but `ClassDefinition` is a struct.

cluster/cluster_config.go
```diff
@@ -21,89 +21,83 @@ import (
...
+	Name              string               `json:",omitempty"`
+	HazelcastCloud    HazelcastCloudConfig `json:",omitempty"`
+	Address           []string             `json:",omitempty"`
+	InvocationTimeout types.Duration       `json:",omitempty"`
+	HeartbeatInterval types.Duration       `json:",omitempty"`
+	HeartbeatTimeout  types.Duration       `json:",omitempty"`
+	ConnectionTimeout types.Duration       `json:",omitempty"`
+	Discovery         DiscoveryConfig      `json:",omitempty"`
+	RedoOperation     bool                 `json:",omitempty"`
+	Unisocket         bool                 `json:",omitempty"`
```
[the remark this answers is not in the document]
yuce: That's right.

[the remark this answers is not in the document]
yuce: This is mostly to make zero value of `hazelcast.Config` the default configuration, but it definitely helps with the JSON stuff too.

cluster/cluster_config.go
```diff
@@ -21,89 +21,83 @@ import (
...
-		SSLConfig:            NewSSLConfig(),
-		HazelcastCloudConfig: NewHazelcastCloudConfig(),
-		DiscoveryConfig:      NewDiscoveryConfig(),
-	}
+type Config struct {
+	Security          SecurityConfig       `json:",omitempty"`
+	SSL               SSLConfig            `json:",omitempty"`
+	Name              string               `json:",omitempty"`
+	HazelcastCloud    HazelcastCloudConfig `json:",omitempty"`
+	Address           []string             `json:",omitempty"`
```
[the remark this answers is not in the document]
yuce: The plural `Addresses` looked a bit strange to me, but I'll be committing some changes that changes the layout of the config a bit. I can rename `Address` to `Addresses` in that commit.

cluster/ssl_config.go
yuce: There are a few fields in `tls.Config` that  could be set via JSON, but most of it cannot. Ideally setting `config.Cluster.SSL.Enabled = true` without customizing `tlsConfig` should be enough for most users.

cluster/cluster_config.go
yuce: It's in the `cluster` package, so `cluster.Config`.

config.go
yuce: It seems stats is off by default in the Python client 🤔

yuce: Honestly, I put `Enabled` setting just to be compatible with other clients. I think it would be better to assume it is on if Period > 0

cluster/cluster_config.go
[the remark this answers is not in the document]
yuce: I still don't like the plural but changed it to be more consistent with other config options: https://github.com/hazelcast/hazelcast-go-client/pull/585/commits/e82fa7ae67b8f1d8acd80f636ed61faa8a5dad48

cluster/cluster_config.go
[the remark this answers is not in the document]
mdumandag: I couldn't come up with a better name, so I am OK with leaving the name as it is

cluster/ssl_config.go
mdumandag: I see. I find it valuable to configure at least client and CA certificates via JSON, as I assume it will be used this way when the users want to use self-signed certificates but I don't know what we can do about it. So, I am resolving this.

cluster/ssl_config.go
```diff
@@ -28,38 +28,39 @@ import (
 )
 
 // SSLConfig is SSL configuration for client.
-// SSLConfig has tls.Config embedded in it so that users can set any field
-// of tls config as they wish. SSL config also has some helpers such as SetCaPath, AddClientCertAndKeyPath to
-// make configuration easier for users.
+// SSLConfig has tls.Config embedded in it so that users can set any field of tls config as they wish.
 type SSLConfig struct {
```
mdumandag, line 32: Should we change the name of this to TLSConfig as I see from the standard library of Golang, they don't even implement the now deprecated SSLv3, etc.

config.go
mdumandag: I am OK with that

cluster/ssl_config.go
yuce: I don't think there are many serious applications that use SSL, but it became something like an umbrella term which covers TLS too. It's also `SSLConfig` in the Java client, so keeping it that way seems better to me.

cluster/cluster_config.go
```diff
@@ -17,112 +17,71 @@
...
-		ConnectionTimeout:    5 * time.Second,
-		HeartbeatInterval:    5 * time.Second,
-		HeartbeatTimeout:     60 * time.Second,
-		InvocationTimeout:    120 * time.Second,
-		SecurityConfig:       NewSecurityConfig(),
-		SSLConfig:            NewSSLConfig(),
-		HazelcastCloudConfig: NewHazelcastCloudConfig(),
-		DiscoveryConfig:      NewDiscoveryConfig(),
-	}
+type Config struct {
```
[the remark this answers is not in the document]
yuce: I've tried to split the Go client into packages in a way that all code in those packages are related while also trying to keep cyclic dependency problems away. So, `cluster` package contains code that relates to the cluster, connections and connection bound invocations that are carried in those connections. I've tried to organize the configuration similarly, so `config.Cluster` (corresponds to `cluster.Config{}`) is configuration related to the cluster, etc. That allows me to pass an instance of `cluster.Config` to connection manager, cluster service, etc. `InvocationTimeout` is currently under `Cluster`, since that's where it is being used. I view the `Network` configuration which is something to be used by connections. So, IMO `HeartbeatXXX` and `Discovery` don't fit there. They seem more related to the cluster, e.g. `config.Cluster.Discovery.UsePublicIP` enables using public IPs of members in the cluster, etc. Similarly, `config.Cluster.HazelcastCloud` has options to enable retrieving members in a Hz Cloud cluster.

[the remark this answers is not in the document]
mdumandag: OK, fair enough

cluster/security_config.go
yuce: I've moved it to `config.Cluster.Security.Credentials`: https://github.com/hazelcast/hazelcast-go-client/pull/585/commits/f3ad108cdecae243f1381d19af33c0723e4d4a96

cluster/hazelcast_cloud_config.go
yuce: Renamed that to `CloudConfig` (`config.cluster.Cloud`) at: https://github.com/hazelcast/hazelcast-go-client/pull/585/commits/f3ad108cdecae243f1381d19af33c0723e4d4a96

mdumandag (approved):
Thanks for the changes, it looks good to me

yuce:
@mdumandag Thanks a lot for the review!

PR #589, opened 2021-07-04
New aggregation serialization improvements [API-496]

yuce:
- Implements aggregates.
- Adds a few Identified data serializable tests
- Fixes flaky replicated map tests


aggregate/aggregate_distinct.go
```diff
@@ -0,0 +1,56 @@
...
+func (a aggDistinct) WriteData(output serialization.DataOutput) {
+	output.WriteString(a.attrPath)
+	// member side field, not used in client
+	output.WriteInt64(0)
+}
+
+func (a *aggDistinct) ReadData(input serialization.DataInput) {
+	a.attrPath = input.ReadString()
+	// member side field, not used in client
+	input.ReadInt64()
```
mdumandag, line 51: Minor: Not just for this, but I think for all aggregators, ReadData is not called at all and can be made no-op. It is okay to leave these like this, though.

internal/proxy/canonicalizing_set.go
```diff
@@ -0,0 +1,64 @@
...
+func (a AggCanonicalizingSet) FactoryID() int32 {
+	return internal.AggregateFactoryID
+}
+
+func (a AggCanonicalizingSet) ClassID() (classID int32) {
+	return canonicalizingSetClassID
+}
+
+func (a AggCanonicalizingSet) WriteData(output serialization.DataOutput) {
+	output.WriteInt32(int32(len(a)))
```
mdumandag, line 51: Minor: And for this one, I think we don't call WriteData at all.

internal/serialization/portable_serializer_test.go
```diff
@@ -37,7 +37,21 @@ func (*portableFactory1) Create(classID int32) serialization.Portable {
 	return nil
 }
 
-func (*portableFactory1) FactoryID() int32 {
+func (portableFactory1) FactoryID() int32 {
+	return 2
+}
+
+type portableFactory2 struct {
```
mdumandag, line 44: Where do we use this?

proxy_map.go
```diff
@@ -997,6 +1024,24 @@ func (m *Map) tryRemove(ctx context.Context, key interface{}, timeout int64) (in
...
+	data := decoder(resp)
+	obj, err := m.convertToObject(data)
+	if err != nil {
+		return nil, err
+	}
+	// if this is a canonicalizing set return a slice of interface[}
+	cs, ok := obj.(*iproxy.AggCanonicalizingSet)
+	if ok {
+		return []interface{}(*cs), nil
+	}
```
mdumandag, line 1041: What happens if we don't do this? I assume the user will receive `AggCanonicalizingSet`. Can't the user use it like `[]interface` already, since it is a type alias for this?

mdumandag (commented):
As far as I can tell, we have the following aggregators missing
- BigIntAverage, BigIntSum
- BigDecimalAverage, BigDecimalSum
- NumberAverage
- MaxBy, MinBy
- FixedPointSum, FloatingPointSum

Is there a reason we don't implement those?

yuce:
> 
> 
> As far as I can tell, we have the following aggregators missing
> 
>     * BigIntAverage, BigIntSum
> 
>     * BigDecimalAverage, BigDecimalSum
> 
>     * NumberAverage
> 
>     * MaxBy, MinBy
> 
>     * FixedPointSum, FloatingPointSum
> 
> 
> Is there a reason we don't implement those?

`FixedPointAggregator` seems to be the same thing with `LongAggregator`, `FloatingPointAggregator` seems to be the same thing with `DoubleAggregor` and `NumberAverage` seem to be same thing as `DoubleAverage` if I am not missing anything.

`BigInteger`, `MaxBy` and `MinBy` wasn't implemented by the NodeJS client, so it seems I've missed them.

Go client has no `BigDecimal` support.

proxy_map.go
yuce: `m.convertToObject` returns a `*AggCanonicalizingSet` which must be dereferenced before returning to the user, since `AggCanonicalizingSet` is an alias to `[]interface{}`, not `*AggCanonicalizingSet`. Casting that to `[]interface{}` is not necessary though.

aggregate/aggregate_count.go
```diff
@@ -0,0 +1,56 @@
...
+package aggregate
+
+import (
+	"fmt"
+
+	"github.com/hazelcast/hazelcast-go-client/internal"
+	"github.com/hazelcast/hazelcast-go-client/serialization"
+)
+
+func Count(attr string) *aggCount {
```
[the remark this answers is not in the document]
yuce: That's OK, the user accesses the aggregators only via the `Aggregator` interface. I was intending to add docs in the documentation PR, but added at: https://github.com/hazelcast/hazelcast-go-client/pull/589/commits/cb4a37e7ad2caf0bb735083b0e47af67c6bfc9ac in any case.

aggregate/aggregate_count.go
```diff
@@ -0,0 +1,56 @@
...
+func (a aggCount) FactoryID() int32 {
+	return internal.AggregateFactoryID
+}
+
+func (a aggCount) ClassID() (classID int32) {
+	return 4
+}
+
+func (a aggCount) WriteData(output serialization.DataOutput) {
+	output.WriteString(a.attrPath)
```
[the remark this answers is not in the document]
yuce: I've added `xxxAll` variants of the aggregators in: https://github.com/hazelcast/hazelcast-go-client/pull/589/commits/cb4a37e7ad2caf0bb735083b0e47af67c6bfc9ac

internal/serialization/portable_serializer_test.go
yuce: That's used in the morphing portable reader test.

yuce:
`MaxBy`, `MinBy` requires a Java `Comparable`, so didn't implement those. It seems our `BigInt` support is broken, removed that altogether.

aggregate/aggregate_min_max.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (a *aggMinBy) ReadData(input serialization.DataInput) {
+	a.attrPath = input.ReadString()
+	// member side, not used in client
+	input.ReadObject()
+}
+
+func (a aggMinBy) String() string {
+	return makeString("MinBy", a.attrPath)
+}
```
mdumandag, line 98: Is this leftover?

proxy_map.go
```diff
@@ -997,6 +1024,24 @@ func (m *Map) tryRemove(ctx context.Context, key interface{}, timeout int64) (in
...
+	resp, err := m.invokeOnRandomTarget(ctx, req, nil)
+	if err != nil {
+		return nil, err
+	}
+	data := decoder(resp)
+	obj, err := m.convertToObject(data)
+	if err != nil {
+		return nil, err
+	}
+	// if this is a canonicalizing set return a slice of interface[}
```
mdumandag, line 1037: We might need to update the comment now

aggregate/aggregate_count.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+// Count returns the number of values which includes the given attribute.
+// Note that this function may not work as expected for Hazelcast versions prior to 5.0.
+func Count(attr string) *aggCount {
+	return &aggCount{attrPath: attr}
+}
+
+// CountAll returns the number of all values.
+func CountAll() *aggCount {
```
mdumandag, line 31: So, `Count("")` is equal to `CountAll()`, right? I would personally be okay with just `Count(attr)` and document that if you pass `""`, we will count all values.

But I am okay with this approach as well. I just wasn't 100% sure about having a different method just for one variant of the input.

mdumandag (commented):
It seems that fixedPointSum, floatingPointSum and numberAvg are useful when you have different types of numbers in your map. For example, for such a code sample, only the methods above work, not longSum etc.

```java
        IMap<String, Object> m = hz.getMap("m10");
        m.put("k1", 100);
        m.put("k2", 50.1);
        m.put("k3", 20.4d);
        long count = m.aggregate(Aggregators.fixedPointSum());
```

From the Javadoc

```
 * In order to operate on a generic Number type use the fixedPointSum(), floatingPointSum() and numberAvg() aggregators.
 * All of them will convert the given number to either Long or Double during the accumulation phase.
 * It will result in a lot of allocations since each number has to be converted, but it enables the user
 * to operate on the whole family of numbers. It is especially useful if the numbers given to the aggregators
 * may not be of one type only.
```

aggregate/aggregate_count.go
yuce: Yeah, `CountAll()` is easier to discover than `Count("")`.

proxy_map.go
yuce: OK: https://github.com/hazelcast/hazelcast-go-client/pull/589/commits/5d27f239bdc4f73739563952479e1dfbc00ccbfe

aggregate/aggregate_min_max.go
yuce: Yep: https://github.com/hazelcast/hazelcast-go-client/pull/589/commits/5d27f239bdc4f73739563952479e1dfbc00ccbfe

yuce:
`DoubleSum` and `DoubleAverage` seem to work in mixed integer and floating point cases, added tests at: https://github.com/hazelcast/hazelcast-go-client/pull/589/commits/5d27f239bdc4f73739563952479e1dfbc00ccbfe

mdumandag (approved):

yuce:
@mdumandag Thanks for the review!

PR #590, opened 2021-07-08
V1 Wrap Up [API-500]

yuce:


client_it_test.go
```diff
@@ -213,3 +214,95 @@ func TestClient_AddDistributedObjectListener(t *testing.T) {
...
+		hz.LifecycleStateClientConnected,
+		hz.LifecycleStateStarted,
+		hz.LifecycleStateClientDisconnected,
+		hz.LifecycleStateClientConnected,
+		hz.LifecycleStateClientDisconnected,
+		hz.LifecycleStateShuttingDown,
+		hz.LifecycleStateShutDown,
+	}
+	fmt.Println("target :", target)
+	fmt.Println("events :", events)
```
mdumandag, line 253: Leftover or printing to stdout during tests is a common thing in our test suite? There is a similar one below

client_it_test.go
```diff
@@ -213,3 +214,95 @@ func TestClient_AddDistributedObjectListener(t *testing.T) {
...
+	time.Sleep(5 * time.Second)
+	cls = it.StartNewClusterWithOptions("go-cli-test-cluster", 15701, 3)
+	time.Sleep(5 * time.Second)
+	cls.Shutdown()
+	c.Shutdown(ctx)
+	mu.Lock()
+	defer mu.Unlock()
+	target := []hz.LifecycleState{
+		hz.LifecycleStateStarting,
+		hz.LifecycleStateClientConnected,
```
mdumandag, line 244: Should we drop the name `Client` from the event names?

I mean `LifecycleStateClientConnected` -> `LifecycleStateConnected`.

We use Client in Java side because member and client use the same enum, but I don't see a reason to do so in our non-Java clients

cluster/connection_strategy_config.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"time"
+
+	"github.com/hazelcast/hazelcast-go-client/types"
+)
+
+type ConnectionStrategyConfig struct {
+	Retry            ConnectionRetryConfig
+	Timeout          types.Duration `json:",omitempty"`
+	DisableReconnect bool           `json:",omitempty"`
```
mdumandag, line 28: Instead of a boolean, should we make this a separate type(ReconnectMode) with values 0 (OFF) and 1 (ON). This way, in the future, we could easily add the 2 (ASYNC) reconnect mode.

cluster/connection_strategy_config.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (c *ConnectionRetryConfig) Validate() error {
+	if c.InitialBackoff <= 0 {
+		c.InitialBackoff = types.Duration(1 * time.Second)
+	}
+	if c.MaxBackoff <= 0 {
+		c.MaxBackoff = types.Duration(30 * time.Second)
+	}
+	if c.Multiplier <= 0 {
```
mdumandag, line 61: Just a thought that comes to my mind after yesterday's review :smile: 

Should we change them such that if they are equal to zero value, set them to default value, if they are negative (or somehow out-of-valid-range), throw?

For example, for multiplier, values between 0 and 1.0 are not valid at all, as it would shorten the duration between backoffs, instead of increasing it. Should we throw on such cases?

This is not just related to this, but to whole configuration

internal/cluster/connection_strategy.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/hazelcast/hazelcast-go-client/types"
+)
+
+func makeRetryPolicy(r *rand.Rand, config *cluster.ConnectionRetryConfig) cb.RetryPolicyFunc {
+	backoff := int64(minDuration(config.MaxBackoff, config.InitialBackoff))
+	jitter := config.Jitter
+	multiplier := config.Multiplier
+	return func(attempt int) time.Duration {
+		d1 := float64(backoff) + float64(backoff)*jitter*(2.0*r.Float64()-1.0)
+		backoff = int64(float64(backoff) * multiplier)
```
mdumandag, line 34: should we also take a minimum of this with max backoff?

cluster/connection_strategy_config.go
yuce: APIs are sync-first in Go, so we probably won't have async mode for the Go client. The user can easily make async calls from sync ones.

internal/cluster/connection_strategy.go
yuce: Ooops! updated at: https://github.com/hazelcast/hazelcast-go-client/pull/590/commits/54049484f1904f787a5d094c107b25f6f1f76de4

client_it_test.go
```diff
@@ -213,3 +214,95 @@ func TestClient_AddDistributedObjectListener(t *testing.T) {
...
+	}
+	time.Sleep(5 * time.Second)
+	cls.Shutdown()
+	time.Sleep(5 * time.Second)
+	cls = it.StartNewClusterWithOptions("go-cli-test-cluster", 15701, 3)
+	time.Sleep(5 * time.Second)
+	cls.Shutdown()
+	c.Shutdown(ctx)
+	mu.Lock()
+	defer mu.Unlock()
```
srknzl, line 241: Do we need these two lines? Stopping the user event dispatcher prevents events' processing, does not it? I deleted these lines and test still passed?

client_it_test.go
```diff
@@ -213,3 +214,95 @@ func TestClient_AddDistributedObjectListener(t *testing.T) {
...
+	// start shutting down members one by one
+	for _, uuid := range cls.MemberUUIDs {
+		time.Sleep(1 * time.Second)
+		cls.RC.ShutdownMember(ctx, cls.ClusterID, uuid)
+	}
+	time.Sleep(1 * time.Second)
+	c.Shutdown(ctx)
+
+	mu.Lock()
+	defer mu.Unlock()
```
srknzl, line 294: same comment as above

client_it_test.go
```diff
@@ -213,3 +214,95 @@ func TestClient_AddDistributedObjectListener(t *testing.T) {
 		mu.Unlock()
 	})
 }
+
+func TestClusterReconnection(t *testing.T) {
```
srknzl, line 217: Maybe you can add a map.put test while cluster is restarted. What happens in that case? Is map.put synchronous? will it work after cluster restart?

client_it_test.go
yuce: `assert.Equal` prints the numeric values of lifecycle events, although we have nice string representations for them. I've replaces prints with logs at: https://github.com/hazelcast/hazelcast-go-client/pull/590/commits/b968570d55f526d7006a824378208c9c5fa8cf5b

cluster/connection_strategy_config.go
mdumandag: The async start might be an unfortunate name but basically, it means the client won't wait to connect to cluster in its start method. So, the method might be sync, but the behavior can be implemented any way

client_it_test.go
yuce: Renamed that at: https://github.com/hazelcast/hazelcast-go-client/pull/590/commits/d19d9f74ec76de7388c57eb1cbe0a5bc1e68a057

cluster/connection_strategy_config.go
yuce: OK, here it is: https://github.com/hazelcast/hazelcast-go-client/pull/590/commits/d19d9f74ec76de7388c57eb1cbe0a5bc1e68a057

client_it_test.go
yuce: Try with `go test -run TestClusterReconnection -race .` ;)

client_it_test.go
yuce: same comment as above

mdumandag (approved):

client_it_test.go
srknzl: :D

cluster/connection_strategy_config.go
mdumandag: Sorry, I mixed this up with the asyncStart boolean in the Java client in my previous comment. ASYNC reconnect mode is still a thing, so it might be good to do it anyways for the future

yuce: Changed it at: https://github.com/hazelcast/hazelcast-go-client/pull/590/commits/1285e0da5b15d82430f385cf509e263f59c25ac5

client_it_test.go
yuce: map.put is synchronous. But it's a bit hard to test it with the current setup since we shut down the cluster in order to be able to trigger client disconnection. So, if that happens right before map.put, a `nil` value is returned instead of the old value since that information is lost when the cluster is down. We need a way to cause an arbitrary disconnection between the client and cluster. I'll think about that after V1.

mdumandag (approved):

PR #597, opened 2021-07-28
Add `encoding/json` package functions to `serialization.JSON`

martijndeheus:
This PR allows the `serialization.JSON` type to be included in the (de)serialization of larger structs using `encoding.json` package. 

It does this by adding `MarshalJSON` and `UnmarshalJSON` functions to prevent the default behaviour for `[]byte` types, since we assume the data is already serialized or has to remain serialized for this type. See [here](https://pkg.go.dev/encoding/json#Marshal) for details.

I ran into this while working on Turbine and writing an HTTP API that includes endpoints that return JSON values received from HZ as part of a larger struct. This change prevents the need to cast to `json.RawMessage` in this case and is (in my opinion) the expected behaviour for this data type.

### DIFF
--- a/serialization/json_value.go
+++ b/serialization/json_value.go
@@ -24,9 +24,9 @@ func (j JSON) String() string {
 	return string(j)
 }
 
-// MarshalJSON is used by functions in the Go `json` package. This function
-// ensures that a `serialization.JSON` value is correctly serialized by this
-// package. This is mostly applicable when `serialization.JSON` is
+// MarshalJSON is used by functions in the Go `encoding/json` package. This
+// function ensures that a `serialization.JSON` value is correctly serialized
+// by this package. This is mostly applicable when `serialization.JSON` is
 // serialized as a field of a larger struct.
 func (j JSON) MarshalJSON() ([]byte, error) {
 	if j == nil {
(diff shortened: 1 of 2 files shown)


serialization/json_value.go
```diff
@@ -16,8 +16,32 @@
...
+import "errors"
+
 type JSON []byte
 
 func (j JSON) String() string {
 	return string(j)
 }
+
+// MarshalJSON is used by functions in the Go `encoding/json` package. This
+// function ensures that a `serialization.JSON` value is correctly serialized
```
yuce, line 28: Could you move "This" after the first sentence to the line below? Similary for other comment lines. I think it would be great if each sentence is in its own line (no matter the length of the line).

serialization/json_value.go
```diff
@@ -16,8 +16,32 @@
...
+	}
+	return j, nil
+}
+
+// UnmarshalJSON is used by functions in the Go `encoding/json` package. This
+// function ensures that a `serialization.JSON` value is correctly deserialized
+// from JSON objects that contain a field of this type.
+func (j *JSON) UnmarshalJSON(data []byte) error {
+	if j == nil {
+		return errors.New("serialization.JSON: UnmarshalJSON on nil pointer")
```
yuce, line 43: Could you change this line to return a serialization error, like:
```
return ihzerrors.NewSerializationError("serialization.JSON: UnmarshalJSON on nil pointer", nil)
```
The corresponding import statement would be:
```
import ihzerrors "github.com/hazelcast/hazelcast-go-client/internal/hzerrors"
```

yuce (changes requested):
Thanks for your PR! Just two trivial comments.

serialization/json_value.go
yuce: ^ It's even better to return an illegal argument error.

martijndeheus: Fixed in ab9de482c6c8c3a68a0c468890c7022da27ea8f6

serialization/json_value.go
martijndeheus: Addressed in ab9de482c6c8c3a68a0c468890c7022da27ea8f6

yuce (approved):
Thanks, looks good!

PR #601, opened 2021-08-03
Add it.SkipIf to skip tests based on conditions

martijndeheus:
Introduce `SkipIf` function to skip test cases based on conditions like versioning or the operating system.

The code to parse the conditions is not so clean. This could be more clean when using option pattern, however this comes at the cost of readability where the function is actually applied. 

I tried to test most cases to confirm the parsing is valid.

### DIFF
--- a/aggregate/aggregate_it_test.go
+++ b/aggregate/aggregate_it_test.go
@@ -49,7 +49,7 @@ func TestCountAll(t *testing.T) {
 		config.Serialization.SetPortableFactories(it.SamplePortableFactory{})
 	}
 	it.MapTesterWithConfig(t, cbCallback, func(t *testing.T, m *hz.Map) {
-		it.SkipIf(t, "hz <= 4.0.3")
+		it.SkipIf(t, "hz < 4.1")
 		ctx := context.Background()
 		it.MustValue(m.Put(ctx, "k1", &it.SamplePortable{A: "foo", B: 10}))
 		it.MustValue(m.Put(ctx, "k2", &it.SamplePortable{A: "bar", B: 30}))
@@ -86,7 +86,7 @@ func TestDistinctValuesAll(t *testing.T) {
 		config.Serialization.SetPortableFactories(it.SamplePortableFactory{})
 	}
 	it.MapTesterWithConfig(t, cbCallback, func(t *testing.T, m *hz.Map) {
-		it.SkipIf(t, "hz <= 4.0.3")
+		it.SkipIf(t, "hz < 4.1")
 		ctx := context.Background()
 		it.MustValue(m.Put(ctx, "k1", &it.SamplePortable{A: "foo", B: 10}))
 		it.MustValue(m.Put(ctx, "k2", &it.SamplePortable{A: "bar", B: 30}))
--- a/aggregate/aggregate_it_test.go
+++ b/aggregate/aggregate_it_test.go
@@ -49,6 +49,7 @@ func TestCountAll(t *testing.T) {
 		config.Serialization.SetPortableFactories(it.SamplePortableFactory{})
 	}
 	it.MapTesterWithConfig(t, cbCallback, func(t *testing.T, m *hz.Map) {
+		it.SkipIf(t, "hz <= 4.0.3")
 		ctx := context.Background()
 		it.MustValue(m.Put(ctx, "k1", &it.SamplePortable{A: "foo", B: 10}))
 		it.MustValue(m.Put(ctx, "k2", &it.SamplePortable{A: "bar", B: 30}))
@@ -85,6 +86,7 @@ func TestDistinctValuesAll(t *testing.T) {
 		config.Serialization.SetPortableFactories(it.SamplePortableFactory{})
 	}
 	it.MapTesterWithConfig(t, cbCallback, func(t *testing.T, m *hz.Map) {
+		it.SkipIf(t, "hz <= 4.0.3")
 		ctx := context.Background()
 		it.MustValue(m.Put(ctx, "k1", &it.SamplePortable{A: "foo", B: 10}))
 		it.MustValue(m.Put(ctx, "k2", &it.SamplePortable{A: "bar", B: 30}))
--- a/internal/it/export_test.go
+++ b/internal/it/export_test.go
@@ -1,3 +1,0 @@
-package it
-
-var InternalSkipIf = skipIf
--- a/internal/it/export_test.go
+++ b/internal/it/export_test.go
@@ -1,0 +1,3 @@
+package it
+
+var InternalSkipIf = skipIf
--- a/internal/it/test_skip.go
+++ b/internal/it/test_skip.go
@@ -68,13 +68,13 @@ func checkCondition(condition string) bool {
 		validateLength(parts, 1, condition, "!oss")
 		return enterprise()
 	default:
-		panic(fmt.Errorf("Unexpected test skip constant \"%s\" in %s", parts[0], condition))
+		panic(fmt.Errorf("unexpected test skip constant \"%s\" in %s", parts[0], condition))
 	}
 }
 
 func validateLength(parts []string, expected int, condition, example string) {
 	if len(parts) != expected {
-		panic(fmt.Errorf("Unexpected format for %s, example of expected condition: \"%s\" ", condition, example))
+		panic(fmt.Errorf("unexpected format for %s, example of expected condition: \"%s\" ", condition, example))
 	}
 }
 
@@ -94,7 +94,7 @@ func checkVersion(given, operator, actual string) bool {
 	case "<=":
 		return greater >= 0
 	default:
-		panic(fmt.Errorf("Unexpected test skip operator \"%s\" to compare versions", operator))
+		panic(fmt.Errorf("unexpected test skip operator \"%s\" to compare versions", operator))
 	}
 }
 
@@ -112,11 +112,11 @@ func compareVersions(given, actual string) int {
 	for i := 0; i < min; i++ {
 		givenNumber, err := strconv.Atoi(givenVersions[i])
 		if err != nil {
-			panic(fmt.Errorf("Could not parse version number (to integer): %d", givenNumber))
+			panic(fmt.Errorf("could not parse version number (to integer): %d", givenNumber))
 		}
 		actualNumber, err := strconv.Atoi(actualVersions[i])
 		if err != nil {
-			panic(fmt.Errorf("Could not parse version number (to integer): %d", actualNumber))
+			panic(fmt.Errorf("could not parse version number (to integer): %d", actualNumber))
 		}
 
 		if givenNumber > actualNumber {
@@ -136,7 +136,7 @@ func checkOS(operator, value string) bool {
 	case "!=":
 		return runtime.GOOS != value
 	default:
-		panic(fmt.Errorf("Unexpected test skip operator \"%s\" in \"%s\" condition", operator, skipOS))
+		panic(fmt.Errorf("unexpected test skip operator \"%s\" in \"%s\" condition", operator, skipOS))
 	}
 }
 
--- a/internal/it/test_skip.go
+++ b/internal/it/test_skip.go
@@ -20,6 +20,13 @@ const (
 	enterpriseKey     = "HAZELCAST_ENTERPRISE_KEY"
 )
 
+// SkipIf can be used to skip a test case based on comma-separated conditions.
+// The following conditions may be set:
+// "hz": Hazelcast version
+// "ver": go client version
+// "os": value of runtime.GOOS
+// "enterprise"/"oss": presence of enterprise key environment variable
+// Example: SkipIf(t, "ver > 1.1, hz = 5")
 func SkipIf(t *testing.T, conditionString string) {
 	skip, err := skipIf(conditionString)
 	if err != nil {
--- a/internal/it/test_skip.go
+++ b/internal/it/test_skip.go
@@ -33,7 +33,7 @@ func SkipIf(t *testing.T, conditionString string) {
 		t.Fatalf(err.Error())
 	}
 	if skip {
-		t.Skip()
+		t.Skip("Skipping test: SkipIf conditions met")
 	}
 }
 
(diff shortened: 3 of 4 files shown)


internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+
+func validateLength(parts []string, expected int, condition, example string) error {
+	if len(parts) != expected {
+		return fmt.Errorf("Unexpected format for %s, example of expected condition: \"%s\" ", condition, example)
+	}
+	return nil
+}
+
+func checkVersion(given, operator, actual string) (bool, error) {
+	isGivenGreater, err := compareVersions(given, actual)
```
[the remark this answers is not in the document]
yuce: Also, would using an int instead of a pointer to bool make more sense here (*greater* variable)? Since there are three states of the comparison result.

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		return true, err
+	}
+
+	switch operator {
+	case "=":
+		if isGivenGreater == nil {
+			return true, nil
+		}
+		return false, nil
```
yuce, line 135: Can be simplified to `return greater == 0, nil` (assuming you are using an int for `greater`)

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	case "!=":
+		if isGivenGreater != nil {
+			return true, nil
+		}
+		return false, nil
```
yuce, line 140: Can be simplified to `return greater != 0, nil` (assuming you are using an int for `greater`)

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	case ">":
+		if isGivenGreater != nil && !*isGivenGreater {
+			return true, nil
+		}
+		return false, nil
```
yuce, line 145: Can be simplified to `return greater > 0, nil` (assuming you are using an int for `greater`)

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	case ">=":
+		if isGivenGreater != nil && *isGivenGreater {
+			return false, nil
+		}
+		return true, nil
```
yuce, line 150: Can be simplified to `return greater >= 0, nil` (assuming you are using an int for `greater`)

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	case "<":
+		if isGivenGreater != nil && *isGivenGreater {
+			return true, nil
+		}
+		return false, nil
```
yuce, line 155: Can be simplified to `return greater < 0, nil` (assuming you are using an int for `greater`)

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	case "<=":
+		if isGivenGreater != nil && !*isGivenGreater {
+			return false, nil
+		}
+		return true, nil
```
yuce, line 160: Can be simplified to `return greater <= 0, nil` (assuming you are using an int for `greater`)

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	default:
+		return true, fmt.Errorf("Unexpected test skip operator \"%s\" to compare versions", operator)
```
yuce, line 162: Instead of escaping the quote, you can use backticks: `Unexpected test skip operator %s`

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+}
+
+func compareVersions(given, actual string) (*bool, error) {
```
yuce, line 166: Like I've suggested above, how about making this one return an int?

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+
+		if givenNumber > actualNumber {
+			res := true
+			return &res, nil
+		}
+		if actualNumber > givenNumber {
+			res := false
+			return &res, nil
+		}
```
yuce, line 194: Although it's usually not a good idea, I think you can just do something like here:
```
num := givenNum - actualNum
if num != 0 {
    return num, nil
}
```

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	return nil, nil
+}
+
+func checkOS(operator, value string) (bool, error) {
+	switch operator {
+	case "=":
+		if runtime.GOOS == value {
+			return true, nil
+		}
```
yuce, line 204: Can be simplified to `return runtime.GOOS == value, nil`

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	case "!=":
+		if runtime.GOOS != value {
+			return true, nil
+		}
```
yuce, line 208: Can be simplified to `return runtime.GOOS != value, nil`

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	default:
+		return true, fmt.Errorf("Unexpected test skip operator \"%s\" in \"%s\" condition", operator, skipOS)
```
yuce, line 210: Errors should start with lowercase letters.

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	return false, nil
+}
+
+func checkEnterprise(expected bool) bool {
```
yuce, line 215: Does it make sense to rename this function to `enterprise()` and return whether enterprise key was set? It can be as simple as:
```
return os.GetEnv(enterpriseKey) != ""
```

aggregate/aggregate_it_test.go
```diff
@@ -49,6 +49,7 @@ func TestCountAll(t *testing.T) {
 		config.Serialization.SetPortableFactories(it.SamplePortableFactory{})
 	}
 	it.MapTesterWithConfig(t, cbCallback, func(t *testing.T, m *hz.Map) {
+		it.SkipIf(t, "hz <= 4.0.3")
```
yuce, line 52: Maybe we should change the condition to: `hz < 4.1`.

aggregate/aggregate_it_test.go
```diff
@@ -85,6 +86,7 @@ func TestDistinctValuesAll(t *testing.T) {
...
```
yuce, line 89: Maybe we should change the condition to: `hz < 4.1`.

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	skip, err := skipIf(conditionString)
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
+	if skip {
+		t.Skip("Skipping test: SkipIf conditions met")
+	}
+}
+
+func skipIf(conditionString string) (bool, error) {
```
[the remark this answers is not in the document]
yuce: Also, I think it's OK to panic here if error is not nil, so we can simplify calling it:
```
func canSkip(condition string) bool
```

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+// The following conditions may be set:
+// "hz": Hazelcast version
+// "ver": go client version
+// "os": value of runtime.GOOS
+// "enterprise"/"oss": presence of enterprise key environment variable
+// Example: SkipIf(t, "ver > 1.1, hz = 5")
+func SkipIf(t *testing.T, conditionString string) {
+	skip, err := skipIf(conditionString)
+	if err != nil {
+		t.Fatalf(err.Error())
```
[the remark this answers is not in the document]
yuce: If you change `skipIf`/`canSkip` to panic instead of returning an error, we can simplify the code here to:
```
if skip := canSkip(condition); skip {
    t.Skip("Skipping test: SkipIf conditions met")
}
```

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+			if err := validateLength(parts, 3, condition, "ver = 4.0"); err != nil {
+				return true, err
+			}
+			skip, err := checkVersion(parts[2], parts[1], internal.ClientVersion)
+			if err != nil {
+				return true, err
+			}
+			if skip {
+				return true, nil
+			}
```
yuce, line 66: It seems we repeat the code a bit here. Would it make sense to remove code duplication?

internal/it/test_skip.go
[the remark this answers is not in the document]
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/export_test.go
```diff
@@ -0,0 +1,3 @@
+package it
+
+var InternalSkipIf = skipIf
```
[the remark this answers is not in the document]
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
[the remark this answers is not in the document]
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	if skip {
+		t.Skip("Skipping test: SkipIf conditions met")
+	}
+}
+
+func skipIf(conditionString string) (bool, error) {
+	conditions := strings.Split(conditionString, ",")
+	for _, condition := range conditions {
+		condition = strings.Trim(condition, " ")
+		switch parts := strings.Split(condition, " "); {
```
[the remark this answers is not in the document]
martijndeheus: This would result in the default case where we also return an error. This error should be sufficient to figure out what's wrong.

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+// "os": value of runtime.GOOS
+// "enterprise"/"oss": presence of enterprise key environment variable
+// Example: SkipIf(t, "ver > 1.1, hz = 5")
+func SkipIf(t *testing.T, conditionString string) {
+	skip, err := skipIf(conditionString)
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
+	if skip {
+		t.Skip("Skipping test: SkipIf conditions met")
```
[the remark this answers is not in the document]
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+				return true, nil
+			}
+		case parts[0] == "!"+skipOSS:
+			if err := validateLength(parts, 1, condition, "!oss"); err != nil {
+				return true, err
+			}
+			skip := checkEnterprise(false)
+			if skip {
+				return true, nil
+			}
```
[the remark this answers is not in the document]
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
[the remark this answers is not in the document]
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
```diff
@@ -0,0 +1,221 @@
...
+	if len(parts) != expected {
+		return fmt.Errorf("Unexpected format for %s, example of expected condition: \"%s\" ", condition, example)
+	}
+	return nil
+}
+
+func checkVersion(given, operator, actual string) (bool, error) {
+	isGivenGreater, err := compareVersions(given, actual)
+	if err != nil {
+		return true, err
```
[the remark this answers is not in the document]
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
martijndeheus: I've cleaned up other parts of the code making it more compact. I think we can leave this here as it's a bit cleaner.

internal/it/test_skip.go
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

internal/it/test_skip.go
martijndeheus: Addressed in a3336bc12bdd1566b7a1c23a74fd002c764bb8e1

internal/it/test_skip.go
martijndeheus: I've cleaned up this part a bit. I think it is a lot simpler now.

aggregate/aggregate_it_test.go
martijndeheus: Addressed in e8aff4dd045b95ef5195f09d84a7a73a00af6e1b

aggregate/aggregate_it_test.go
martijndeheus: Addressed in e8aff4dd045b95ef5195f09d84a7a73a00af6e1b

internal/it/test_skip.go
martijndeheus: Addressed in 5071a2fabacf61a0c6750af893641b4dfc45f051

martijndeheus:
Thanks for the thorough review @yuce . I addressed your comments. Can you take another look?

yuce (approved):
Thanks, looks great!

PR #602, opened 2021-08-04
Add GetDistributedObjects API.

enozcan:
Adds GetDistributedObjects API to the client.

I also would like to add unit tests to proxy manager - especially to assert no remote proxy is created when the newly added remote flag is false. But I couldn't find the proper place to put such a test. Do we have a plan to add a test suite together with mocked services?

### DIFF
--- a/client.go
+++ b/client.go
@@ -31,6 +31,7 @@ import (
 	"github.com/hazelcast/hazelcast-go-client/internal/invocation"
 	ilogger "github.com/hazelcast/hazelcast-go-client/internal/logger"
 	"github.com/hazelcast/hazelcast-go-client/internal/proto"
+	"github.com/hazelcast/hazelcast-go-client/internal/proto/codec"
 	iproxy "github.com/hazelcast/hazelcast-go-client/internal/proxy"
 	"github.com/hazelcast/hazelcast-go-client/internal/security"
 	"github.com/hazelcast/hazelcast-go-client/internal/serialization"
@@ -198,6 +199,44 @@ func (c *Client) GetPNCounter(ctx context.Context, name string) (*PNCounter, err
 		return nil, hzerrors.ErrClientNotActive
 	}
 	return c.proxyManager.getPNCounter(ctx, name)
+}
+
+func (c *Client) GetDistributedObjects(ctx context.Context) (map[string]interface{}, error) {
+	if atomic.LoadInt32(&c.state) != ready {
+		return nil, hzerrors.ErrClientNotActive
+	}
+
+	request := codec.EncodeClientGetDistributedObjectsRequest()
+	resp, err := c.proxyManager.invokeOnRandomTarget(ctx, request, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	cachedObjects := make(map[codec.DistributedObjectInfo]struct{})
+	for _, o := range c.proxyManager.getCachedObjectsInfo() {
+		cachedObjects[o] = struct{}{}
+	}
+
+	for _, o := range codec.DecodeClientGetDistributedObjectsResponse(resp) {
+		if !isServiceSupported(o.ServiceName()) {
+			continue
+		}
+		delete(cachedObjects, o)
+		// Ensure the remote proxy resides in the manager's cache. This should
+		// not create a remote proxy as it might be destroyed cluster-wide after
+		// member's response.
+		_, err = c.proxyManager.proxyFor(ctx, o.ServiceName(), o.Name(), false)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	for o := range cachedObjects {
+		// Purge the stale object from the manager's cache.
+		c.proxyManager.remove(o.ServiceName(), o.Name())
+	}
+
+	return c.proxyManager.getCachedObjects()
 }
 
 func (c *Client) start(ctx context.Context) error {
--- a/client_it_test.go
+++ b/client_it_test.go
@@ -321,3 +321,112 @@ func TestClusterReconnection_ReconnectModeOff(t *testing.T) {
 	time.Sleep(100 * time.Millisecond)
 	assert.Equal(t, false, c.Running())
 }
+
+func TestClient_GetDistributedObjects(t *testing.T) {
+	const (
+		testMapName = "get-distributed-objects-map"
+		testSetName = "get-distributed-objects-set"
+	)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 40*time.Second)
+	defer cancel()
+
+	cls := it.StartNewClusterWithOptions("go-cli-test-cluster", 15701, 1)
+	time.Sleep(2 * time.Second)
+	defer cls.Shutdown()
+	config := cls.DefaultConfig()
+
+	c, err := hz.StartNewClientWithConfig(ctx, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = c.GetMap(ctx, testMapName)
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = c.GetSet(ctx, testSetName)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err = c.Shutdown(ctx); err != nil {
+		t.Fatal(err)
+	}
+
+	c, err = hz.StartNewClientWithConfig(ctx, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+	objects, err := c.GetDistributedObjects(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, 2, len(objects))
+	hzMap := objects[testMapName]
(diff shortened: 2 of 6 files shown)


proxy_manager.go
```diff
@@ -45,61 +45,115 @@ func newProxyManager(bundle creationBundle) *proxyManager {
...
-	if p, err := m.proxyFor(ctx, ServiceNamePNCounter, name); err != nil {
+	if p, err := m.proxyFor(ctx, ServiceNamePNCounter, name, true); err != nil {
 		return nil, err
 	} else {
 		return newPNCounter(p), nil
 	}
 }
 
+func (m *proxyManager) invokeOnRandomTarget(ctx context.Context, request *proto.ClientMessage, handler proto.ClientMessageHandler) (*proto.ClientMessage, error) {
+	p, err := m.createProxy(ctx, "", "", false)
```
enozcan, line 104: I couldn't decide if it's a good idea to create an unnamed proxy as a field of the manager just for this purpose - not sure if any other api possibly needs it. Suggestions on this manner are welcome.

client.go
```diff
@@ -200,6 +201,44 @@ func (c *Client) GetPNCounter(ctx context.Context, name string) (*PNCounter, err
...
+			continue
+		}
+		delete(cachedObjects, o)
+		// Ensure the remote proxy resides in the manager's cache. This should
+		// not create a remote proxy as it might be destroyed cluster-wide after
+		// member's response.
+		_, err = c.proxyManager.proxyFor(ctx, o.ServiceName(), o.Name(), false)
+		if err != nil {
+			return nil, err
+		}
```
yuce, line 231: Can be simplified to:
```
if _, err = c.proxyManager....; err != nil {
    return nil, err
}
```

client_it_test.go
```diff
@@ -321,3 +321,112 @@ func TestClusterReconnection_ReconnectModeOff(t *testing.T) {
...
 	assert.Equal(t, false, c.Running())
 }
+
+func TestClient_GetDistributedObjects(t *testing.T) {
+	const (
+		testMapName = "get-distributed-objects-map"
+		testSetName = "get-distributed-objects-set"
+	)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 40*time.Second)
```
yuce, line 331: Is there a reason a  context with 40s timeout is created? How about using `context.Background()`?

client_it_test.go
```diff
@@ -321,3 +321,112 @@ func TestClusterReconnection_ReconnectModeOff(t *testing.T) {
...
+	defer cancel()
+
+	cls := it.StartNewClusterWithOptions("go-cli-test-cluster", 15701, 1)
```
yuce, line 334: Is there a reason a new cluster is created for this test? How about using `it.Tester` or `it.TesterWithConfigBuilder`? They re-use a pre-existing cluster and would simplify this test.

client_it_test.go
```diff
@@ -321,3 +321,112 @@ func TestClusterReconnection_ReconnectModeOff(t *testing.T) {
 	time.Sleep(100 * time.Millisecond)
 	assert.Equal(t, false, c.Running())
 }
+
+func TestClient_GetDistributedObjects(t *testing.T) {
+	const (
+		testMapName = "get-distributed-objects-map"
+		testSetName = "get-distributed-objects-set"
```
yuce, line 328: How about using auto-generated map and set names, like: https://github.com/hazelcast/hazelcast-go-client/blob/7326ba425837f2cefb9c16f4eb2479bc282ec301/internal/it/map.go#L35

client_it_test.go
```diff
@@ -321,3 +321,112 @@ func TestClusterReconnection_ReconnectModeOff(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, 2, len(objects))
+	hzMap := objects[testMapName]
+	hzSet := objects[testSetName]
+	assert.NotNil(t, hzMap)
+	assert.NotNil(t, hzSet)
+	assert.IsType(t, new(hz.Map), hzMap)
```
yuce, line 369: Doesn't `assert.IsType(t, *hz.Map, hzMap)` work?

client_it_test.go
```diff
@@ -321,3 +321,112 @@ func TestClusterReconnection_ReconnectModeOff(t *testing.T) {
...
+	assert.IsType(t, new(hz.Set), hzSet)
+
+	if err = c.Shutdown(ctx); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestClient_PurgeDestroyedDistributedObjects(t *testing.T) {
```
yuce, line 377: The same comments for `TestClient_GetDistributedObjects` test

proxy_manager.go
```diff
@@ -45,61 +45,115 @@ func newProxyManager(bundle creationBundle) *proxyManager {
...
+	return p.invokeOnRandomTarget(ctx, request, handler)
+}
+
+func (m *proxyManager) getCachedObjectsInfo() []codec.DistributedObjectInfo {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	objects := make([]codec.DistributedObjectInfo, 0, len(m.proxies))
+	for _, p := range m.proxies {
+		objects = append(objects, codec.NewDistributedObjectInfo(p.name, p.serviceName))
+	}
```
yuce, line 117: Using the index in the for loop may simplify this code a bit:
```
objects := make([]codec.DistributedObjectInfo, len(m.proxies))
for i, p := range m.proxies {
	objects[i] = codec.NewDistributedObjectInfo(p.name, p.serviceName)
}
```

proxy_manager.go
```diff
@@ -45,61 +45,115 @@ func newProxyManager(bundle creationBundle) *proxyManager {
...
+	return objects
+}
+
+func (m *proxyManager) getCachedObjects() (map[string]interface{}, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	objects := make(map[string]interface{})
```
yuce, line 124: This one is a bit simpler:
```
objects := map[string]interface{}{}
```

yuce (changes requested):
Thanks for the PR! It looks good, just a few comments...

client_it_test.go
enozcan: It expects not type but expression to be passed. So no, it does not work.

client_it_test.go
enozcan: Do you mean using `Map/SetTester` methods or only generating this kind of names? If it's the latter, is there a particular reason to keep DDS names for tests in this format? I'm asking since I couldn't get the importance of naming. Bcz I'd say any string would work here.

proxy_manager.go
enozcan: `m.proxies` is of type map, so `i` refers to string keys in your ranged snippet. And I'd prefer to keep it that way instead of using another index variable outside of the loop's scope. Wdyt?

client_it_test.go
yuce: We add a random suffix to DSS names in tests, in order to not accidentally use the same name accross tests, like both `TestClient_GetDistributedObjects` and `TestClient_PurgeDestroyedDistributedObjects` use `get-distributed-objects-map` as the map name.

enozcan: Doesn't using `it.StartNewClusterWithOptions` in both tests ensure isolation? Otherwise these newly added tests would not work at all - as it asserts the number of objects created within the test body. It of course makes sense to use different names for the tests using the same cluster, but I was asking for the two tests I just added.

proxy_manager.go
```diff
@@ -148,10 +202,10 @@ func (m *proxyManager) proxyFor(ctx context.Context, serviceName string, objectN
 	}
 }
 
-func (m *proxyManager) createProxy(ctx context.Context, serviceName string, objectName string) (*proxy, error) {
+func (m *proxyManager) createProxy(ctx context.Context, serviceName string, objectName string, remote bool) (*proxy, error) {
```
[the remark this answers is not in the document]
enozcan: I also considered this but then decided to keep `m.createProxy` in favor of not passing manager's fields with each call to `newProxy`. However, going to change as you suggested.

client_it_test.go
enozcan: I assumed context timeout is used as the test timeout value based on the other tests through the file. And since multiple clients are created within the tests, I set it to 40s. I can change it to background if it's not supposed to be that way - I mean the test timeout duration.

client_it_test.go
enozcan: Are we sure all the tests destroy their objects after complete? I couldn't see such a cleanup but if so, and if all the integration tests run serially (not sure if they do), I'll change them to use an existing cluster. Otherwise, we'll not be able to assert the number of objects returned by the new API.

enozcan:
Closing this as the final decision is to return cluster-wide objects' information only, not data structure types like Map, Set, etc. To be implemented in a separate PR.

PR #605, opened 2021-08-10
Add paging predicate [API-308]

### LINKED ISSUE #484: Paging Predicate
Currently, the Go client does not support Hazelcast's [paging predicates](https://docs.hazelcast.org/docs/latest-dev/manual/html-single/index.html#filtering-with-paging-predicates).

martijndeheus:
This PR adds the paging predicate to Go-client.

Fixes #484

### DIFF
--- a/internal/proto/predicate_paging.go
+++ b/internal/proto/predicate_paging.go
@@ -1,0 +1,47 @@
+package proto
+
+import (
+	"github.com/hazelcast/hazelcast-go-client/internal/serialization"
+)
+
+type PagingPredicateHolder struct {
+	AnchorDataListHolder AnchorDataListHolder
+	PredicateData        *serialization.Data
+	ComparatorData       *serialization.Data
+	PageSize             int32
+	Page                 int32
+	IterationTypeId      byte
+	PartitionKeyData     *serialization.Data
+}
+
+func NewPagingPredicateHolder(
+	anchorDataListHolder AnchorDataListHolder,
+	predicateData *serialization.Data,
+	comparatorData *serialization.Data,
+	pageSize int32,
+	page int32,
+	iterationTypeId byte,
+	partitionKeyData *serialization.Data) PagingPredicateHolder {
+
+	return PagingPredicateHolder{
+		AnchorDataListHolder: anchorDataListHolder,
+		PredicateData:        predicateData,
+		ComparatorData:       comparatorData,
+		PageSize:             pageSize,
+		Page:                 page,
+		IterationTypeId:      iterationTypeId,
+		PartitionKeyData:     partitionKeyData,
+	}
+}
+
+type AnchorDataListHolder struct {
+	AnchorPageList []int32
+	AnchorDataList []Pair
+}
+
+func NewAnchorDataListHolder(anchorPageList []int32, anchorDataList []Pair) AnchorDataListHolder {
+	return AnchorDataListHolder{
+		AnchorPageList: anchorPageList,
+		AnchorDataList: anchorDataList,
+	}
+}
--- a/internal/proto/predicate_paging.go
+++ b/internal/proto/predicate_paging.go
@@ -5,13 +5,13 @@ import (
 )
 
 type PagingPredicateHolder struct {
-	AnchorDataListHolder AnchorDataListHolder
+	PartitionKeyData     *serialization.Data
 	PredicateData        *serialization.Data
 	ComparatorData       *serialization.Data
+	AnchorDataListHolder AnchorDataListHolder
 	PageSize             int32
 	Page                 int32
 	IterationTypeId      byte
-	PartitionKeyData     *serialization.Data
 }
 
 func NewPagingPredicateHolder(
--- a/internal/proto/predicate_paging.go
+++ b/internal/proto/predicate_paging.go
@@ -14,23 +14,15 @@ type PagingPredicateHolder struct {
 	IterationTypeId      byte
 }
 
-func NewPagingPredicateHolder(
-	anchorDataListHolder AnchorDataListHolder,
-	predicateData *serialization.Data,
-	comparatorData *serialization.Data,
-	pageSize int32,
-	page int32,
-	iterationTypeId byte,
-	partitionKeyData *serialization.Data) PagingPredicateHolder {
-
+func NewPagingPredicateHolder(a AnchorDataListHolder, pred *serialization.Data, c *serialization.Data, ps int32, p int32, i byte, pkd *serialization.Data) PagingPredicateHolder {
 	return PagingPredicateHolder{
-		AnchorDataListHolder: anchorDataListHolder,
-		PredicateData:        predicateData,
-		ComparatorData:       comparatorData,
-		PageSize:             pageSize,
-		Page:                 page,
-		IterationTypeId:      iterationTypeId,
-		PartitionKeyData:     partitionKeyData,
+		AnchorDataListHolder: a,
+		PredicateData:        pred,
+		ComparatorData:       c,
+		PageSize:             ps,
+		Page:                 p,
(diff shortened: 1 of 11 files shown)


predicate/predicate_paging.go
```diff
@@ -0,0 +1,260 @@
...
+	p.page = page
+}
+
+// GetPage return the current page of the paging predicate.
+func (p predPaging) GetPage() int32 {
+	return p.page
+}
+
+// AnchorList returns the anchor list of the paging predicate containing the first item of every page fetched with the predicate.
+func (p predPaging) AnchorList() []types.Entry {
```
yuce, line 90: This function is only required in the same package, isn't it? How about unexporting it? It's OK to have tests in the `predicate` package to access it if necessary.

predicate/predicate_paging.go
```diff
@@ -0,0 +1,260 @@
...
+func (p predPaging) WriteData(output serialization.DataOutput) {
+	output.WriteObject(p.internalPredicate)
+	output.WriteObject(p.comparator)
+	output.WriteInt32(p.page)
+	output.WriteInt32(p.pageSize)
+	output.WriteString(string(p.iterationType))
+	output.WriteInt32(int32(len(p.anchorList)))
+	for _, a := range p.anchorList {
+		output.WriteInt32(a.Key.(int32))
+		anchorEntry := a.Value.(types.Entry)
```
yuce, line 128: How about renaming `anchorEntry` to just `e`?

predicate/predicate_paging.go
```diff
@@ -0,0 +1,260 @@
...
+		output.WriteObject(anchorEntry.Key)
+		output.WriteObject(anchorEntry.Value)
+	}
+}
+
+func (p predPaging) String() string {
+	return fmt.Sprintf("paging, predicate=%s, pageSize=%d, page=%d",
```
yuce, line 135: How about returning something like (`Paging(predicate=%s, ...)` to be consistent with other predicates?

predicate/predicate_paging.go
```diff
@@ -0,0 +1,260 @@
...
+	"github.com/hazelcast/hazelcast-go-client/internal/util/nilutil"
+	"github.com/hazelcast/hazelcast-go-client/serialization"
+	"github.com/hazelcast/hazelcast-go-client/types"
+)
+
+type predPaging struct {
+	comparator        serialization.IdentifiedDataSerializable
+	internalPredicate Predicate
+	iterationType     IterationType
+	anchorList        []types.Entry
```
yuce, line 18: How about using shorter names for fields, such as: `cmp, pred, iterType, anchors`?

predicate/predicate_paging.go
```diff
@@ -0,0 +1,260 @@
...
+		output.WriteObject(anchorEntry.Value)
+	}
+}
+
+func (p predPaging) String() string {
+	return fmt.Sprintf("paging, predicate=%s, pageSize=%d, page=%d",
+		p.internalPredicate.String(), p.pageSize, p.page)
+}
+
+func ValidateAndSerializePagingPredicate(
```
yuce, line 139: We shouldn't expose `iserialization.Data`. Let me think a bit about this function...

yuce (changes requested):
Thanks a lot for the PR! Here are few comments. I will continue to review.

predicate/predicate_paging.go
martijndeheus: One of the things I ran into is that if we don't want to expose the `predPaging` object (like for the other predicates). We need all the serialization logic in this package causing us to need the `func(interface{}) (*iserialization.Data, error)` in this package.

yuce: `iserialization.Data` is a real nuisance. It's only a `[]byte` so I'll probably expose it. I'm currently fixing the client binary protocol generator for the Go client. I think we'll have a better way to handle this one once that's over.

predicate/predicate_paging.go
martijndeheus: I believe this may be useful to the user as well as part of the API.

predicate/predicate_paging.go
```diff
@@ -0,0 +1,260 @@
...
+
+func (p predPaging) FactoryID() int32 {
+	return factoryID
+}
+
+func (p predPaging) ClassID() int32 {
+	return 15
+}
+
+func (p predPaging) ReadData(input serialization.DataInput) {
```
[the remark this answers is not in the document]
martijndeheus: It is not used because the actual type that is serialized is `PagingPredicateHolder`. The same is true in other clients, but there is also an implementation for this method (which is a bit confusing). I'll remove the implementation of the function and add a comment.

internal/proto/predicate_paging.go
```diff
@@ -0,0 +1,47 @@
...
+
+func NewPagingPredicateHolder(
+	anchorDataListHolder AnchorDataListHolder,
+	predicateData *serialization.Data,
+	comparatorData *serialization.Data,
+	pageSize int32,
+	page int32,
+	iterationTypeId byte,
+	partitionKeyData *serialization.Data) PagingPredicateHolder {
+
```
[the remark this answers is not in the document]
martijndeheus: Addressed in d226845785e034a27f9e4d255c6879f1f22aa3d5

predicate/predicate_it_test.go
```diff
@@ -219,3 +285,54 @@ func createFixture(m *hz.Map) {
...
+		if !assert.Subset(t, target, values) {
+			t.FailNow()
+		}
+		if !assert.Subset(t, values, target) {
+			t.FailNow()
+		}
+	})
+}
+
+// The paging predicate needs a separate fixture because the paging predicate
```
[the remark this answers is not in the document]
martijndeheus: I tried to move it into the other function but it becomes a bit messy. I think it is cleaner to separate the functions. I updated the comment in d226845785e034a27f9e4d255c6879f1f22aa3d5.

predicate/predicate_paging.go
```diff
@@ -0,0 +1,260 @@
...
+	case KEY:
+		return 0
+	case VALUE:
+		return 1
+	case ENTRY:
+		return 2
+	default:
+		return -1
+	}
+}
```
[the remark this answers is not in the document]
martijndeheus: Addressed in d226845785e034a27f9e4d255c6879f1f22aa3d5

predicate/predicate_paging.go
```diff
@@ -0,0 +1,260 @@
...
+
+// NextPage sets the page of the paging predicate to the next page to fetch the next results in the following operation.
+func (p *predPaging) NextPage() {
+	p.page++
+}
+
+// PrevPage sets the page of the paging predicate to the previous page to fetch the previous results in the following operation.
+func (p *predPaging) PrevPage() {
+	p.page--
```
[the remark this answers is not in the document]
martijndeheus: Addressed in d226845

predicate/predicate_paging.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// SetPage sets the page to a given page number to fetch the results at that page in the following operation.
+func (p *predPaging) SetPage(page int32) {
+	p.page = page
+}
+
+// GetPage return the current page of the paging predicate.
+func (p predPaging) GetPage() int32 {
```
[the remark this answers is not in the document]
martijndeheus: Addressed in d226845

predicate/predicate_paging.go
martijndeheus: Addressed in d226845

predicate/predicate_paging.go
martijndeheus: Addressed in d226845

yuce:
Predicates are deprecated, and the project that requires this feature is over, closing.

PR #606, opened 2021-08-10
Add GetDistributedObjectsInfo API.

enozcan:
Replaces #602 

- Moves DistributedObjectInfo from internal pkg to types. It's also the return type of the new API.
- Unlike #602, this does not create local proxies for fetched distributed objects until client makes `Get*` call.

### DIFF
--- a/client.go
+++ b/client.go
@@ -211,21 +211,7 @@ func (c *Client) GetDistributedObjectsInfo(ctx context.Context) ([]types.Distrib
 	if err != nil {
 		return nil, err
 	}
-	objects := codec.DecodeClientGetDistributedObjectsResponse(resp)
-	result := make([]types.DistributedObjectInfo, 0, len(objects))
-	cached := make(map[types.DistributedObjectInfo]struct{})
-	for _, o := range c.proxyManager.getCachedObjectsInfo() {
-		cached[o] = struct{}{}
-	}
-	for _, o := range objects {
-		result = append(result, o)
-		delete(cached, o)
-	}
-	for o := range cached {
-		// Purge the stale object from the manager's cache.
-		c.proxyManager.remove(o.ServiceName, o.Name)
-	}
-	return result, nil
+	return codec.DecodeClientGetDistributedObjectsResponse(resp), nil
 }
 
 func (c *Client) start(ctx context.Context) error {
--- a/client.go
+++ b/client.go
@@ -206,34 +206,25 @@ func (c *Client) GetDistributedObjectsInfo(ctx context.Context) ([]types.Distrib
 	if atomic.LoadInt32(&c.state) != ready {
 		return nil, hzerrors.ErrClientNotActive
 	}
-
 	request := codec.EncodeClientGetDistributedObjectsRequest()
 	resp, err := c.proxyManager.invokeOnRandomTarget(ctx, request, nil)
 	if err != nil {
 		return nil, err
 	}
-
 	objects := codec.DecodeClientGetDistributedObjectsResponse(resp)
 	result := make([]types.DistributedObjectInfo, 0, len(objects))
-
 	cached := make(map[types.DistributedObjectInfo]struct{})
 	for _, o := range c.proxyManager.getCachedObjectsInfo() {
 		cached[o] = struct{}{}
 	}
-
 	for _, o := range objects {
-		if !serviceSupported(o.ServiceName()) {
-			continue
-		}
 		result = append(result, o)
 		delete(cached, o)
 	}
-
 	for o := range cached {
 		// Purge the stale object from the manager's cache.
-		c.proxyManager.remove(o.ServiceName(), o.Name())
+		c.proxyManager.remove(o.ServiceName, o.Name)
 	}
-
 	return result, nil
 }
 
--- a/client.go
+++ b/client.go
@@ -31,6 +31,7 @@ import (
 	"github.com/hazelcast/hazelcast-go-client/internal/invocation"
 	ilogger "github.com/hazelcast/hazelcast-go-client/internal/logger"
 	"github.com/hazelcast/hazelcast-go-client/internal/proto"
+	"github.com/hazelcast/hazelcast-go-client/internal/proto/codec"
 	iproxy "github.com/hazelcast/hazelcast-go-client/internal/proxy"
 	"github.com/hazelcast/hazelcast-go-client/internal/security"
 	"github.com/hazelcast/hazelcast-go-client/internal/serialization"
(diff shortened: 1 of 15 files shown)


client.go
```diff
@@ -200,6 +201,42 @@ func (c *Client) GetPNCounter(ctx context.Context, name string) (*PNCounter, err
...
+	objects := codec.DecodeClientGetDistributedObjectsResponse(resp)
+	result := make([]types.DistributedObjectInfo, 0, len(objects))
+
+	cached := make(map[types.DistributedObjectInfo]struct{})
+	for _, o := range c.proxyManager.getCachedObjectsInfo() {
+		cached[o] = struct{}{}
+	}
+
+	for _, o := range objects {
+		if !serviceSupported(o.ServiceName()) {
```
yuce, line 225: There's no need to check whether the service is supported, let's return everything.

client.go
```diff
@@ -200,6 +201,42 @@ func (c *Client) GetPNCounter(ctx context.Context, name string) (*PNCounter, err
...
+		return nil, err
+	}
+
+	objects := codec.DecodeClientGetDistributedObjectsResponse(resp)
+	result := make([]types.DistributedObjectInfo, 0, len(objects))
+
+	cached := make(map[types.DistributedObjectInfo]struct{})
+	for _, o := range c.proxyManager.getCachedObjectsInfo() {
+		cached[o] = struct{}{}
+	}
```
yuce, line 222: I think we shouldn't do anything with the proxy cache anymore.

client.go
```diff
@@ -200,6 +201,42 @@ func (c *Client) GetPNCounter(ctx context.Context, name string) (*PNCounter, err
...
+
+	for _, o := range objects {
+		if !serviceSupported(o.ServiceName()) {
+			continue
+		}
+		result = append(result, o)
+		delete(cached, o)
```
yuce, line 229: I think we shouldn't do anything with the proxy cache anymore.

client.go
```diff
@@ -200,6 +201,42 @@ func (c *Client) GetPNCounter(ctx context.Context, name string) (*PNCounter, err
...
+	}
+
+	for o := range cached {
+		// Purge the stale object from the manager's cache.
+		c.proxyManager.remove(o.ServiceName(), o.Name())
+	}
```
yuce, line 235: I think we shouldn't do anything with the proxy cache anymore.

internal/it/util.go
```diff
@@ -190,6 +190,10 @@ func MustClient(client *hz.Client, err error) *hz.Client {
 	return client
 }
 
+func NewUniqueServiceName(service string) string {
```
yuce, line 193: This function seems to be creating object names, can you rename it to `NewObjectName`?

types/distributed_objects_info.go
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package types
+
+type DistributedObjectInfo struct {
+	name        string
+	serviceName string
```
yuce, line 21: Let's export these fields.

types/distributed_objects_info.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func (i *DistributedObjectInfo) Name() string {
+	return i.name
+}
+
+func (i *DistributedObjectInfo) ServiceName() string {
+	return i.serviceName
+}
```
yuce, line 30: No need for these anymore.

client.go
```diff
@@ -200,6 +201,42 @@ func (c *Client) GetPNCounter(ctx context.Context, name string) (*PNCounter, err
 	return c.proxyManager.getPNCounter(ctx, name)
 }
 
+// GetDistributedObjectsInfo returns the information of all objects created cluster-wide.
+func (c *Client) GetDistributedObjectsInfo(ctx context.Context) ([]types.DistributedObjectInfo, error) {
+	if atomic.LoadInt32(&c.state) != ready {
+		return nil, hzerrors.ErrClientNotActive
+	}
+
```
yuce, line 209: Could you remove the blank lines in this function?

internal/proto/codec/builtin.go
```diff
@@ -620,6 +620,24 @@ func DecodeListMultiFrameForDataContainsNullable(frameIterator *proto.ForwardFra
 	return result
 }
 
+func DecodeListMultiFrameForDistributedObjectInfo(frameIterator *proto.ForwardFrameIterator) []types.DistributedObjectInfo {
+	result := make([]types.DistributedObjectInfo, 0)
```
yuce, line 624: Can be simplified to:
```
var result []types.DistributedObjectInfo
```

proxy.go
```diff
@@ -53,6 +53,21 @@ const (
...
+		ServiceNameTopic,
+		ServiceNameList,
+		ServiceNameSet,
+		ServiceNamePNCounter:
+		return true
+	default:
+		return false
+	}
+}
+
```
yuce, line 70: We can remove this function.

types/distributed_objects_info.go
```diff
@@ -0,0 +1,34 @@
...
+
+func NewDistributedObjectInfo(name string, serviceName string) DistributedObjectInfo {
+	return DistributedObjectInfo{name: name, serviceName: serviceName}
+}
```
yuce, line 34: The user doesn't need this function, so it's better to move it to internal. How about `internal/util.go`?

yuce (changes requested):
Thanks for the PR! Just a few comments to make it even better...

client.go
enozcan: Isn't it still a good way of removing stale proxies?

types/distributed_objects_info.go
enozcan: I think no one needs it after exporting the fields. Removing it.

client.go
enozcan: addressed in f2a4add

internal/it/util.go
enozcan: addressed in f2a4add

types/distributed_objects_info.go
enozcan: addressed in f2a4add

types/distributed_objects_info.go
enozcan: addressed in f2a4add

client.go
enozcan: addressed in f2a4add

internal/proto/codec/builtin.go
enozcan: addressed in f2a4add

proxy.go
enozcan: addressed in f2a4add

types/distributed_objects_info.go
enozcan: addressed in f2a4add

enozcan:
@yuce addressed your comments in f2a4add except for purging the proxy cache - I left it unchanged to discuss further. Could you PTAL?

client.go
yuce: I think purging stale proxies is a separate concern. As a user, I wouldn't expect `GetObjectsInfo` call to mutate any internal state.

yuce: Not sure about what to call the purge function though. If there's no immediate need from the Turbine team, maybe we can postpone adding it to another PR (unless you have a good name for that function) .

yuce:
@enozcan Thanks for the update! I think this PR is good to be merged once the code related to proxy cache is removed.

client.go
enozcan: Alright, fair enough.  Addressed in cd2887a

client.go
enozcan: addressed in cd2887a

client.go
enozcan: addressed in cd2887a

client.go
enozcan: Apart from the context of this pr, could it be handled via DistributedObjectListeners in proxy manager - i.e. purge the cache on object destroy or something?

yuce (approved):
Thanks for the PR, it looks great!

client.go
yuce: That can be an option, I can't think of a problem off the top of my head.

enozcan:
Thanks for your review!

PR #612, opened 2021-08-19
Return same proxy struct instance from Get calls

### LINKED ISSUE #609: Client does not return the same object for consecutive Get* calls.
Hazelcast Go Client version: v1.0.0

#### Actual behaviour

When getting data structures from client, it creates a **new** struct for each call after wrapping the existing proxy of it. It's not a big problem for Map, Set, etc. But that causes having different vector clocks for PN counters with the same name. This also causes allocation of new structs for each GetDDS call even with the same name.

#### Expected behaviour

Not proxy but the struct itself could've been cached and served upon GetDDS calls. That would prevent creating new (and distinct) structs for each call.

```go
counter1, _ := cli.GetPNCounter(ctx, "pn-counter")
counter2, _ := cli.GetPNCounter(ctx, "pn-counter")
fmt.Printf("c1:%p, clk:%p\nc2:%p, clk:%p\n", counter1, counter1.clock, counter2, counter2.clock)
// c1:0xc00022e460, clk:0xc000232240
// c2:0xc00022e480, clk:0xc000232270

counter1.GetAndIncrement(ctx)
counter2.GetAndIncrement(ctx)
fmt.Printf("clock1:%v\nclock2:%v\n", counter1.clock, counter2.clock)
// clock1:map[9d83fd60-081c-49f6-bd28-33373f4fea37:-9223372036854775800]
// clock2:map[9d83fd60-081c-49f6-bd28-33373f4fea37:-9223372036854775799]
```

puzpuzpuz:
Fixes #609

Also moves linter checks before tests in GH actions definition. It makes no sense to wait for all tests to pass when `fieldalignment` or `go vet` will fail the GHA run. So, it's better to fail fast.

### DIFF
--- a/proxy_manager.go
+++ b/proxy_manager.go
@@ -53,7 +53,7 @@ func newProxyManager(bundle creationBundle) *proxyManager {
 }
 
 func (m *proxyManager) getMap(ctx context.Context, name string) (*Map, error) {
-	p, err := m.proxyWrapperFor(ctx, ServiceNameMap, name, func(p *proxy) (interface{}, error) {
+	p, err := m.proxyFor(ctx, ServiceNameMap, name, func(p *proxy) (interface{}, error) {
 		return newMap(p), nil
 	})
 	if err != nil {
@@ -63,7 +63,7 @@ func (m *proxyManager) getMap(ctx context.Context, name string) (*Map, error) {
 }
 
 func (m *proxyManager) getReplicatedMap(ctx context.Context, name string) (*ReplicatedMap, error) {
-	p, err := m.proxyWrapperFor(ctx, ServiceNameReplicatedMap, name, func(p *proxy) (interface{}, error) {
+	p, err := m.proxyFor(ctx, ServiceNameReplicatedMap, name, func(p *proxy) (interface{}, error) {
 		return newReplicatedMap(p, m.refIDGenerator)
 	})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


proxy_manager.go
```diff
@@ -142,24 +156,37 @@ func (m *proxyManager) remove(serviceName string, objectName string) bool {
...
 	m.mu.RLock()
-	obj, ok := m.proxies[name]
+	wrapper, ok := m.proxies[name]
 	m.mu.RUnlock()
 	if ok {
-		return obj, nil
+		return wrapper, nil
+	}
+	m.mu.Lock()
+	defer m.mu.Unlock()
```
enozcan, line 173: I see it's still WIP but an early comment: Here, deferring the unlock holds the writer lock until the member responses back. This will block readers especially when - say, there is a retry or slow operation happens for this call. Do you think that makes sense to consider or it is negligible? I mean, all operations are supposed to be slow under such a condition?

puzpuzpuz: > EDIT: never mind, I missed the assignment at the very end. The comment does not make sense.

Good to know. I think, we were blocking all callers while the network I/O was happening here before my PR. In theory, we could improve that with striped locking, but I have a feeling that such optimization would only help in edge scenarios. WDYT?

enozcan: > we could improve that with striped locking, but I have a feeling that such optimization would only help in edge scenarios.

Yes I also was wondering how likely such edge cases are to happen. But I feel like it's not worth considering - at least for this block.

puzpuzpuz: Such edge cases assume that you're doing crazy things, say, like creating a new data structure on each HTTP request, or something that weird.

enozcan (approved):
LGTM. Thanks for the fix.

yuce (approved):
THanks, looks good!

puzpuzpuz:
@enozcan @yuce thanks for your code reviews!

PR #613, opened 2021-08-20
Fix cluster view listener logic [API-693]

### LINKED ISSUE #600: Client tries to connect non-existent members
Cluster size, i.e. the number of Hazelcast cluster members: 2+

#### Expected behaviour

When a connection to a member is lost, use only the addresses from member list for reconnection attempts, not the cached addresses of connection manager.

#### Actual behaviour

When the member address doesn't exist anymore, getting the following log, although it's not possible to fix that connection:
```
INFO : found a broken connection to: 10.1.202.216:5701, trying to fix it
```

Follow https://github.com/hazelcast-guides/kubernetes/blob/master/doc/modules/ROOT/pages/index.adoc to create a Hazelcast cluster on Kubernetes and run the Go client code. Then, stop one of the members and check client logs:
```
kubectl delete pod hz-hazelcast-0
kubectl logs hazelcast-client
```

puzpuzpuz:
Fixes #600

Depends on #608 (the fix is in the very last commit of this branch)

Members (i.e. cluster) view update handler incorrectly dealt with MemberInfo pointers while iterating over the received members view. This led to corruption of the stored members list.
    
Another problem was that connection manager was registering cluster view listeners only on the seed connections while it's enough to register it only on a single connection. It also didn't re-register the listener when connections were removed which led to infinite re-connection attempts.

Note: this is a quick attempt to do the fix (e.g. that's why there are no new tests), so there may be better approaches.

### DIFF
--- a/client.go
+++ b/client.go
@@ -33,7 +33,6 @@ import (
 	"github.com/hazelcast/hazelcast-go-client/internal/proto"
 	"github.com/hazelcast/hazelcast-go-client/internal/proto/codec"
 	iproxy "github.com/hazelcast/hazelcast-go-client/internal/proxy"
-	"github.com/hazelcast/hazelcast-go-client/internal/security"
 	"github.com/hazelcast/hazelcast-go-client/internal/serialization"
 	"github.com/hazelcast/hazelcast-go-client/internal/stats"
 	"github.com/hazelcast/hazelcast-go-client/types"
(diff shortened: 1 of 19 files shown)


puzpuzpuz:
@yuce this PR is a quick attempt to fix #600. Feel free to merge it as is, or close it, or build a better fix on top of it.

yuce:
@puzpuzpuz Hi Andrey, I hope everything's going great with you! Closing this one in favor of https://github.com/hazelcast/hazelcast-go-client/pull/617 Thanks again for the PR!

puzpuzpuz:
> @puzpuzpuz Hi Andrey, I hope everything's going great with you! Closing this one in favor of #617 Thanks again for the PR!

Sure. Good to know that the issue is now solved.

PR #614, opened 2021-08-25
Add code sample for FlakeIDGenerator.

enozcan:
Follow up to #611

Adds code sample for FlakeIDGenerator. It generates`(idPrefetchCount * 2)` ids to indicate that prefetch count does not limit the total number of generated IDs by generator. Instead, the result of prefetch behavior can be observed from the output at 10th id:
```
flakeID(8):     482885333181530113
flakeID(9):     482885333181595649
			  ^^
flakeID(10):    482885333185200129
			  ^^
flakeID(11):    482885333185265665
flakeID(12):    482885333185331201
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -17,9 +17,11 @@
 package hazelcast
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/hazelcast/hazelcast-go-client/cluster"
+	"github.com/hazelcast/hazelcast-go-client/internal/hzerrors"
 	validate "github.com/hazelcast/hazelcast-go-client/internal/util/validationutil"
 	"github.com/hazelcast/hazelcast-go-client/logger"
 	"github.com/hazelcast/hazelcast-go-client/serialization"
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -139,6 +139,7 @@ func (c *Config) ensureMembershipListeners() {
 	}
 }
 
+// AddFlakeIDGenerator validates the values and adds new FlakeIDGeneratorConfig with the given name.
```
yuce, line 142: Not related to this line, but can you check if a flake ID generator was created with the same name?

yuce (approved):
Looks good, apart from checking the existence of a flakeID generator before it is added.

config.go
enozcan: Should we assume it as an error or just override it like we do in both _client and member_ configs in Java (might be valid for other clients as well, not sure about them) - i.e. `Config#addMapConfig`, `Config#addListConfig`, `Config#addFlakeIdGeneratorConfig`, etc.? Asking since I feel this is the config idiom we follow so far.

yuce: Let's make it an error. That's what we did everywhere else, e..g, addding serialization factories.

enozcan: addressed in fa6c1da

enozcan:
Merging this as the latest changes are trivial. Thanks for the review.

PR #615, opened 2021-08-26
Failover Updates [API-626]

yuce:
- Updated failover support to work in a way closer to how Java client works
- Fixed failover support, so only clusters in the failover config are used when it is enabled
- Fixed cluster connection timeouts.
- Added a code sample for failover support.

### DIFF
--- a/cluster/connection_strategy_config.go
+++ b/cluster/connection_strategy_config.go
@@ -63,6 +63,7 @@ type ConnectionStrategyConfig struct {
 	// Retry contains the backoff configuration.
 	Retry ConnectionRetryConfig
 	// Timeout is the maximum time before giving up reconnecting to a cluster.
+	// Default is 0, infinite duration.
 	Timeout types.Duration `json:",omitempty"`
 	// ReconnectMode enables or disables reconnecting to a cluster.
 	ReconnectMode ReconnectMode `json:",omitempty"`
--- a/cluster/connection_strategy_config.go
+++ b/cluster/connection_strategy_config.go
@@ -63,7 +63,7 @@ type ConnectionStrategyConfig struct {
 	// Retry contains the backoff configuration.
 	Retry ConnectionRetryConfig
 	// Timeout is the maximum time before giving up reconnecting to a cluster.
-	// Default is 0, infinite duration.
+	// Default is 0, infinite duration when failover is not enabled and 120 seconds when it is enabled.
 	Timeout types.Duration `json:",omitempty"`
 	// ReconnectMode enables or disables reconnecting to a cluster.
 	ReconnectMode ReconnectMode `json:",omitempty"`
--- a/cluster/connection_strategy_config.go
+++ b/cluster/connection_strategy_config.go
@@ -73,7 +73,8 @@ func (c ConnectionStrategyConfig) Clone() ConnectionStrategyConfig {
 }
 
 func (c *ConnectionStrategyConfig) Validate() error {
-	if err := validate.NonNegativeDuration(&c.Timeout, 1<<63-1, "invalid timeout"); err != nil {
+	// we set the default to 0 in order to be able to override it in failover and cluster configs
+	if err := validate.NonNegativeDuration(&c.Timeout, 0, "invalid timeout"); err != nil {
 		return err
 	}
 	return c.Retry.Validate()
--- a/cluster/failover_config.go
+++ b/cluster/failover_config.go
@@ -22,6 +22,8 @@ import (
 	"reflect"
 
 	"github.com/hazelcast/hazelcast-go-client/hzerrors"
+	"github.com/hazelcast/hazelcast-go-client/internal"
+	"github.com/hazelcast/hazelcast-go-client/types"
 )
 
 // FailoverConfig allows configuring multiple client configs to be used
@@ -79,7 +81,8 @@ func (c *FailoverConfig) Clone() FailoverConfig {
 	}
 }
 
-func (c *FailoverConfig) Validate(rootConfig Config) error {
+func (c *FailoverConfig) Validate(root Config) error {
+	// TODO: move code that references root config to the main config
 	if !c.Enabled {
 		return nil
 	}
@@ -92,12 +95,19 @@ func (c *FailoverConfig) Validate(rootConfig Config) error {
 	if len(c.Configs) == 0 {
 		return fmt.Errorf("at least one failover cluster config expected: %w", hzerrors.ErrIllegalArgument)
 	}
-	for i, cc := range c.Configs {
-		if err := cc.Validate(); err != nil {
+	for i := range c.Configs {
+		ccp := &c.Configs[i]
+		// if a cluster connection timeout was not specified, set it to default.
+		t := ccp.ConnectionStrategy.Timeout
+		if t == 0 {
+			t = types.Duration(internal.DefaultConnectionTimeoutWithFailover)
+		}
+		if err := ccp.Validate(); err != nil {
 			return err
 		}
+		ccp.ConnectionStrategy.Timeout = t
 		// validate diffs in the cluster configs
-		if !equalConfigs(rootConfig, cc) {
+		if !equalConfigs(root, *ccp) {
 			return fmt.Errorf("cluster config %d has unallowed differences with the first config: %w",
 				i, hzerrors.ErrIllegalArgument)
 		}
@@ -118,12 +128,14 @@ func equalConfigs(c1, c2 Config) bool {
 // * Network.SSL
 // * Network.Addresses
 // * Cloud
+// * ConnectionStrategy
 func sanitizedConfig(c Config) Config {
 	c.Name = ""
 	c.Security = SecurityConfig{}
 	c.Network.SSL = SSLConfig{}
 	c.Network.Addresses = nil
 	c.Cloud = CloudConfig{}
+	c.ConnectionStrategy = ConnectionStrategyConfig{}
 	return c
 }
 
--- a/config_test.go
+++ b/config_test.go
@@ -375,7 +375,7 @@ func checkDefault(t *testing.T, c *hazelcast.Config) {
 	assert.Equal(t, "", c.Cluster.Cloud.Token)
 
 	assert.Equal(t, cluster.ReconnectModeOn, c.Cluster.ConnectionStrategy.ReconnectMode)
-	assert.Equal(t, types.Duration(9223372036854775807), c.Cluster.ConnectionStrategy.Timeout)
+	assert.Equal(t, types.Duration(31622400000000000), c.Cluster.ConnectionStrategy.Timeout)
 	cr := &c.Cluster.ConnectionStrategy.Retry
 	assert.Equal(t, types.Duration(1*time.Second), cr.InitialBackoff)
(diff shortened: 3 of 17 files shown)


yuce:
@sancar @alisengul53 Could you you check this PR? It should be ready to review.

cluster/connection_strategy_config.go
```diff
@@ -73,7 +73,8 @@ func (c ConnectionStrategyConfig) Clone() ConnectionStrategyConfig {
 }
 
 func (c *ConnectionStrategyConfig) Validate() error {
-	if err := validate.NonNegativeDuration(&c.Timeout, 1<<63-1, "invalid timeout"); err != nil {
+	// we set the default to 0 in order to be able to override it in failover and cluster configs
+	if err := validate.NonNegativeDuration(&c.Timeout, 0, "invalid timeout"); err != nil {
```
sancar, line 77: As user can set any non-negative value. I think setting it to -1 here makes more sense. 
If user sets it to 0, we can differentiate it. See https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/config/ConnectionRetryConfig.java#L31

cluster/failover_config.go
```diff
@@ -92,12 +95,19 @@ func (c *FailoverConfig) Validate(rootConfig Config) error {
 	if len(c.Configs) == 0 {
 		return fmt.Errorf("at least one failover cluster config expected: %w", hzerrors.ErrIllegalArgument)
 	}
-	for i, cc := range c.Configs {
-		if err := cc.Validate(); err != nil {
+	for i := range c.Configs {
+		ccp := &c.Configs[i]
+		// if a cluster connection timeout was not specified, set it to default.
+		t := ccp.ConnectionStrategy.Timeout
+		if t == 0 {
```
sancar, line 102: Instead of 0, it would make sense to compare against a well named static to make it more obvious why we are comparing against it. 
see https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/connection/tcp/TcpClientConnectionManager.java#L299

cluster/failover_config.go
```diff
@@ -92,12 +95,19 @@ func (c *FailoverConfig) Validate(rootConfig Config) error {
...
+			t = types.Duration(internal.DefaultConnectionTimeoutWithFailover)
+		}
+		if err := ccp.Validate(); err != nil {
 			return err
 		}
+		ccp.ConnectionStrategy.Timeout = t
```
sancar, line 108: This is updating users config right ? 
I think we should avoid changing what user provided. It should be read only anywhere internal to the client. 
This way, it will be easier to reason. Changing the users config caused really unexpected bugs in the past for java client.

config_test.go
```diff
@@ -375,7 +375,7 @@ func checkDefault(t *testing.T, c *hazelcast.Config) {
 	assert.Equal(t, "", c.Cluster.Cloud.Token)
 
 	assert.Equal(t, cluster.ReconnectModeOn, c.Cluster.ConnectionStrategy.ReconnectMode)
-	assert.Equal(t, types.Duration(9223372036854775807), c.Cluster.ConnectionStrategy.Timeout)
+	assert.Equal(t, types.Duration(31622400000000000), c.Cluster.ConnectionStrategy.Timeout)
```
sancar, line 378: What is this value ? Can we make it more obvious so that we understand what it is ?

internal/cb/options.go
```diff
@@ -47,6 +48,7 @@ func DefaultCircuitBreakerOptions() *CircuitBreakerOptions {
 		MaxRetries:      0,
 		MaxFailureCount: 0,
 		ResetTimeout:    0,
+		Timeout:         time.Duration(1<<63 - 1),
```
sancar, line 51: I keep seeing this value throughout the code base. I understand that this is max duration but can we assign it to a utility constant and use the variable instead.

internal/cluster/connection_manager.go
```diff
@@ -392,48 +390,42 @@ func (m *ConnectionManager) removeConnection(conn *Connection) {
...
-		if attempt == maxAttempts {
-			m.logger.Debug(func() string {
-				return fmt.Sprintf("ConnectionManager: giving up connection attempts to cluster: %s", cluster.ClusterName)
-			})
-			return "", cb.WrapNonRetryableError(fmt.Errorf("giving up connection attempts to cluster: %w", hzerrors.ErrIllegalState))
-		}
+func (m *ConnectionManager) tryConnectCandidateCluster(ctx context.Context, cluster *CandidateCluster, cs *pubcluster.ConnectionStrategyConfig) (pubcluster.Address, error) {
+	tryCount := math.MaxInt32
+	if m.failoverConfig.Enabled {
+		tryCount = 1
```
sancar, line 416: The old way was much more simpler to understand. 
I am against using CircuitBreaker for everything before we really decide to utilize its functionality. 

When I read this code it is hard to reason about. It feels like when failover config is not enabled we will try infinitely many but we are not. There is a second timeout parameter of CircuitBreaker which handles it.

internal/cluster/failover_service.go
```diff
@@ -24,63 +24,57 @@ import (
...
-		candidateClusters = append(candidateClusters, ctx)
+		cc.AddressProvider, cc.AddressTranslator = addrFn(&c, logger)
+		candidates = append(candidates, cc)
 	}
 
 	return &FailoverService{
-		isClientRunningFn: isClientRunningFn,
-		maxTryCount:       maxTryCount,
-		candidateClusters: candidateClusters,
+		clientRunningFn:   clientRunningFn,
```
sancar, line 66: It looks like this field is not used anymore.

internal/common.go
```diff
@@ -20,8 +20,8 @@ const (
 	// ClientType is used in the Management
 	ClientType         = "GOO"
 	AggregateFactoryID = -29
-)
 
-// ClientVersion is the build time version
-// TODO: This should be replace with a build time version variable, BuildInfo etc.
-var ClientVersion = "1.0.0"
+	// ClientVersion is the build time version
+	// TODO: This should be replaced with a build time version variable, BuildInfo etc.
```
sancar, line 25: Can we remove TODO's from the codebase and open issues for them if needed instead ?

internal/defaults.go
```diff
@@ -0,0 +1,8 @@
+package internal
+
+import "time"
+
+const (
+	DefaultConnectionTimeoutWithoutFailover = 24 * 366 * time.Hour
```
sancar, line 6: Why is this different than the value that I see for the max duration ?  
(1<<63 - 1)
This is a year right , why don't we use infinite ?

cluster/connection_strategy_config.go
yuce: Why do we allow `0` as a timeout value in the Java client?

cluster/failover_config.go
yuce: Since the zero value of `ccp.ConnectionStrategy.Timeout` is `0` it is idiomatic to use it.

cluster/failover_config.go
yuce: It wasn't supposed to, but it did, due to validation of the main config instead of a clone of it. Fixed in: https://github.com/hazelcast/hazelcast-go-client/pull/615/commits/bd77cd8b5727b5cbfd93ebe26473f2ea38becb19

internal/cluster/connection_manager.go
yuce: The circuit breaking behavior of CircuitBreakers in the code are not activated, so they just retry. Isn't it better to have a single way of having retry logic instead of repeating it throughout the code?

internal/cluster/failover_service.go
```diff
@@ -24,63 +24,57 @@ import (
...
-	for s.isClientRunningFn() && tryCount < s.maxTryCount {
-		for i := 0; i < len(s.candidateClusters); i++ {
-			if addr, connected := fn(s.Next()); connected {
-				return addr, true
-			}
-		}
-		tryCount++
-	}
-	return "", false
+func (s *FailoverService) TryNextCluster(fn func(next *CandidateCluster) (pubcluster.Address, error)) (pubcluster.Address, error) {
```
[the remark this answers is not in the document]
yuce: Removed in: https://github.com/hazelcast/hazelcast-go-client/pull/615/commits/bd77cd8b5727b5cbfd93ebe26473f2ea38becb19

internal/cluster/failover_service.go
yuce: Removed in: https://github.com/hazelcast/hazelcast-go-client/pull/615/commits/12c18628c7cb62180f904ad1f6e6c98a8a96a77e

internal/defaults.go
yuce: The timeout is added to the current time to create the deadline. If we use max duration, then it would overflow. Isn't 1 year equivalent to infinity?

cluster/connection_strategy_config.go
sancar: You can disallow it if you want. then don't check for non-negative but allow only positivies. 
this way, user will not be able to set zero and we can differentiate if it is a user set value or not. 

When it comes to why, it is hard for me to remember the reasoning when we first design it. 
If I try to think something today, a user would set `0` if s/he don't want the client to retry.

cluster/failover_config.go
sancar: 0 is ok but it should have a name to make it more obvious to the readers like me. 
I would like to see it assign it to a constant and named something like DEFAULT_CONNECTION_TIMEOUT/UNSET_CONNECTION_TIMEOUT.

internal/cluster/connection_manager.go
sancar: Then lets make this part much more easier to read. 
How about if we have two kind of CircuitBreakers, one with the `tryCount` and other with `timeout` . 
And one can never set both of them. We don't necessarily create another class, I think we can easily have two flavors of constructors for CircuitBreaker. What do you think ?

internal/cb/options.go
yuce: I don't want the `cb` package to depend on anything else, so intrdocude a constant in that package: https://github.com/hazelcast/hazelcast-go-client/pull/615/commits/a5f8e538d5c91e4d16758c59c2acb159f18c5a8c

internal/common.go
yuce: OK, will do that in a separate PR.

internal/defaults.go
sancar: I think we can easily fix the code to avoid addition so that we can get rid of confusing magic numbers

cluster/connection_strategy_config.go
sancar: I see. Since we don't want to get rid of the convenience of public Config, we have to live with it. 
Then, make sure that meaning of zero is well documented. Something like `0 means default and it is infinite for  normal setup and 120 seconds for failover`

internal/cluster/connection_manager.go
sancar: We come to the conclusion that we can not have two different constructors to achieve simplicity. 
Eventough it looks more complex to me comparing to what we have java , I will not insist on removing circuit breaker because of `a single way of having retry logic` argument.

sancar (approved):

PR #617, opened 2021-09-03
Fixed ConnectionManager member removed handler [API-693]

### LINKED ISSUE #600: Client tries to connect non-existent members

yuce:
* Fixes #600
* Send member list view request is sent to all members, not just the first seed.
* Various connection manager fixes

Tested with:
* https://github.com/SeriyBg/hazelcast-enterprise-operator/tree/hazelcast-status
* https://github.com/hazelcast-guides/kubernetes

### DIFF
--- a/client.go
+++ b/client.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/hazelcast/hazelcast-go-client/cluster"
 	"github.com/hazelcast/hazelcast-go-client/hzerrors"
+	"github.com/hazelcast/hazelcast-go-client/internal"
 	"github.com/hazelcast/hazelcast-go-client/internal/cloud"
 	icluster "github.com/hazelcast/hazelcast-go-client/internal/cluster"
 	"github.com/hazelcast/hazelcast-go-client/internal/event"
@@ -37,6 +38,10 @@ import (
 	"github.com/hazelcast/hazelcast-go-client/internal/serialization"
 	"github.com/hazelcast/hazelcast-go-client/internal/stats"
 	"github.com/hazelcast/hazelcast-go-client/types"
+)
+
+const (
+	ClientVersion = internal.ClientVersion
 )
 
 var nextId int32
--- a/client.go
+++ b/client.go
@@ -459,7 +459,7 @@ func (c *Client) createComponents(config *Config) {
 		FailoverConfig:       &config.Failover,
 		Labels:               config.Labels,
 	})
-	viewListener := icluster.NewViewListenerService(clusterService, connectionManager, c.eventDispatcher)
+	viewListener := icluster.NewViewListenerService(clusterService, connectionManager, c.eventDispatcher, c.logger)
 	invocationHandler := icluster.NewConnectionInvocationHandler(icluster.ConnectionInvocationHandlerCreationBundle{
 		ConnectionManager: connectionManager,
 		ClusterService:    clusterService,
--- a/client.go
+++ b/client.go
@@ -80,6 +80,7 @@ type Client struct {
 	connectionManager       *icluster.ConnectionManager
 	clusterService          *icluster.Service
 	partitionService        *icluster.PartitionService
+	viewListenerService     *icluster.ViewListenerService
 	invocationService       *invocation.Service
 	serializationService    *serialization.Service
 	eventDispatcher         *event.DispatchService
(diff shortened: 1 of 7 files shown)


client.go
```diff
@@ -39,6 +40,10 @@ import (
 	"github.com/hazelcast/hazelcast-go-client/types"
 )
 
+const (
+	ClientVersion = internal.ClientVersion
```
sancar, line 44: ide says unused? Is this a leftover?

yuce: No, it's not a leftover. I'll add a fake test so it doesn't cause any linter warnings.

internal/cluster/connection_manager.go
```diff
@@ -60,6 +60,10 @@ const (
 	serializationVersion = 1
 )
 
+const (
+	clusterConnectionFixTimeout = 10 * time.Second
```
sancar, line 64: what is this needed for ?  We already have connection timeout in NetworkConfig and also heartbeat timeout for this area? Why do we need another timeout ?

internal/cluster/connection_manager.go
```diff
@@ -324,30 +328,39 @@ func (m *ConnectionManager) handleInitialMembersAdded(e event.Event) {
...
 		return
 	}
 	if e, ok := event.(*MembersAdded); ok {
+		m.logger.Trace(func() string {
+			return fmt.Sprintf("connectionManager.handleMembersAdded: %v", e.Members)
+		})
 		missingAddrs := m.connMap.FindAddedAddrs(e.Members, m.clusterService)
 		for _, addr := range missingAddrs {
-			if _, err := m.ensureConnection(context.TODO(), addr); err != nil {
+			if _, err := connectMember(context.TODO(), m, addr); err != nil {
```
sancar, line 341: As far as I understand this also registers a membership listener to the connected member. 
We should not listen from all members. Single member delivers them already. In the case of split brain scenerious, we should follow single members view. 
Same goes for here. 
https://github.com/hazelcast/hazelcast-go-client/pull/617/files#diff-8235afecd77026dd0bddf3a2cd47ca4843345b3ddec987b0a5d45c127d14fadcR686

sancar:
@yuce The issue seems testable to me. Can we write a unittest for this ?

internal/cluster/connection_manager.go
yuce: That limits the connection time for the context created in `checkFixConnection`.That's to prevent a very slow connection to

client_it_test.go
```diff
@@ -553,3 +558,61 @@ func TestClientFailover_EECluster_Reconnection(t *testing.T) {
...
+	if !ok {
+		log.Fatalf("could not terminate member: %s", err.Error())
+	}
+	m, err := cls.RC.StartMember(ctx, cls.ClusterID)
+	if err != nil {
+		log.Fatal(err)
+	}
+	highlight(t, "Started member: %s", m.UUID)
+	time.Sleep(30 * time.Second)
+	assert.Equal(t, int64(memberCount+1), atomic.LoadInt64(&addedCount))
```
sancar, line 612: Can we implement something like assertTrueEventually that we have on the java to avoid sleep statements in the tests. https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/test/HazelcastTestSupport.java#L1251

yuce: We can something like timed waitgroups. I've created a ticket for that: https://github.com/hazelcast/hazelcast-go-client/issues/621

internal/cluster/connection_manager.go
yuce: Removed that bit.

internal/cluster/connection_manager.go
yuce: Fixed.

sancar (approved):

PR #619, opened 2021-09-07
[1.1.1] Add examples [API-683]

martijndeheus:
Added some additional tests to `examples` folder and added tests to `org-website` folder to present on the website.

I had to exclude the `org-website` folder from the build/test scripts to prevent errors caused by that.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -59,6 +59,9 @@ func Example() {
 func ExampleSet() {
 	ctx := context.TODO()
 	client, err := hazelcast.StartNewClient(ctx)
+	if err != nil {
+		log.Fatal(err)
+	}
 	// Retrieve the set named my-set.
 	set, err := client.GetSet(ctx, "my-set")
 	if err != nil {
@@ -81,6 +84,9 @@ func ExampleSet() {
 func ExamplePNCounter() {
 	ctx := context.TODO()
 	client, err := hazelcast.StartNewClient(ctx)
+	if err != nil {
+		log.Fatal(err)
+	}
 	// Retrieve the PN counter named my-pn.
 	pn, err := client.GetPNCounter(ctx, "my-pn")
 	if err != nil {
@@ -122,6 +128,9 @@ func ExampleMap() {
 func ExampleMap_Aggregate() {
 	ctx := context.TODO()
 	client, err := hazelcast.StartNewClient(ctx)
+	if err != nil {
+		log.Fatal(err)
+	}
 	myMap, err := client.GetMap(ctx, "my-map")
 	if err != nil {
 		log.Fatal(err)
@@ -143,10 +152,16 @@ func ExampleMap_AddEntryListener() {
 	// error handling was omitted for brevity
 	ctx := context.TODO()
 	client, err := hazelcast.StartNewClient(ctx)
+	if err != nil {
+		log.Fatal(err)
+	}
 	entryListenerConfig := hazelcast.MapEntryListenerConfig{
 		IncludeValue: true,
 	}
-	m, _ := client.GetMap(ctx, "somemap")
+	m, err := client.GetMap(ctx, "somemap")
+	if err != nil {
+		log.Fatal(err)
+	}
 	// enable receiving entry added events
 	entryListenerConfig.NotifyEntryAdded(true)
 	// enable receiving entry removed events
@@ -177,22 +192,22 @@ func ExampleMap_AddEntryListener() {
 		}
 	})
 	if err != nil {
-		panic(err)
+		log.Fatal(err)
 	}
 	// performing modifications on the map entries
 	key := strconv.Itoa(int(time.Now().Unix()))
 	if err := m.Set(ctx, key, "1"); err != nil {
-		panic(err)
+		log.Fatal(err)
 	}
 	if err := m.Set(ctx, key, "2"); err != nil {
-		panic(err)
+		log.Fatal(err)
 	}
 	if err := m.Delete(ctx, key); err != nil {
-		panic(err)
+		log.Fatal(err)
 	}
 	// you can use the subscriptionID later to remove the event listener.
 	if err := m.RemoveEntryListener(ctx, subscriptionID); err != nil {
-		panic(err)
+		log.Fatal(err)
 	}
 }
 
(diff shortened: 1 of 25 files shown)


examples/serialization/portable/main.go
```diff
@@ -99,10 +105,13 @@ func main() {
...
-	classDefinition.AddStringField("name")
-	classDefinition.AddInt32Field("age")
-	config.Serialization.SetClassDefinitions(classDefinition)
+	employeeClassDefinition := serialization.NewClassDefinition(factoryID, employeeClassID, 1)
+	employeeClassDefinition.AddStringField("name")
+	employeeClassDefinition.AddInt32Field("age")
+	config.Serialization.SetClassDefinitions(employeeClassDefinition)
+	baseObjectClassDefinition := serialization.NewClassDefinition(factoryID, baseObjectClassID, 1)
+	baseObjectClassDefinition.AddPortableField("employee", employeeClassDefinition)
+	config.Serialization.SetClassDefinitions(baseObjectClassDefinition)
```
yuce, line 114: The same comments of the previous sample.

examples/set/main.go
```diff
@@ -0,0 +1,58 @@
...
+	// Get a random set
+	rand.Seed(time.Now().Unix())
+	setName := fmt.Sprintf("sample-%d", rand.Int())
+	set, err := client.GetSet(ctx, setName)
+	if err != nil {
+		log.Fatal(err)
+	}
+	// Add items to set
+	for i := 0; i < setItems; i++ {
+		set.Add(ctx, fmt.Sprintf("Item %d", i))
```
yuce, line 46: Error is ignored.

examples/topic/main.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	// Get a random topic
+	rand.Seed(time.Now().Unix())
+	topicName := fmt.Sprintf("sample-%d", rand.Int())
+	topic, err := client.GetTopic(ctx, topicName)
+	if err != nil {
+		log.Fatal(err)
+	}
+	// Add a message listener to the topic
+	topic.AddMessageListener(ctx, messageListener)
```
yuce, line 50: Error is ignored.

examples/topic/main.go
```diff
@@ -0,0 +1,57 @@
...
+	// Publish messages to topic
+	for i := 0; i < messageCount; i++ {
+		topic.Publish(ctx, fmt.Sprintf("Message %d", i))
```
yuce, line 53: Error is ignored.

serialization/example_custom_serialization_test.go
```diff
@@ -0,0 +1,62 @@
...
+		log.Fatal(err)
+	}
+	// Get a random map
+	m, err := client.GetMap(ctx, "map-1")
+	if err != nil {
+		log.Fatal(err)
+	}
+	// Store an object in the map
+	emp := Employee{"Doe"}
+	m.Put(ctx, "employee-1", emp)
```
yuce, line 53: Error is ignored.

yuce: Does it make sense to remove `emp` variable and move `Employe{ ... }` to `Put`?

serialization/example_portable_serialization_test.go
```diff
@@ -0,0 +1,130 @@
...
+	config := hazelcast.NewConfig()
+	config.Serialization.PortableVersion = 1
+	config.Serialization.SetPortableFactories(&MyPortableFactory{})
+	employeeClassDefinition := serialization.NewClassDefinition(factoryID, employeeClassID, 1)
+	employeeClassDefinition.AddStringField("name")
+	employeeClassDefinition.AddInt32Field("age")
+	config.Serialization.SetClassDefinitions(employeeClassDefinition)
+	baseObjectClassDefinition := serialization.NewClassDefinition(factoryID, baseObjectClassID, 1)
+	baseObjectClassDefinition.AddPortableField("employee", employeeClassDefinition)
+	config.Serialization.SetClassDefinitions(baseObjectClassDefinition)
```
yuce, line 94: The same comment about the previous samples.

yuce (changes requested):
Thanks for this PR, it looks great! Except, I think there are a few things that could be improved. Especially, some of the errors are being ignored.

example_test.go
```diff
@@ -62,7 +58,8 @@ func Example() {
 
 func ExampleSet() {
 	ctx := context.TODO()
-	// Retrieve the set named my-set
+	client, err := hazelcast.StartNewClient(ctx)
```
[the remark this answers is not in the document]
martijndeheus: These examples will be put in the godocs. I left the error handling out here as it doesn't concern the main topic of the example and might clutter the godoc. Do you prefer I handle the error or leave a `_`?

examples/map/main.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	fmt.Println(value)
+	// Atomic putIfAbsent method
+	val, err := m.PutIfAbsent(ctx, "key-2", "value-2")
+	if err != nil {
+		log.Fatal(err)
+	}
+	if nilutil.IsNil(val) {
+		fmt.Println("Inserted value")
+	}
```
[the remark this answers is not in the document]
martijndeheus: I agree. I didn't notice this was an internal package.

example_test.go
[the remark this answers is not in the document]
yuce: I think the best thing to do is just handling the error, so the user knows a function returns an error value. If you think that makes the examples too verbose, I would add a warning comment like: "error handling is omitted for brevity".

examples/org-website/custom_serializer_sample.go
```diff
@@ -0,0 +1,66 @@
...
+	employee, ok := object.(*Employee)
+	if !ok {
+		panic("can serialize only Employee")
+	}
+	output.WriteString(employee.Surname)
+}
+
+func main() {
+	// Configure serializer
+	config := hazelcast.NewConfig()
```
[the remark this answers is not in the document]
martijndeheus: Just out of curiousity, why do we have `NewConfig()` if we recommend initializing the struct directly?

example_test.go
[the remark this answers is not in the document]
martijndeheus: Alright! I will handle the error for these cases and add the comment for examples where we have to perform a bunch of operations on a data structure. Thanks!

examples/org-website/custom_serializer_sample.go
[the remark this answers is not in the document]
yuce: I've left `NewConfig` in v1 since in the future we may have similar specialized config creation functions, like `NewCloudConfig("my-cluster", "my-token")`.

martijndeheus:
Thanks a lot for the thorough review. I addressed your comments.

yuce (approved):
Thanks, it looks great!

Issue #621, opened 2021-09-08
Refactor integration tests to minimize sleep


yuce:
Sleeps in tests increases the running time and decreases the reliability of tests. Consider refactoring integrations tests to minimize sleep statements.

sancar:
There are sleep usages on both tests and in the production code. These will cause real headaches if we don't remove them soon. 
Time is not something we should rely on in a concurrent/distributed system.

yuce:
There are only 2 places where a sleep is used (apart from tests):
1. In order to give a chance to shutdown event to be dispatched at client shutdown: https://github.com/hazelcast/hazelcast-go-client/blob/af1dfa73712c08d0fa3348e67a9028ab8bc2f5ce/client.go#L261
2. Wait until the next attempt in the retry logic: https://github.com/hazelcast/hazelcast-go-client/blob/af1dfa73712c08d0fa3348e67a9028ab8bc2f5ce/internal/cb/circuitbreaker.go#L120

For 2, the sleep is required. For 1, we can use something other than a sleep, but that would complicate the code. I think these sleep usages have a low probability of affecting the production code.

I agree that we should use less sleeps in the tests, if we can't remove them all.

sancar:
The first one needs to be fixed. It does not guarantee that the event is dispatched, and it should be guaranteed.
I think we already discussed it here. 
https://github.com/hazelcast/hazelcast-go-client/pull/552#discussion_r622259029

PR #644, opened 2021-09-17
Uses a mutex instead of channels in dispatch service [API-892, API-888]

yuce:
We select on two channels in the dispatch service, one for events (publish) and one for control (subscribe, unsubscribe). In some cases that causes a race condition, since the order of subsribe/unsubscribe and publish is not deterministic. That caused occasional test failure of https://github.com/hazelcast/hazelcast-go-client/blob/5017195b7f03afba8fcfd9d8955694e8b3a172e8/internal/event/dispatch_service_test.go#L58

Fixed that by using a mutex instead of two channels.

Fixes https://github.com/hazelcast/hazelcast-go-client/issues/621

### DIFF
--- a/internal/event/dispatch_service.go
+++ b/internal/event/dispatch_service.go
@@ -24,37 +24,50 @@ import (
 	"github.com/hazelcast/hazelcast-go-client/internal/logger"
 )
 
+// DefaultSubscriptionID enables automatically setting a subscription ID.
 const DefaultSubscriptionID = -1
 
+// Event is a value published by an event publishers and delivered to one or more subscribbers via the dispatch service.
 type Event interface {
 	EventName() string
 }
 
-type Handler func(event Event)
+// SyncHandler is the type for synchronous callbacks.
+type SyncHandler func(event Event)
+
+// AsyncHandler is the type for asynchronous callbacks.
+type AsyncHandler func(event Event, wg *sync.WaitGroup)
 
 const (
 	ready   = 0
 	stopped = 1
 )
 
+type subscriptionMap map[string]*sync.Map
+
+// DispatchService delivers events from a publisher to one or more subscribers.
 type DispatchService struct {
-	logger            logger.Logger
-	syncSubscriptions map[string]map[int64]Handler
-	subscriptions     map[string]map[int64]Handler
-	eventMu           *sync.RWMutex
-	state             int32
+	logger             logger.Logger
+	asyncSubscriptions atomic.Value
+	syncSubscriptions  atomic.Value
+	subMu              *sync.Mutex
+	state              int32
 }
 
+// NewDispatchService creates a dispatch service with the given logger.
 func NewDispatchService(logger logger.Logger) *DispatchService {
 	service := &DispatchService{
-		subscriptions:     map[string]map[int64]Handler{},
-		syncSubscriptions: map[string]map[int64]Handler{},
-		eventMu:           &sync.RWMutex{},
-		logger:            logger,
+		subMu:  &sync.Mutex{},
+		logger: logger,
 	}
+	service.asyncSubscriptions.Store(subscriptionMap{})
+	service.syncSubscriptions.Store(subscriptionMap{})
 	return service
 }
 
+// Stop terminates the dispatch service and releases the resources.
+// It is not possible to subscribe to or publish events once the dispatch service is terminated.
+// Calling Stop more than once is no-op.
 func (s *DispatchService) Stop() {
 	// stopping a not-running service is no-op
 	if !atomic.CompareAndSwapInt32(&s.state, ready, stopped) {
@@ -62,10 +75,12 @@ func (s *DispatchService) Stop() {
 	}
 }
 
-// Subscribe attaches handler to listen for events with eventName.
-// Do not rely on the order of handlers, they may be shuffled.
-func (s *DispatchService) Subscribe(eventName string, subscriptionID int64, handler Handler) {
-	// subscribing to a not-runnning service is no-op
+// Subscribe attaches a handler to listen for events with the given event name.
+// Do not rely on the subscription order of handlers, they may be called back in a random order.
+// The handler will be called in a goroutine, and the dispatch service makes sure that the corresponding goroutine is started before the goroutine for the next handler.
+// Once the goroutines for handlers start, it's up to the runtime to schedule them.
+func (s *DispatchService) Subscribe(eventName string, subscriptionID int64, handler SyncHandler) {
+	// subscribing to a not-running service is no-op
 	if atomic.LoadInt32(&s.state) != ready {
 		return
 	}
@@ -75,15 +90,18 @@ func (s *DispatchService) Subscribe(eventName string, subscriptionID int64, hand
 	s.logger.Trace(func() string {
 		return fmt.Sprintf("event.DispatchService.Subscribe: %s, %d, %v", eventName, subscriptionID, handler)
 	})
-	s.eventMu.Lock()
-	s.subscribe(eventName, subscriptionID, handler)
-	s.eventMu.Unlock()
+	s.subscribeAsync(eventName, subscriptionID, func(event Event, wg *sync.WaitGroup) {
+		wg.Done()
+		handler(event)
+	})
 }
 
-// SubscribeSync attaches handler to listen for events with eventName.
+// SubscribeSync attaches handler to listen for events with the given event name.
 // Sync handlers are dispatched first, the events are ordered.
-// Do not rely on the order of handlers, they may be shuffled.
-func (s *DispatchService) SubscribeSync(eventName string, subscriptionID int64, handler Handler) {
+// Do not rely on the subscription order of handlers, they may be called back in a random order.
+// The handler is called in the same goroutine with the dispatch service and executes until the end of the corresponding function.
+// Make sure sync handlers do no block.
+func (s *DispatchService) SubscribeSync(eventName string, subscriptionID int64, handler SyncHandler) {
 	// subscribing to a not-runnning service is no-op
 	if atomic.LoadInt32(&s.state) != ready {
 		return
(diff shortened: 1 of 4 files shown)


internal/event/dispatch_service.go
```diff
@@ -31,50 +32,28 @@ type Event interface {
...
-type controlType int
-
-const (
-	subscribe controlType = iota
-	subscribeSync
-	unsubscribe
-)
-
 const (
 	created int32 = iota
```
sancar, line 36: Do we really need 3 states ? I think we can eliminate one of them. `created` and `ready` seems like same thing to me.

yuce: Removed at: https://github.com/hazelcast/hazelcast-go-client/pull/644/commits/77bff36b7a0d762a4af5b25150d7d694c9d8028f

internal/event/dispatch_service.go
```diff
@@ -100,12 +78,9 @@ func (s *DispatchService) Subscribe(eventName string, subscriptionID int64, hand
 	s.logger.Trace(func() string {
 		return fmt.Sprintf("event.DispatchService.Subscribe: %s, %d, %v", eventName, subscriptionID, handler)
 	})
```
sancar, line 77: Should our Subscribe guarantee the order of events ? 
The doc says `Do not rely on the order of handlers, they may be shuffled.` . Do we mean order of `events` or `handlers` ?
Can we make the behaviour of events more clear on the doc as well ?

internal/event/dispatch_service.go
```diff
@@ -84,7 +63,6 @@ func (s *DispatchService) Stop() {
 	if !atomic.CompareAndSwapInt32(&s.state, ready, stopped) {
 		return
 	}
-	close(s.doneCh)
```
sancar, line 87: I think the stop method of a dispath service should guarantee that all events are handled before stopping.

internal/event/dispatch_service.go
yuce: That's really the order of handlers. If the user registers handlerA then handlerB to an event, then we don't guarantee that handlerA will be dispatched before handlerB.

Events should be received in order, and if the user subscribed using the sync variant, then they should observe it in the same order. I'll add a test for that.

sancar: I was asking for Subscribe method not SubscribeSync.
Should the order be guaranteed for `Subscribe` ?
And is the order guaranteed for `Subscribe` ?

yuce: The test is here: https://github.com/hazelcast/hazelcast-go-client/pull/644/commits/2e74a09a6915ca8100eb663f70d081022789e8fa

yuce: We can't guarantee that easily for Subscribe, since the corresponding handlers are dispatched in a goroutine.

internal/event/dispatch_service.go
yuce: For sync events that's guaranteed. For async events, that would require an extra mechanism, like passing a channel to the handler which would be very error-prone to use.

internal/event/dispatch_service.go
sancar: Is not guaranteeing the order of `Subscribe` will cause any correctness problems where it is used ?

internal/event/dispatch_service.go
sancar: When we don't guarantee that all events are consumed, would that cause any correctness problems where it is used ?

internal/event/dispatch_service.go
yuce: I think we should be using Sync events where the order of events may cause problems.

sancar: I see that almost all usages are problematic where open/closed added/removed listened via same service. 
https://github.com/hazelcast/hazelcast-go-client/blob/master/internal/cluster/connection_listener_binder.go#L73
https://github.com/hazelcast/hazelcast-go-client/blob/master/internal/cluster/connection_manager.go#L317
https://github.com/hazelcast/hazelcast-go-client/blob/master/internal/cluster/view_listener_service.go#L27

What do you think ? I did a quick scan over the code. I will let you to check the rest.

sancar: Also note that if you use SubsribeSync in internals, it means that we should be make sure that design is deadlock free.

sancar: I was thinking how to solve this issue. I can say that making sure the design is deadlock free is hard from experience.
I think if we design the dispatch service in only async way where we guarantee the order, it will be a lot easier to implement the rest. 

What do you think ?

yuce: Can you explain a bit about what you mean by guaranteeing the order? Should an event handler be dispatched in the order of the "Publish" calls and run to completion before the other event handler starts? Or just starting an event handler (but not waiting for its completion) before the next event handler is adequate?

Sync handlers are required only in a few places and they are just used to publish an event on the user event dispatcher. In another PR, that probably will mean writing to a channel, so that won't block.

internal/event/dispatch_service.go
yuce: In the current design, events are not buffered, so we can assume they are consumed. Even if they were buffered, `Stop` of the dispatcher would only be called during client shutdown, so it should be safe to assume that the remaining events are not useful (and probably shouldn't even be delivered to the user).

internal/event/dispatch_service.go
sancar: > Should an event handler be dispatched in the order of the "Publish" calls and run to completion before the other event handler starts? Or just starting an event handler (but not waiting for its completion) before the next event handler is adequate?

The first one. The second one does not add any value. In the eyes of the caller of the service, the two are still concurrent. 


Lets say that I do two Publishes where there is an order between them, lets say that from the single thread to make it simple. It could be a lot more complicated(like two events coming from remote where the order is guaranteed by the rest of the system). 

service.Publish(MemberConnected(address))
service.Publish(MemberDisconnected(address))

//lets say these two are subscribed to dispatchService 
func connectedHandler(event) {
    //TAG1
    //.... Users code
}

func disconnectedHandler(event) {
    //TAG2
    //.... Users code
}

Lets say that we are doing the second one and `not waiting for its completion`
1. The connected handler is called, we made sure it is started. That means `disconnectedHandler` can be called via dispathService . 
2. At the beginning of `connectedHandler` , at TAG1, we hit a garbage collection and go decided to park our thread. 
3. dispathService calls disconnectedHandler. 
4. disconnectedHandler is finished. 
5. Go decided that now it is ok to let `connectedHandler` to continue from TAG1. 

As you can see `disconnectedHandler` worked before `connectedHandler`

internal/event/dispatch_service.go
sancar: We wait for events to be handled in the java, otherwise we consider it as a resource leak. 
If you open and close a client fast enough, the system can go out of memory. 

Imagine a user I puts a long running task on the Shutdown method . Opens a client, put an item, and shutdown the client per request. And it has very high throughput. 
Result: The system does not slows down as it should be but it goes out of memory eventually crashing the whole service.

sancar: https://github.com/hazelcast/hazelcast-go-client/pull/644#discussion_r711845847

I am confused about this comment of yours. Let me ask the following. 

When hazelcastClient.shutdown() finished. 

Is it guaranteed that handling LifecycleServiceShutdown is started ? And also, 
is it guaranteed that handling LifecycleServiceShutdown is finished ?

sancar: Ok. If for the async subscribe if it is not guaranteed, lets put it on the Subscribe doc. Calling this method does not guarantee that this event will be delivered to handlers when dispatch service is stopped from another thread.

internal/event/dispatch_service.go
yuce: If we wait the end of a handler to completion before launching another handler, then it's not asynchronous. I don't think there's a solution that we can apply to all cases. But for this specific issue, how about making sure that the corresponding connection in disconnected handler is connected before going with disconnected logic? Maybe having a separate event which is fired only after `ConnectionOpened` is completed?

internal/event/dispatch_service.go
yuce: In the current design (this PR), we don't use channels, and the corresponding handlers are guaranteed to be started before starting the next handler and once started, handlers run to completion. I would say events are guaranteed to be delivered to handlers, even if the dispatch service is stopped.

internal/event/dispatch_service.go
sancar: > If we wait the end of a handler to completion before launching another handler, then it's not asynchronous
We can still make it async in the sense that caller of publish will not be blocked. I think that is enough right ?

>  how about making sure that the corresponding connection in disconnected handler is connected before going with disconnected logic? Maybe having a separate event which is fired only after ConnectionOpened is completed?

We can always have more complex solutions. But I believe there is a much simpler solution.
We still focus on a simple dispatch service where 
1- It is async for the caller of Publish. Meaning Publish will not be blocked. 
2- Events fired from the same thread will be handled in the same order. One will finish, then other will start.  (it is enough to make sure this per EventName)
3 - If we block an event, it should not block all events only ones that are related to same EventName. 
4 - [OPTIONALLY] a close after publish in the same thread waits for published item to be handled(finished) . 

I think implementing a dispatch service guaranteeing the at least first three should be simple. I think the forth item will come free after that but lets wait and see. 
And it looks like what all we want. 
What do you think ?

internal/event/dispatch_service.go
sancar: Firstly I need to correct my wording on the https://github.com/hazelcast/hazelcast-go-client/pull/644#discussion_r711962431

I wrote :
Calling this method does not guarantee that this event will be delivered to handlers when dispatch service is stopped from **another** thread.
This is true because a second thread can set `state` to stopped and the callers events may not be fired. Right ? 
 And callers of the Publish is not aware of it. So they need to know at least from doc.
 Or we can return a boolean to denote if the publish will be handled eventually or not. 

But actually, I meant to say :
Calling this method does not guarantee that this event will be **handled(and finished running)** when dispatch service is stopped from **same** goroutine. Meaning: 
```
s.Publish(event)
s.Close()
```
When close is finished, we don't know that the handler is finished handling the event. This is not guaranteed. 
We can document this as well clearly on the Subscribe.

internal/event/dispatch_service.go
yuce: > Events fired from the same thread will be handled in the same order

Currently events may be published from different goroutines, E.g,. ConnectionOpened is published from connection manager, ConnectionClosed is published from socket read/write loops and the connection manager.

I think the problem is not with the goroutine that publishes the events but with the subscribers. E.g., we want to make sure that ConnectionClosed handler for a connection starts iff the corresponding ConnectionOpened handler ran to completion.  How about keeping a set of connections for which the ConnectionOpened handler ran to completion? Then ConnectionClosed handlers would block until the corresponding connection is in that set.

internal/event/dispatch_service.go
yuce: In the current design, stopping the dispatch service only disables publishing/subscribing/unsubscribing an event. Once an event is published, all handlers for that event run in a goroutine, even if the dispatch service is stopped, and stopping the dispatch service does not "kill" those goroutines. If a handler goroutine is running when the client is shutdown, it will continue to run, until the process itself is killed.

internal/event/dispatch_service.go
sancar: I does not matter whether ConnectionOpened and ConnectionClosed is published from different threads. What matters is that is there a happens-before relation between them or not ? I believe there is. 

Again. We can solve a simple problem once in the dispatch service, or we can solve this every place again and again that uses this dispatch service. I prefer solving it once. 

As I said here https://github.com/hazelcast/hazelcast-go-client/pull/644#discussion_r713845921
Do you think we can find this simple solution , or am I asking something impossible ?
 Or would it be ok if I try to send a pr for it ?

internal/event/dispatch_service.go
sancar: Yes, I got how it works. I am just saying documentation of the method can state this so that users of it will not make mistake.

internal/event/dispatch_service.go
yuce: We both agree that ConnectionClosed handler should be executed after ConnectionOpened. Of course, please send a PR if there is an easier way of accomplishing that than I described.

sancar: Here it is 
https://github.com/hazelcast/hazelcast-go-client/pull/652

yuce:
Closing, in favor of: https://github.com/hazelcast/hazelcast-go-client/pull/652

PR #648, opened 2021-09-20
Add missing tests [API-886]

### LINKED ISSUE #635: Client has untested Publıc API on MapProxy
Some are listed here as TODO.
https://github.com/hazelcast/hazelcast-go-client/blob/master/map_it_test.go#L540

I have seen more when I look at the code coverage. These should be tested

yuce:
- Fixes #635
- Fixes PNCounter, so it requests max replica count and uses that to select the target replica. Also PNCounter now excludes problematic replicas.

### DIFF
--- a/internal/cluster/cluster_service.go
+++ b/internal/cluster/cluster_service.go
@@ -19,10 +19,8 @@ package cluster
 import (
 	"context"
 	"fmt"
-	"math/rand"
 	"strings"
 	"sync"
-	"time"
 
 	pubcluster "github.com/hazelcast/hazelcast-go-client/cluster"
 	"github.com/hazelcast/hazelcast-go-client/internal/event"
@@ -32,8 +30,6 @@ import (
 	"github.com/hazelcast/hazelcast-go-client/internal/proto/codec"
 	"github.com/hazelcast/hazelcast-go-client/types"
 )
-
-var commonRand = rand.New(rand.NewSource(time.Now().UnixNano()))
 
 type Service struct {
 	logger            ilogger.Logger
@@ -99,15 +95,8 @@ func (s *Service) MemberAddrs() []pubcluster.Address {
 	return s.membersMap.MemberAddrs()
 }
 
-func (s *Service) RandomReplica(n int) (pubcluster.MemberInfo, bool) {
-	return s.membersMap.RandomReplica(n, nil)
-}
-
-func (s *Service) RandomReplicaExcluding(n int, excluded map[types.UUID]struct{}) (pubcluster.MemberInfo, bool) {
-	return s.membersMap.RandomReplica(n, func(mem *pubcluster.MemberInfo) bool {
-		_, found := excluded[mem.UUID]
-		return !found
-	})
+func (s *Service) OrderedMembers() []pubcluster.MemberInfo {
+	return s.membersMap.OrderedMembers()
 }
 
 func (s *Service) RefreshedSeedAddrs(clusterCtx *CandidateCluster) ([]pubcluster.Address, error) {
@@ -255,13 +244,12 @@ func (m *membersMap) MemberAddrs() []pubcluster.Address {
 	return addrs
 }
 
-// RandomReplica returns one of the replicas (first n data members).
-// filter, if provided should return true for considering the data member in the list of replicas.
-// Returns false if no suitable data member was found.
-func (m *membersMap) RandomReplica(n int, filter func(mem *pubcluster.MemberInfo) bool) (pubcluster.MemberInfo, bool) {
-	m.membersMu.RLock()
-	defer m.membersMu.RUnlock()
-	return randomReplica(m.orderedMembers, n, filter)
+func (m *membersMap) OrderedMembers() []pubcluster.MemberInfo {
+	m.membersMu.Lock()
+	members := make([]pubcluster.MemberInfo, len(m.orderedMembers))
+	copy(members, m.orderedMembers)
+	m.membersMu.Unlock()
+	return members
 }
 
 // addMember adds the given memberinfo if it doesn't already exist and returns true in that case.
@@ -302,29 +290,6 @@ func (m *membersMap) reset() {
 	m.addrToMemberUUID = map[pubcluster.Address]types.UUID{}
 	m.version = -1
 	m.membersMu.Unlock()
-}
-
-func randomReplica(members []pubcluster.MemberInfo, n int, filter func(mem *pubcluster.MemberInfo) bool) (pubcluster.MemberInfo, bool) {
-	if n > len(members) {
-		n = len(members)
-	}
-	if n == 0 {
-		return pubcluster.MemberInfo{}, false
-	}
-	// scans first n members, starting from idx, wrapping at n
-	idx := commonRand.Intn(n)
-	for i, mem := range members {
-		if mem.LiteMember {
-			continue
-		}
-		if i < idx {
-			continue
-		}
-		if filter == nil || filter(&mem) {
-			return mem, true
-		}
-	}
-	return pubcluster.MemberInfo{}, false
 }
 
 func (m *membersMap) logMembers(version int32) {
--- a/internal/cluster/cluster_service.go
+++ b/internal/cluster/cluster_service.go
@@ -19,7 +19,9 @@ package cluster
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"sync"
+	"time"
 
 	pubcluster "github.com/hazelcast/hazelcast-go-client/cluster"
 	"github.com/hazelcast/hazelcast-go-client/internal/event"
@@ -29,6 +31,8 @@ import (
 	"github.com/hazelcast/hazelcast-go-client/internal/proto/codec"
 	"github.com/hazelcast/hazelcast-go-client/types"
 )
+
+var commonRand = rand.New(rand.NewSource(time.Now().UnixNano()))
 
 type Service struct {
 	logger            ilogger.Logger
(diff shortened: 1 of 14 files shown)


map_it_test.go
```diff
@@ -1071,6 +1063,145 @@ func TestMap_ExecuteOnEntries(t *testing.T) {
...
+	it.MapTester(t, func(t *testing.T, m *hz.Map) {
+		ctx := context.Background()
+		prefix := "My Prefix"
+		id, err := m.AddInterceptor(ctx, &MapGetInterceptor{Prefix: prefix})
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer func() {
+			if _, err := m.RemoveInterceptor(ctx, id); err != nil {
+				panic(err)
```
sancar, line 1076: Why do we panic instead of t.Fatal() ?

yuce: It wouldn't matter, but switched to `t.Fatal` to be more consistent.

pn_counter_it_test.go
```diff
@@ -157,6 +157,12 @@ func TestPNCounter_SubtractAndGet(t *testing.T) {
 	})
 }
 
+func TestPNCounter_Reset(t *testing.T) {
+	it.PNCounterTester(t, func(t *testing.T, pn *hz.PNCounter) {
+		pn.Reset()
```
sancar, line 162: Can we verify that reset is worked ? 
Increment the counter first. 
Close the instance to make sure that consistency is lost. 
Try an increment operation where an error should be returned.  `ConsistencyLostException` 
Then Reset the PnCounter.
Verify that increment is working without the exception.

map_it_test.go
```diff
@@ -1071,6 +1063,145 @@ func TestMap_ExecuteOnEntries(t *testing.T) {
...
+			}
+		}()
+		v, err := m.Get(ctx, "foo")
+		if err != nil {
+			t.Fatal(err)
+		}
+		assert.Equal(t, prefix, v)
+		ok, err := m.RemoveInterceptor(ctx, id)
+		if err != nil {
+			log.Fatal(err)
```
sancar, line 1086: Is there a reason that we use log.Fatal instead of t.Fatal

map_it_test.go
```diff
@@ -1071,6 +1063,145 @@ func TestMap_ExecuteOnEntries(t *testing.T) {
...
+	it.MapTester(t, func(t *testing.T, m *hz.Map) {
+		ctx1 := m.NewLockContext(context.Background())
+		if err := m.Lock(ctx1, "foo"); err != nil {
+			t.Fatal(err)
+		}
+		// unlock 5 seconds later
+		go func() {
+			time.Sleep(5 * time.Second)
+			if err := m.Unlock(ctx1, "foo"); err != nil {
+				panic(err)
```
sancar, line 1129: t.Fatal instead of panic ?

map_it_test.go
```diff
@@ -1071,6 +1063,145 @@ func TestMap_ExecuteOnEntries(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		if err := m.Lock(ctx1, "foo"); err != nil {
+			t.Fatal(err)
+		}
+		// unlock 5 seconds later
+		go func() {
+			time.Sleep(5 * time.Second)
+			if err := m.Unlock(ctx1, "foo"); err != nil {
+				panic(err)
```
sancar, line 1186: https://github.com/hazelcast/hazelcast-go-client/pull/648/files#r714638636

map_it_test.go
```diff
@@ -1071,6 +1063,145 @@ func TestMap_ExecuteOnEntries(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		assert.Equal(t, nil, v)
+	})
+}
+
+func TestMap_TryPut(t *testing.T) {
+	it.MapTester(t, func(t *testing.T, m *hz.Map) {
+		ctx1 := m.NewLockContext(context.Background())
+		if err := m.Lock(ctx1, "foo"); err != nil {
```
sancar, line 1100: I think we forget to Unlock in this test ?

map_it_test.go
```diff
@@ -1071,6 +1063,145 @@ func TestMap_ExecuteOnEntries(t *testing.T) {
...
+		}()
+		// TryPut with a different lock context returns false
+		ctx2 := m.NewLockContext(context.Background())
+		ok, err := m.TryPutWithTimeout(ctx2, "foo", "bar", 1*time.Second)
+		if err != nil {
+			t.Fatal(err)
+		}
+		assert.Equal(t, false, ok)
+		// TryPut after the timeout
+		ok, err = m.TryPutWithTimeout(ctx2, "foo", "bar", 10*time.Second)
```
sancar, line 1139: Lets put the timeouts at least a couple of minutes(or even an hour) when it should be able to acquire. 
A couple of seconds can fail when system is under load.

map_it_test.go
yuce: It seems I've used log.Fatal instead of t.Fatal erroneously in several places, fixed them at: https://github.com/hazelcast/hazelcast-go-client/pull/648/commits/c46f0434df79f9506aa80ebbd7854bede29b4c8b

map_it_test.go
yuce: No, `t.Fatal` can only be used in the same goroutine that starts the test.

map_it_test.go
yuce: Likewise.

map_it_test.go
```diff
@@ -1071,6 +1063,145 @@ func TestMap_ExecuteOnEntries(t *testing.T) {
...
+		// unlock 5 seconds later
+		go func() {
+			time.Sleep(5 * time.Second)
+			if err := m.Unlock(ctx1, "foo"); err != nil {
+				panic(err)
+			}
+		}()
+		// TryPut with a different lock context returns false
+		ctx2 := m.NewLockContext(context.Background())
+		ok, err := m.TryPutWithTimeout(ctx2, "foo", "bar", 1*time.Second)
```
sancar, line 1133: We can start the goroutine after this first failed put attempt to make sure it will never run before this. Sleep is not enough to make sure this test will not fail randomly.

map_it_test.go
```diff
@@ -1071,6 +1062,145 @@ func TestMap_ExecuteOnEntries(t *testing.T) {
...
+		// TryPut with the same lock context returns true
+		ok, err = m.TryRemove(ctx1, "foo")
+		if err != nil {
+			t.Fatal(err)
+		}
+		assert.Equal(t, true, ok)
+	})
+}
+
+func TestMap_TryRemoveWithTimeout(t *testing.T) {
```
sancar, line 1172: My comments on timing for TryPutWithTimeout should be applied on this one as well.

replicated_map_it_test.go
```diff
@@ -208,58 +208,35 @@ func TestReplicatedMap_AddEntryListener_EntryNotifiedEventWithKey(t *testing.T)
...
-	// Skipping, since predicates are not supported with portable and JSON values
-	t.SkipNow()
-	configCallback := func(config *hz.Config) {
-		config.Serialization.SetPortableFactories(it.SamplePortableFactory{})
-	}
-	it.ReplicatedMapTesterWithConfig(t, configCallback, func(t *testing.T, m *hz.ReplicatedMap) {
-		handlerCalled := int32(0)
+func TestReplicatedMap_AddEntryListenerToKeyWithPredicate(t *testing.T) {
+	it.ReplicatedMapTester(t, func(t *testing.T, m *hz.ReplicatedMap) {
+		const targetCallCount = int32(1)
```
sancar, line 229: should not this target call count be 2 ?
There are two puts with `foo` as value.

yuce: The key and value should both match,

map_it_test.go
yuce: Updated: https://github.com/hazelcast/hazelcast-go-client/pull/648/commits/7c277e2402b453759283fb1c77ceca950daad082

map_it_test.go
yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/648/commits/7c277e2402b453759283fb1c77ceca950daad082

map_it_test.go
yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/648/commits/7c277e2402b453759283fb1c77ceca950daad082

map_it_test.go
yuce: Map names are random, so leaving out unlocking has no effect on this test. But added for completeness at: https://github.com/hazelcast/hazelcast-go-client/pull/648/commits/7c277e2402b453759283fb1c77ceca950daad082

pn_counter_it_test.go
yuce: Updated the reset test and fixed PNCounter.

internal/cluster/cluster_service.go
```diff
@@ -95,12 +99,15 @@ func (s *Service) MemberAddrs() []pubcluster.Address {
 	return s.membersMap.MemberAddrs()
 }
 
-func (s *Service) RandomDataMember() *pubcluster.MemberInfo {
-	return s.membersMap.RandomDataMember()
+func (s *Service) RandomReplica(n int) (pubcluster.MemberInfo, bool) {
```
sancar, line 102: Replica is related to partition table normally except CRDT Pn Counter. 
Like a partition has 7 replicas replica 0 is the owner and the others are backups. And this is related to all partitioned Data Structures. 
`Replica` in the PnCounter seems to have a different logic which has no relation to the partition table. It works over member-list only. 
When we put `replica` name in the cluster service it will be confusing for readers of the code. 
Can we somehow keep this logic in PNCounterProxy so that it will not be confusing.

yuce: I see. I've put those methods in the cluster service in order not to copy all members. Would it be OK to add `PNCounter` somewhere in the method names? Such as: `RandomPNCounterDataMember` ?

yuce: Moved them to PNCounter at: https://github.com/hazelcast/hazelcast-go-client/pull/648/commits/cfbd7a892b479e95c017e1cdae733ee5a09edcfc

sancar (approved):

PR #652, opened 2021-09-23
Refactor Dispatch Service

sancar:
DispathService has the following properties now:

1- It is async for the caller of Publish. Meaning Publish will not be blocked.
2- Events fired from the same thread will be handled in the same order. One will finish, then other will start.
(it is enough to make sure this per EventName)
3 - If we block an event, it should not block all events only ones that are related to same EventName.
4 - A close after publish in the same thread waits for published item to be handled(finished) .

We removed subscribeSync because Subscribe seems to be enough for all use-cases.
We  have single dispatcher now since it is enough for all. We will seperate the
events for internal and external only to make sure that users' code will not block
internals.

In order to keep the order of lifecycleEvents we had to remove the indirection between
to dispatch services. These events should directly fired from the related services like
connection manager. There I had to apply a workaround for cyclic dependency where I can not
use lifecycle events from connection manager because they are defined in the `hazelcast`
package. The workaround is that we have two lifecycle events one is internal other is
external. We map internal ones to external ones on the handler before giving them to the
user. A better solution would have been seperating lifecyle events to its own package where
both internal and hazelcast packages can access(but this is a API breaking change)

fixes https://github.com/hazelcast/hazelcast-go-client/pull/644


sancar:
Test failure will be fixed in https://github.com/hazelcast/hazelcast-go-client/pull/647/files

yuce:
Doesn't this serialize the handlers of events of the same type? E.g., all ConnectionOpened handlers would be serialized. What's wrong with running all ConnectionOpened handlers concurrently?

internal/event/dispatch_service.go
```diff
@@ -31,209 +30,91 @@ type Event interface {
...
+		logger:        logger,
 	}
-	startCh := make(chan struct{})
-	go service.start(startCh)
-	<-startCh
-	atomic.StoreInt32(&service.state, ready)
 	return service
 }
-
+// Stop the DispatchService and make sure that all the successful Publish events are handled before close
```
yuce, line 60: No, this is not guranteed. Closing a channel does not wait for all events in it to be consumed. When the process terminates, the goroutines that read from `orderChan` are terminated as well. So, there may still be events waiting in `orderChan`.

yuce:
Also, isn't it still possible that ConnectionClosed handler(s) run before ConnectionOpened ones? One of the ConnectionOpened handlers may be parked/blocked (which blocks all other handlers) when the ConnectionClosed is dispatched.

sancar:
> Doesn't this serialize the handlers of events of the same type? E.g., all ConnectionOpened handlers would be serialized. What's wrong with running all ConnectionOpened handlers concurrently?

I think my wording was not correct. I have fixed the pr description and the related method comment. I have added a test to show this behavior as well. See https://github.com/hazelcast/hazelcast-go-client/pull/652/commits/969beb5e008dd135e0d84b795bee22d46b7d6b41

internal/event/dispatch_service.go
sancar: Yes, you are right. I assumed to go behaviour wrong. I have added waitgroup to wait for this. see https://github.com/hazelcast/hazelcast-go-client/pull/652/commits/3a9d317844176b198f99e65912b90f7d5161b35b

sancar:
> Doesn't this serialize the handlers of events of the same type? E.g., all ConnectionOpened handlers (of the same subscription) would be serialized. What's wrong with running all ConnectionOpened handlers concurrently?

It is not wrong. I started with the simple requirements. We may extend the solution to run them concurrently as well. 
But I think this should not be the concern of dispatch service. One needing to handle them concurrently can easily implement it on the handler.

sancar:
> Also, isn't it still possible that ConnectionClosed handler(s) run before ConnectionOpened ones? One of the ConnectionOpened handlers may be parked/blocked (which blocks all other handlers) when the ConnectionClosed is dispatched.

I write the code assuming that ConnectionAdded/Closed are same event as in Java. It turns out it is not. I fixed those via merging related events. https://github.com/hazelcast/hazelcast-go-client/pull/652/commits/870acd7304a5576d08cbb1a6daeb4cd2844a824b

sancar:
Test failure caused by a bug on invocation system. And reported here 
https://github.com/hazelcast/hazelcast-go-client/issues/653

PR #655, opened 2021-09-26
Refactored invocation sending [API-902, API-903, API-891, API-895]

### LINKED ISSUE #627: Client hangs at the start of connection manager
https://github.com/hazelcast/hazelcast-go-client/pull/622
A failure seen on when building the pr on of the benchmarks. 
Logs are attached.
[logs_1020.zip](https://github.com/hazelcast/hazelcast-go-client/files/7141830/logs_1020.zip)

### LINKED ISSUE #638: InvocationTimeout is not honored in map.get()

I am running the following test 
```
	hazelcastClient , _ := hazelcast.StartNewClient(ctx)
	config := hazelcast.Config{}
	config.Cluster.InvocationTimeout = types.Duration(time.Second)
	myMap, err := hazelcastClient.GetMap(ctx, "my-map")
	if err != nil {
		log.Fatal(err)
	}
	myMap.Get(ctx, "k1")
```
Where I changed the cluster to throw always `HazelcastInstanceNotActive` exception for map.get operation.
	
#### Expected behaviour

We expect that myMap.get will exit with error after a second. 

#### Actual behaviour

It looks like it is retrying indefinitely. 

#### A second test similar failing test 

This time I am shutting down the client after initial setup. Again the map.get should fail with timeout error

```
	var tc = it.StartNewCluster(1)
	config := tc.DefaultConfig()
	config.Cluster.InvocationTimeout = types.Duration(time.Second * 5)
	client, err := hazelcast.StartNewClientWithConfig(context.Background(), config)
	if err != nil {
		fmt.Printf("%v\n", err)
	}
	// Create the Hazelcast client.
	ctx := context.Background()
	//hazelcastClient , _ := hazelcast.StartNewClient(ctx)
	//config := hazelcast.Config{}
	//config.Cluster.InvocationTimeout = types.Duration(time.Second)
	myMap, err := client.GetMap(ctx, "my-map")
	if err != nil {
		log.Fatal(err)
	}

	tc.Shutdown()

	myMap.Get(ctx, "k1")
```

### LINKED ISSUE #653: Invocations hang indefinitely on shutdown
Both internal and user invocations should be canceled at client shutdown. 

Following test fails at current master. 932d362
We never see last `FINISH` print. 
```
        client, _ := hz.StartNewClientWithConfig(ctx, cfg)
	m, _ := client.GetMap(ctx, "test")
	m.Put(ctx, "key", "va")

	client.Shutdown(ctx)
	m.Put(ctx, "key", "va")
```

This is the simplest test to show the problem. More complicated tests like another thread calling shutdown while concurrent operations are done on the client should pass as well.

### LINKED ISSUE #654: Client is leaking memory
Following is tested on 932d362 against 3 member cluster. 

```
        cfg := hz.Config{}
	ctx := context.Background()
	for {
		client, _ := hz.StartNewClientWithConfig(ctx, cfg)
		client.Shutdown(ctx)
	}
```

Memory usage of this code increases rapidly, while we expect it to stay at the same level.

yuce:
- Moved invocation channels to the invocation service
- Returns `hzerrors.ErrClientNotActive` if an invocation is sent after the invocation service is closed
- Added 120 seconds timeout for waiting the initial member list
- Fixes #654 
- Fixes #653 
- Fixes #627
- Fixes #638

### DIFF
--- a/client_it_test.go
+++ b/client_it_test.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"log"
 	"reflect"
+	"runtime"
 	"sync"
 	"sync/atomic"
 	"testing"
@@ -604,5 +605,37 @@ func TestInvocationTimeout(t *testing.T) {
 	}
 	if !errors.Is(err, hzerrors.ErrIO) {
 		t.Fatalf("expected hzerrors.ErrIO, got: %v", err)
+	}
+}
+
+func TestClientStartShutdownMemoryLeak(t *testing.T) {
+	tc := it.StartNewClusterWithOptions("start-shutdown-memory-leak", 42701, it.MemberCount())
+	defer tc.Shutdown()
+	config := tc.DefaultConfig()
+	if it.TraceLoggingEnabled() {
+		config.Logger.Level = logger.TraceLevel
+	}
+	if it.NonSmartEnabled() {
+		config.Cluster.Unisocket = true
+	}
+	ctx := context.Background()
+	var maxAlloc uint64
+	var m runtime.MemStats
+	for i := 0; i < 100; i++ {
+		client, err := hz.StartNewClientWithConfig(ctx, config)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if err := client.Shutdown(ctx); err != nil {
+			t.Fatal(err)
+		}
+		runtime.ReadMemStats(&m)
+		if m.Alloc > maxAlloc {
+			maxAlloc = m.Alloc
+		}
+	}
+	const allocLimit = 4 * 1024 * 1024 // 4MB
+	if maxAlloc > allocLimit {
+		t.Fatalf("memory allocation: %d > %d", maxAlloc, allocLimit)
 	}
 }
--- a/client_it_test.go
+++ b/client_it_test.go
@@ -634,7 +634,7 @@ func TestClientStartShutdownMemoryLeak(t *testing.T) {
 			maxAlloc = m.Alloc
 		}
 	}
-	const allocLimit = 4 * 1024 * 1024 // 4MB
+	const allocLimit = 6 * 1024 * 1024 // 4MB
 	if maxAlloc > allocLimit {
 		t.Fatalf("memory allocation: %d > %d", maxAlloc, allocLimit)
 	}
--- a/client_it_test.go
+++ b/client_it_test.go
@@ -598,7 +598,7 @@ func TestInvocationTimeout(t *testing.T) {
 	tc.Shutdown()
 	tic := time.Now()
 	_, err = myMap.Get(ctx, "k1")
-	took := time.Now().Sub(tic)
+	took := time.Since(tic)
 	if took > timeout+1*time.Second {
 		t.Fatalf("map get took too long!: %d", took)
 	}
--- a/client_it_test.go
+++ b/client_it_test.go
@@ -639,3 +639,26 @@ func TestClientStartShutdownMemoryLeak(t *testing.T) {
 		t.Fatalf("memory allocation: %d > %d", maxAlloc, allocLimit)
 	}
 }
+
+func TestClientInvocationAfterShutdown(t *testing.T) {
+	tc := it.StartNewClusterWithOptions("invocation-after-shutdown", 43701, it.MemberCount())
+	defer tc.Shutdown()
+	config := tc.DefaultConfig()
+	if it.TraceLoggingEnabled() {
+		config.Logger.Level = logger.TraceLevel
+	}
+	if it.NonSmartEnabled() {
+		config.Cluster.Unisocket = true
+	}
+	ctx := context.Background()
+	client := it.MustClient(hz.StartNewClientWithConfig(ctx, config))
+	m, err := client.GetMap(ctx, "my-map")
+	if err != nil {
+		t.Fatal(err)
+	}
+	it.Must(client.Shutdown(ctx))
+	_, err = m.Get(ctx, "foo")
+	if !errors.Is(err, hzerrors.ErrClientNotActive) {
+		t.Fatalf("expected hzerrors.ErrClientNotActive but received: %s", err.Error())
+	}
+}
--- a/client_it_test.go
+++ b/client_it_test.go
@@ -701,7 +701,7 @@ func TestClusterShutdownThenCheckOperationsNotHanging(t *testing.T) {
 		}
 		it.WaitEventually(t, startWg)
 		it.Must(client.Shutdown(ctx))
-		wg.Wait()
+		it.WaitEventually(t, wg)
 	})
 }
 
--- a/client_it_test.go
+++ b/client_it_test.go
@@ -587,11 +587,12 @@ func TestInvocationTimeout(t *testing.T) {
 	}
 	const timeout = 3 * time.Second
 	config.Cluster.InvocationTimeout = types.Duration(timeout)
-	client, err := hz.StartNewClientWithConfig(context.Background(), config)
+	ctx := context.Background()
+	client, err := hz.StartNewClientWithConfig(ctx, config)
 	if err != nil {
 		t.Fatal(err)
 	}
-	ctx := context.Background()
+	defer client.Shutdown(ctx)
 	myMap, err := client.GetMap(ctx, "my-map")
 	if err != nil {
 		t.Fatal(err)
@@ -661,4 +662,52 @@ func TestClientInvocationAfterShutdown(t *testing.T) {
 	if !errors.Is(err, hzerrors.ErrClientNotActive) {
 		t.Fatalf("expected hzerrors.ErrClientNotActive but received: %s", err.Error())
 	}
+}
+
+func TestClusterShutdownThenCheckOperationsNotHanging(t *testing.T) {
+	tc := it.StartNewClusterWithOptions("invocation-after-shutdown-2", 44701, it.MemberCount())
+	defer tc.Shutdown()
+	config := tc.DefaultConfig()
+	cc := &config.Cluster
+	cc.InvocationTimeout = types.Duration(24 * time.Hour)
+	cc.RedoOperation = true
+	cc.ConnectionStrategy.Timeout = types.Duration(5 * time.Second)
+	if it.TraceLoggingEnabled() {
+		config.Logger.Level = logger.TraceLevel
+	}
+	if it.NonSmartEnabled() {
+		config.Cluster.Unisocket = true
+	}
+	ctx := context.Background()
+	client := it.MustClient(hz.StartNewClientWithConfig(ctx, config))
+	m, err := client.GetMap(ctx, it.NewUniqueObjectName("my-map"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	const mapSize = 1000
+	const gc = 100
+	wg := &sync.WaitGroup{}
+	wg.Add(gc)
+	startWg := &sync.WaitGroup{}
+	startWg.Add(1)
+	o := &sync.Once{}
+	var cnt int32
+	for i := 0; i < gc; i++ {
+		go func(i int) {
+			defer wg.Done()
+			for j := 0; j < mapSize; j++ {
+				if j == mapSize/4 {
+					o.Do(func() {
+						startWg.Done()
+					})
+				}
+				// ignoring the error below, it's not relevant
+				_, _ = m.Put(ctx, j, j)
+				fmt.Println(atomic.AddInt32(&cnt, 1))
+			}
+		}(i)
+	}
+	startWg.Wait()
+	it.Must(client.Shutdown(ctx))
+	wg.Wait()
 }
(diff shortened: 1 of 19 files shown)


client_it_test.go
```diff
@@ -573,3 +574,68 @@ func TestClientVersion(t *testing.T) {
...
+	ctx := context.Background()
+	myMap, err := client.GetMap(ctx, "my-map")
+	if err != nil {
+		t.Fatal(err)
+	}
+	tc.Shutdown()
+	tic := time.Now()
+	_, err = myMap.Get(ctx, "k1")
+	took := time.Since(tic)
+	if took > timeout+1*time.Second {
```
sancar, line 603: This test will fail when there is a load on the system. I think it is ok to see that it returned an error eventually. Or make the timeout 2 minutes again if you really want to assert time.

client_it_test.go
```diff
@@ -573,3 +574,68 @@ func TestClientVersion(t *testing.T) {
...
+	config := tc.DefaultConfig()
+	if it.TraceLoggingEnabled() {
+		config.Logger.Level = logger.TraceLevel
+	}
+	if it.NonSmartEnabled() {
+		config.Cluster.Unisocket = true
+	}
+	const timeout = 3 * time.Second
+	config.Cluster.InvocationTimeout = types.Duration(timeout)
+	client, err := hz.StartNewClientWithConfig(context.Background(), config)
```
sancar, line 590: the client seems to be not shutdown in this test.

sancar:
https://github.com/hazelcast/hazelcast-go-client/issues/653 
I could not see a test verifying this issue. 

We need a test to verify that an API call after client shutdown is returning HazelcastClientNotActive. 
And secondly, if user made an API call concurrently with client.shutdown, none of the API calls should hang. 
See this test https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/client/ClientRegressionWithMockNetworkTest.java#L739

yuce:
@sancar Added both tests at https://github.com/hazelcast/hazelcast-go-client/pull/655/commits/570994bf614ac92295fb6239b47ac432122ddcfe and https://github.com/hazelcast/hazelcast-go-client/pull/655/commits/df655048cdc39b3a9d15079995fd80d0a7edfc47

client_it_test.go
```diff
@@ -573,3 +574,140 @@ func TestClientVersion(t *testing.T) {
...
+						startWg.Done()
+					})
+				}
+				// ignoring the error below, it's not relevant
+				_, _ = m.Put(ctx, j, j)
+				fmt.Println(atomic.AddInt32(&cnt, 1))
+			}
+		}(i)
+	}
+	startWg.Wait()
```
sancar, line 710: it.WaitEventually

client_it_test.go
```diff
@@ -573,3 +574,140 @@ func TestClientVersion(t *testing.T) {
...
+			defer wg.Done()
+			for j := 0; j < mapSize; j++ {
+				if j == mapSize/4 {
+					o.Do(func() {
+						startWg.Done()
+					})
+				}
+				// ignoring the error below, it's not relevant
+				_, _ = m.Put(ctx, j, j)
+				fmt.Println(atomic.AddInt32(&cnt, 1))
```
sancar, line 706: Probably a leftover from debugging

internal/cluster/invocation_factory.go
```diff
@@ -38,36 +38,36 @@ func NewConnectionInvocationFactory(config *pubcluster.Config) *ConnectionInvoca
...
-func (f *ConnectionInvocationFactory) NewInvocationOnTarget(message *proto.ClientMessage, addr pubcluster.Address) *invocation.Impl {
+func (f *ConnectionInvocationFactory) NewInvocationOnTarget(message *proto.ClientMessage, addr pubcluster.Address, start time.Time) *invocation.Impl {
 	message.SetCorrelationID(f.makeCorrelationID())
-	inv := invocation.NewImpl(message, -1, addr, time.Now().Add(f.invocationTimeout), f.redoOperation)
+	inv := invocation.NewImpl(message, -1, addr, start.Add(f.invocationTimeout), f.redoOperation)
 	return inv
 }
 
-func (f *ConnectionInvocationFactory) NewConnectionBoundInvocation(message *proto.ClientMessage, conn *Connection, handler proto.ClientMessageHandler) *ConnectionBoundInvocation {
+func (f *ConnectionInvocationFactory) NewConnectionBoundInvocation(message *proto.ClientMessage, conn *Connection, handler proto.ClientMessageHandler, start time.Time) *ConnectionBoundInvocation {
```
sancar, line 59: The old way looks more convenient. Every caller of the API seems to be calling time.Now()
What was the reason of taking `start` as parameter ?

yuce: Invocations are not mutated once created, so for each attempt a new invocation should be created (the message is copied). So, the deadline should be calculated at the time of the first invocation creation. (Otherwise the deadline would apply to a particular invocation, not the put/get, etc. request).

client_it_test.go
yuce: Removed that at: https://github.com/hazelcast/hazelcast-go-client/pull/655/commits/24c7e6c984d78284d27d169348faa7438dfb1dee

client_it_test.go
yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/655/commits/24c7e6c984d78284d27d169348faa7438dfb1dee

client_it_test.go
yuce: Removed that at: https://github.com/hazelcast/hazelcast-go-client/pull/655/commits/24c7e6c984d78284d27d169348faa7438dfb1dee

internal/invocation/invocation_service.go
```diff
@@ -17,68 +17,147 @@
...
+func (s *Service) SendRequest(ctx context.Context, inv Invocation) error {
+	if atomic.LoadInt32(&s.state) != ready {
+		return cb.WrapNonRetryableError(hzerrors.ErrClientNotActive)
+	}
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("sending invocation: %w", ctx.Err())
+	case <-s.doneCh:
+		return cb.WrapNonRetryableError(fmt.Errorf("sending urgent invocation: %w", hzerrors.ErrClientNotActive))
+	case s.requestCh <- inv:
```
sancar, line 95: There is race between `SendRequest` and `Stop` in which we can end up with writing to a closed channel. 
Lets say that we have two goroutines A and B
- From A, I called `SendRequest`
- We have checked that state is ready and we are about to run `select` statement. 
- From B, I called Stop. Change state to `stopped`
- And continued and closed all the channels. 
- A continues on select statement which will randomly select a case. If it selects the `s.requestCh <- inv:` , `PANIC`

yuce: I was only able to produce this by adding a sleep right before `select` in `SendRequest` so there's no test for this. But I've removed the lines which closes channels (except `doneCh`) since closing them is not required in this case: https://github.com/hazelcast/hazelcast-go-client/pull/655/commits/08fb8d091789a14d402521047a9e7fb88dd997fa

yuce:
@sancar I've pushed a fix to prevent a race at `connectionMap.RandomConn`. The race caused this test failure: https://github.com/hazelcast/hazelcast-go-client/runs/3725069446?check_suite_focus=true and most probably was the cause of failure to send/receive initial member list (and #627). Before the fix, I was able to reproduce the issue every time with about 10_000 iterations in `TestClientStartShutdownMemoryLeak`. After the fix, I am not able to reproduce the issue. So, I think that issue is resolved.

internal/invocation/invocation_service.go
```diff
@@ -17,68 +17,147 @@
...
+	case <-ctx.Done():
+		return fmt.Errorf("sending urgent invocation: %w", ctx.Err())
+	case <-s.doneCh:
+		return cb.WrapNonRetryableError(fmt.Errorf("sending urgent invocation: %w", hzerrors.ErrClientNotActive))
+	case s.urgentRequestCh <- inv:
+		return nil
+	}
+}
+
+func (s *Service) SendResponse(msg *proto.ClientMessage) error {
```
sancar, line 110: This method name is confusing to me. 
It sounds like we are the server and sending the response to the client. 
Can we make it something like `ReceiveResponse` /`SetResponse`

yuce: This method writes to the response channel. How about `WriteResponse`?

internal/invocation/invocation_service.go
sancar: We still have problems. It looks like we can end up with hanging invocations. 
Again lets say that we have three goroutines `A` , `B` and `C`
- `C` is the goroutine that runs `processIncoming` 
- From `A` I called `SendRequest`
- We have checked that state is ready and I am waiting on `requestCh` for `C` to take the request.
- From `B` I called `Stop`. Changed the state and close the `doneCh`
- `C` runs the `select` statement which will pick a random one. If it selects the `doneCh` , we will break the loop. 
and it seems like no one will consume from `requestCh` anymore. The problem seems to be same all other channels read via `processIncoming` . A sender to these channels can wait indefinitely. 
Make sense ?

yuce: No, the only way of writing to request channel is calling `SendRequst`, and it has the following `select`:
```
select {
case <-ctx.Done():
	return fmt.Errorf("sending invocation: %w", ctx.Err())
case <-s.doneCh:
	return cb.WrapNonRetryableError(fmt.Errorf("sending urgent invocation: %w", hzerrors.ErrClientNotActive))
case s.requestCh <- inv:
	return nil
}
```
Since `doneCh` is closed, `SendRequest` will return an error.

internal/invocation/invocation_service.go
yuce: Changed it to `WriteResponse` at: https://github.com/hazelcast/hazelcast-go-client/pull/655/commits/1ce567c695eccf2c344dcb5a41a51dd2a0de1218

internal/invocation/invocation_service.go
sancar: Ok my mistake. I fail to reason about the blocking calls on the select statements. It is new to me.  Since `requestCh` is not buffered, it will be unblocked via `case <-s.doneCh:` when channel is closed.

internal/invocation/invocation_service.go
```diff
@@ -17,68 +17,143 @@
...
+	}
+}
+
 func (s *Service) processIncoming() {
 loop:
 	for {
 		select {
-		case inv := <-s.requestCh:
+		case inv, ok := <-s.requestCh:
+			if !ok {
```
sancar, line 139: We don't need these checks anymore since we are not closing these channels.

internal/invocation/invocation_service.go
```diff
@@ -17,68 +17,143 @@
...
+	}
 	close(s.doneCh)
 }
 
 func (s *Service) SetHandler(handler Handler) {
 	s.handler = handler
 }
 
+func (s *Service) SendRequest(ctx context.Context, inv Invocation) error {
+	if atomic.LoadInt32(&s.state) != ready {
```
sancar, line 83: It looks like we don't need this state to be checked when writing to channels anymore. 
We will return from `<-s.doneCh` always when it is closed. 

It looks like it is only needed on `Stop` to make sure it is closed only once. And maybe to make it the purpose obvious ,we could use `once.Do`

client_it_test.go
```diff
@@ -573,3 +574,146 @@ func TestClientVersion(t *testing.T) {
...
+						})
+					}
+					// ignoring the error below, it's not relevant
+					_, _ = m.Put(ctx, j, j)
+				}
+			}(i)
+		}
+		it.WaitEventually(t, startWg)
+		it.Must(client.Shutdown(ctx))
+		wg.Wait()
```
sancar, line 704: it.WaitEventually

internal/cluster/invocation_factory.go
sancar: We discussed with @yuce that we might use CircutirBreaker for this purpose, instead of passing time around internals. It will be addressed in another pr if it looks like a good fit.

internal/invocation/invocation_service.go
yuce: If you don't have a strong opinion about this, I'd say let's keep the state. Even if we use `sync.Once` we'd still need a field for that. Also other components work with state, so we stay more consistent.

client_it_test.go
yuce: updated

internal/invocation/invocation_service.go
yuce: OK: https://github.com/hazelcast/hazelcast-go-client/pull/655/commits/348b8d64bf5b9b85f73f01a47bad2d244cf0ef98

internal/invocation/invocation_service.go
sancar: Lets just remove the checks from these methods then. As far as I see, they are not needed. Less code is always better. 
We just need check it on Stop

sancar (approved):

yuce:
Thanks a lot for the through review!

PR #663, opened 2021-10-04
[1.2.0] Implement ExecuteOnKeys functionality [API-912]

### LINKED ISSUE #662: Implement Map ExecuteOnKeys method
`func (m *Map) ExecuteOnKeys(ctx context.Context, processor interface{}, key interface{}...) ([]interface{}, error)` is missing. We need to add it and the corresponding test.

### Pointers
* https://github.com/hazelcast/hazelcast-go-client/pull/626/files
* https://github.com/hazelcast/hazelcast-go-client/pull/630/files

utku-caglayan:
Resolves #662

### DIFF
--- a/internal/proto/codec/map_execute_on_key_codec.go
+++ b/internal/proto/codec/map_execute_on_key_codec.go
@@ -30,7 +30,7 @@ const (
 	MapExecuteOnKeyCodecRequestInitialFrameSize = MapExecuteOnKeyCodecRequestThreadIdOffset + proto.LongSizeInBytes
 )
 
-// Applies the user defined EntryProcessor to the entry mapped by the key. Returns the object which is result of
+// Applies the user defined EntryProcessor to the entry mapped by the key. Returns the the object which is result of
 // the process() method of EntryProcessor.
 
 func EncodeMapExecuteOnKeyRequest(name string, entryProcessor *iserialization.Data, key *iserialization.Data, threadId int64) *proto.ClientMessage {
--- a/internal/proto/codec/map_execute_on_key_codec.go
+++ b/internal/proto/codec/map_execute_on_key_codec.go
@@ -30,7 +30,7 @@ const (
 	MapExecuteOnKeyCodecRequestInitialFrameSize = MapExecuteOnKeyCodecRequestThreadIdOffset + proto.LongSizeInBytes
 )
 
-// Applies the user defined EntryProcessor to the entry mapped by the key. Returns the the object which is result of
+// Applies the user defined EntryProcessor to the entry mapped by the key. Returns the object which is result of
 // the process() method of EntryProcessor.
 
 func EncodeMapExecuteOnKeyRequest(name string, entryProcessor *iserialization.Data, key *iserialization.Data, threadId int64) *proto.ClientMessage {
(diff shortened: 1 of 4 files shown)


internal/proto/codec/map_execute_on_key_codec.go
```diff
@@ -30,7 +30,7 @@ const (
 	MapExecuteOnKeyCodecRequestInitialFrameSize = MapExecuteOnKeyCodecRequestThreadIdOffset + proto.LongSizeInBytes
 )
 
-// Applies the user defined EntryProcessor to the entry mapped by the key. Returns the the object which is result of
+// Applies the user defined EntryProcessor to the entry mapped by the key. Returns the object which is result of
```
yuce, line 33: No need to change the codecs. These are supposed to be auto-generated.

proxy_map.go
```diff
@@ -908,20 +930,29 @@ func (m *Map) loadAll(ctx context.Context, replaceExisting bool, keys ...interfa
...
 }
 
+func (m *Map) convertToDataList(keys []interface{}) ([]*serialization.Data, error) {
+	keyDatas := make([]*serialization.Data, 0, len(keys))
+	for _, key := range keys {
+		// todo check if this null check done with validateAndSerialize cause problem for keys
+		if keyData, err := m.validateAndSerialize(key); err != nil {
+			return nil, err
+		} else {
+			keyDatas = append(keyDatas, keyData)
```
yuce, line 950: Can you change this to:
```
keyData, err := err := m.validateAndSerialize(key)
if err != nil  {
   return nil, err
}
keyDatas = append(keyDatas, keyData)
```

yuce (approved):
Looks good! Except two minor things.

proxy_map.go
utku-caglayan: I will be fixing that, thanks for that catch. I wasn't sure to use `validateAndSerialize` with alternative being `convertToData` which is essentially same minus a nil check. 
I know that we support "null" as value on our map structure but do we support null keys? If not I validateAndSerialize seems fine here

yuce: No, we don't support `nil` for neither the key nor the value. `validateAndSerialize` is fine here. I think just reformatting that code is fine.

proxy_map.go
```diff
@@ -931,20 +953,29 @@ func (m *Map) loadAll(ctx context.Context, replaceExisting bool, keys ...interfa
...
 		request = codec.EncodeMapLoadGivenKeysRequest(m.name, keyDatas, replaceExisting)
 	}
 	_, err := m.invokeOnRandomTarget(ctx, request, nil)
 	return err
 }
 
+func (m *Map) convertToDataList(keys []interface{}) ([]*serialization.Data, error) {
+	keyDatas := make([]*serialization.Data, 0, len(keys))
+	for _, key := range keys {
+		// todo check if this null check done with validateAndSerialize cause problem for keys
```
yuce, line 969: This shouldn't be a problem.

yuce (changes requested):

proxy_map.go
utku-caglayan: Thanks for checking that, removed

yuce (approved):
Thanks, looks good!

PR #666, opened 2021-10-07
[1.2.0] Add GetRawDataInput/Output to portable reader/writer.

### LINKED ISSUE #664: Missing getRawDataInput/Output in portable reader/writer
Hazelcast Go Client version: v1.1.0

See the Java equivalents of the APIs: [PortableReader#getRawDataInput](https://github.com/hazelcast/hazelcast/blob/2aaca65dad7546714e807eed942c52e504a02bf0/hazelcast/src/main/java/com/hazelcast/nio/serialization/PortableReader.java#L358), [PortableWriter#getRawDataOutput](https://github.com/hazelcast/hazelcast/blob/2aaca65dad7546714e807eed942c52e504a02bf0/hazelcast/src/main/java/com/hazelcast/nio/serialization/PortableWriter.java#L355)

With the current version it's required to name all serialized fields and not possible to deserialize fields serialized with raw data output.

enozcan:
Closes #664


yuce:
@enozcan Thanks for the PR, but we can't merge it, since it would break backward compatibility (`PortableReader` interface). We would probably fix that by introducing a new interface which derives from that. I'll think about that.

enozcan:
I don't think users are supposed to implement custom Portable serialization hence this won't be a compatibility problem. However, I can change the PR such that it introduces a new interface for raw in/out. WDYT? (cc @sancar)

yuce:
@sancar `PortableReader` and `PortableWriter` are both public interfaces, so everyone can use it however they like, maybe for some of their internal code? We should probably not make them interfaces in the first place, because they are huge for a Go interface. In another iteration we may deprecate the `Portable` interface, and related interfaces and convert some of them to structs (like `PortableReader`, ...).

We already discussed this with @enozcan before, here's what we discussed about how to add `GetRawDataInput`, etc. without making `PortableReader`, etc. even bigger by introducing two interfaces:
```go
type PortableRawDataOutputWriter interface {
	GetRawDataOutput() DataOutput
}

type PortableRawDataInputReader interface {
	GetRawDataInput() DataInput
}```

```go
type rawDataSt struct {
	arr []byte
}

func (r rawDataSt) FactoryID() int32 {
	return 2
}

func (r rawDataSt) ClassID() int32 {
	return 100
}

func (r rawDataSt) WritePortable(writer serialization.PortableWriter) {
	writer.(serialization.PortableRawDataOutputWriter).GetRawDataOutput().WriteByteArray(r.arr)
}

func (r *rawDataSt) ReadPortable(reader serialization.PortableReader) {
	r.arr = reader.(serialization.PortableRawDataInputReader).GetRawDataInput().ReadByteArray()
}

func TestNewPortableSerializer_RawData(t *testing.T) {
	config := &serialization.Config{}
	config.SetPortableFactories(&portableFactory1{})
	service, err := NewService(config)
	if err != nil {
		t.Fatal(err)
	}
	expected := &rawDataSt{arr: []byte("FOO")}
	data, err := service.ToData(expected)
	if err != nil {
		t.Fatal(err)
	}
	ret, err := service.ToObject(data)
	if err != nil {
		t.Fatal(err)
	}
	assert.Equal(t, expected, ret)
}
```

yuce:
@sancar There's nothing in our documentation that prevents the user from implementing that interface. Besides that, I don't want an already bloated interface to get bigger.

The method I've suggested doesn't change any public API (`PortableRawDataWriter` and `PortableRawDataReader` don't need to be in the library code, and can be added to the user code) so there's nothing related to decisions about other client libraries.

Also, this feature is needed internally and is a blocker, so I think we should include this change in v1.1.1.

yuce (approved):

yuce:
@enozcan Thanks for the PR!

enozcan:
many thanks for your reviews @sancar @yuce!

PR #684, opened 2021-11-02
[v1.1.1] Handle Lost Invocations [API-992]

yuce:
- Closes invocations when the corresponding connection is disconnected. When a connection is disconnected, the connection publishes a `GroupLost` event which contains the connection ID (group ID). and the reason of disconnection. Invocation service subscribes to that event, and closes matching invocations. The `ConnectionClosed` event is not used for this purpose in order to avoid depending on `internal/cluster` for the `internal/invocation` package which would result in cyclic dependency.
- Changes user events to use the the global subscription ID.

### DIFF
--- a/client_internals_it_test.go
+++ b/client_internals_it_test.go
@@ -59,7 +59,7 @@ func TestNotReceivedInvocation(t *testing.T) {
 		ctx := context.Background()
 		config := tc.DefaultConfig()
 		config.Cluster.Unisocket = !smart
-		client := it.MustClient(hz.StartNewClientWithConfig(ctx, tc.DefaultConfig()))
+		client := it.MustClient(hz.StartNewClientWithConfig(ctx, config))
 		defer client.Shutdown(ctx)
 		ci := hz.NewClientInternal(client)
 		var cnt int32
--- a/client_internals_it_test.go
+++ b/client_internals_it_test.go
@@ -26,7 +26,10 @@ import (
 	"time"
 
 	hz "github.com/hazelcast/hazelcast-go-client"
+	"github.com/hazelcast/hazelcast-go-client/hzerrors"
+	"github.com/hazelcast/hazelcast-go-client/internal/invocation"
 	"github.com/hazelcast/hazelcast-go-client/internal/it"
+	"github.com/hazelcast/hazelcast-go-client/internal/proto/codec"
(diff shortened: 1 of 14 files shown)


client_internals_it_test.go
```diff
@@ -47,6 +50,42 @@ func TestListenersAfterClientDisconnected(t *testing.T) {
 	})
 }
 
+func TestNotReceivedInvocation(t *testing.T) {
```
sancar, line 53: If you want to avoid accessing internals and do a test from public API, you can send a long running task to the remote. An EntryProcessor for example which just sleeps. And then close the member.

internal/cluster/connection_bound_invocation.go
```diff
@@ -71,3 +71,7 @@ func (i *ConnectionBoundInvocation) CanRetry(err error) bool {
 	}
 	return false
 }
+
+func (i *ConnectionBoundInvocation) HasGroup(groupID int64) bool {
```
sancar, line 75: Is this a leftover ?

sancar:
Can we rename GroupID to ConnectionID ? We don't have a concept like group in other clients.

internal/invocation/invocation_service.go
```diff
@@ -42,30 +45,40 @@ type Service struct {
...
 		handler:         handler,
+		eventDispatcher: eventDispacher,
 		logger:          logger,
 		state:           ready,
 	}
-	go service.processIncoming()
-	return service
+	s.eventDispatcher.Subscribe(EventGroupLost, serviceSubID, func(event event.Event) {
+		go func() {
+			s.groupLostCh <- event.(*GroupLostEvent)
```
sancar, line 77: If we are doing this in a goroutine ? Is there a need for `groupLostCh` to have a buffer of 1 ? Can it be removed as well ?

yuce: Yes, I've removed that at: https://github.com/hazelcast/hazelcast-go-client/pull/684/commits/d288c52993c2e3029d9eb3c9d73f455133b205d8

client_internals_it_test.go
yuce: I wonder whether there is an entry processor you mentioned exists in the test jar. I'll try to find that.

sancar (approved):

PR #665, opened 2021-10-06
[1.2.0] Public Log API [API-872]

utku-caglayan:
Fixes #625

PR #686, opened 2021-11-03
[1.2.0] Public Log Interface [API-872] 

### LINKED ISSUE #625: Can't use own logger with a new client version
Go version: 1.17
Hazelcast server version: 4.x
Number of the clients:
Cluster size, i.e. the number of Hazelcast cluster members:
OS version : Linux

#### Expected behaviour
logger.Config has method SetLogger that allows to set a custom logger matching Logger interface

#### Actual behaviour
logger.Config struct has no way to specify a custom logger implementation. Also there is no way to redirect log output to a file for example.

utku-caglayan:
Fixes #625 

Left PR: https://github.com/hazelcast/hazelcast-go-client/pull/665
as is on purpose, so that I thought we can compare and discuss the changes between them.

/internal/logger => logger.Logger interface can be removed by using concrete type "LogAdaptor" all over the codebase. But I think it leads to bad couplings. I would love to hear your thoughts.

Also I thought DefaultLogger type can be exported to user, as guidance or custom use. However I am not sure if this is right way, so did not go through with it yet


yuce:
> Also I thought DefaultLogger type can be exported to user, as guidance or custom use. However I am not sure if this is right way, so did not go through with it yet

That's an internal implementation detail, let's not make it public.

> /internal/logger => logger.Logger interface can be removed by using concrete type "LogAdaptor" all over the codebase. But I think it leads to bad couplings.

I am not sure about how it leads to bad coupling but without that we have to change quite a bit of logging related code, and also the logging code becomes much more verbose.

internal/cluster/failover_service.go
```diff
@@ -39,9 +39,9 @@ type CandidateCluster struct {
 	ClusterName        string
 }
 
-type addrFun func(*pubcluster.Config, ilogger.Logger) (AddressProvider, AddressTranslator)
+type addrFun func(*pubcluster.Config, logger.Logger) (AddressProvider, AddressTranslator)
```
yuce, line 42: Wer can name import aliases for public packages with `pub` prefix, like `publogger`.

yuce (changes requested):

internal/cluster/failover_service.go
utku-caglayan: This is internal logger package. I could not understand the point. Can you explain further?

internal/cluster/connection_manager.go
```diff
@@ -26,15 +26,16 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/hazelcast/hazelcast-go-client/internal/lifecycle"
+	"github.com/hazelcast/hazelcast-go-client/internal/logger"
+
```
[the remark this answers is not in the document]
utku-caglayan: Oh thanks

internal/cloud/cloud_discovery.go
```diff
@@ -24,19 +24,19 @@ import (
...
 
 type DiscoveryClient struct {
-	logger     logger.Logger
+	logger     logger2.Logger
 	httpClient *rest.HTTPClient
 	token      string
 }
 
-func NewDiscoveryClient(config *cluster.CloudConfig, logger logger.Logger) *DiscoveryClient {
+func NewDiscoveryClient(config *cluster.CloudConfig, logger logger2.Logger) *DiscoveryClient {
```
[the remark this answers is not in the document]
utku-caglayan: Yes, similar cases were present on other files too. Fixed that, thanks

internal/cluster/failover_service.go
yuce: I don't think we need the internal `logger.Logger` anymore, do we?

yuce:
@utku-caglayan Could you add package level documentation to the public `logger` package in `doc,go`? It would also be nice to add `example.go` with an example about how to override the default logger.

internal/logger/default_logger.go
```diff
@@ -21,124 +21,122 @@ import (
...
-	if l.canLogInfo() {
-		callerName := l.findCallerFuncName()
-		s := callerName + "\n" + infoPrefix + ": " + fmt.Sprint(args...)
-		l.Output(logCallDepth, s)
+	loggerLevel, err := logger.GetLogLevel(l.Level)
+	if err != nil {
+		fmt.Println("logger has invalid logger level, will not logger something useful")
+		return false
 	}
+	return loggerLevel >= numericLevel
```
yuce, line 75: The variable names are not very clear. How about `currentLevel >= wantedLevel`?

internal/logger/default_logger.go
```diff
@@ -21,124 +21,122 @@ import (
...
 }
 
-func (l *DefaultLogger) Infof(format string, values ...interface{}) {
-	if l.canLogInfo() {
-		s := fmt.Sprintf("INFO : %s", fmt.Sprintf(format, values...))
-		l.Output(logCallDepth, s)
+func (l *DefaultLogger) Log(level logger.Level, formatter func() string) {
+	if !l.canLog(level) || formatter == nil {
```
yuce, line 79: No need to check for `format == nil`. In that case the code will panic, and that's OK.

internal/logger/default_logger.go
```diff
@@ -21,124 +21,122 @@ import (
...
+		return
 	}
+	s := fmt.Sprintf("%s: %s", strings.ToUpper(level.String()), formatter())
+	_ = l.Output(logCallDepth, s)
```
yuce, line 83: How about a comment explaining why the error was ignored?

internal/logger/default_logger.go
yuce: How about moving `canLog` here in order to get rid of one function call?

internal/logger/default_logger.go
```diff
@@ -21,124 +21,122 @@ import (
...
 
-func (l *DefaultLogger) Infof(format string, values ...interface{}) {
-	if l.canLogInfo() {
-		s := fmt.Sprintf("INFO : %s", fmt.Sprintf(format, values...))
-		l.Output(logCallDepth, s)
+func (l *DefaultLogger) Log(level logger.Level, formatter func() string) {
+	if !l.canLog(level) || formatter == nil {
+		return
 	}
+	s := fmt.Sprintf("%s: %s", strings.ToUpper(level.String()), formatter())
```
yuce, line 82: Note that, in the old default logger the level was padded to 5 chars, so it displayed like:
```
2021/11/09 11:19:52 TRACE: event.DispatchService.Publish: internal.cluster.connection
2021/11/09 11:19:52 INFO : connected to cluster: integration-test
```
Currently it displays:
```
2021/11/09 11:19:52 TRACE: event.DispatchService.Publish: internal.cluster.connection
2021/11/09 11:19:52 INFO: connected to cluster: integration-test
```

internal/logger/default_logger.go
```diff
@@ -21,124 +21,122 @@ import (
...
+func (la LogAdaptor) Trace(f func() string) {
+	la.Log(logger.TraceLevel, f)
 }
 
-func (l *DefaultLogger) findCallerFuncName() string {
-	pc, _, _, _ := runtime.Caller(logCallDepth)
-	return runtime.FuncForPC(pc).Name()
+// Info runs the given function to generate the logger string, if logger level is info or finer.
+func (la LogAdaptor) Info(f func() string) {
+	la.Log(logger.InfoLevel, f)
```
yuce, line 108: This `Info` function does something different than the old one. If it's not being used, let's remove it.

internal/logger/default_logger.go
```diff
@@ -21,124 +21,122 @@ import (
...
+	la.Log(logger.InfoLevel, func() string {
+		return fmt.Sprintf(format, values...)
+	})
 }
 
-func (l *DefaultLogger) canLogInfo() bool {
-	return l.Level >= infoLevel
+// Warn formats the given string with the given values, if logger level is warn or finer.
+func (la LogAdaptor) Warn(f func() string) {
+	la.Log(logger.WarnLevel, f)
```
yuce, line 120: This `Warn` function does something different than the old one. If it's not being used, let's remove it.

internal/logger/default_logger.go
```diff
@@ -21,124 +21,122 @@ import (
...
+		return err.Error()
+	})
 }
 
-func (l *DefaultLogger) CanLogDebug() bool {
-	return l.Level >= debugLevel
+// Errorf formats the given string with the given values, if logger level is error or finer.
+func (la LogAdaptor) Errorf(format string, values ...interface{}) {
+	la.Log(logger.ErrorLevel, func() string {
+		return fmt.Sprintf(format, values...)
```
yuce: This `Error` method is different from the old one, which created an error with the given format and values.

logger/logger.go
```diff
@@ -0,0 +1,51 @@
...
+	_, found := nameToLevel[Level(logLevelStr)]
+	return found
+}
+
+// GetLogLevel returns the corresponding logger level with the given string if it exists, otherwise returns an error.
+func GetLogLevel(logLevel Level) (int, error) {
+	if !isValidLogLevel(logLevel) {
+		return 0, hzerrors.NewIllegalArgumentError(fmt.Sprintf("no logger level found for %s", logLevel), nil)
+	}
+	return nameToLevel[logLevel], nil
```
yuce, line 50: Do we need to check `nameToLevel` twice? I think we could get rid of `isValidLogLevel` too.

logger/logger_config.go
```diff
@@ -25,6 +25,8 @@ type Level string
 const (
 	// OffLevel disables logging.
 	OffLevel Level = "off"
+	// CriticalLevel level. Used for errors that should definitely be noted.
```
yuce, line 28: Maybe we can rewrite these doc lines like:
```
CriticalLevel is used for errors that halts the client.
```

logger/logger_config.go
```diff
@@ -44,11 +46,16 @@ func (l Level) String() string {
 }
 
 type Config struct {
-	Level Level `json:",omitempty"`
+	Level  Level  `json:",omitempty"`
+	Custom Logger `json:"-"`
+}
+
+func (c *Config) SetLogger(customLogger Logger) {
+	c.Custom = customLogger
```
yuce, line 54: I think we don't need this method, since it doesn't make setting the logger easier.

logger/logger.go
yuce: I think we can convert `nameToLevel` to a switch statement.

proxy.go
```diff
@@ -394,7 +394,7 @@ func (p *proxy) makeEntryNotifiedListenerHandler(handler EntryNotifiedHandler) e
 		affectedEntries int32) {
 		key, value, oldValue, mergingValue, err := p.decodeEntryNotified(binKey, binValue, binOldValue, binMergingValue)
 		if err != nil {
-			p.logger.Errorf("error at AddEntryListener: %w", err)
+			p.logger.Errorf("error at AddEntryListener: %s", err)
```
yuce: Let's keep `%w`

yuce (approved):

yuce (changes requested):
Thanks for the PR. A few comments to make it even better.

client.go
```diff
@@ -217,6 +225,11 @@ func (c *Client) GetDistributedObjectsInfo(ctx context.Context) ([]types.Distrib
 	return codec.DecodeClientGetDistributedObjectsResponse(resp), nil
 }
 
+// SetLogger sets a logger implementing logger.LogAdaptor
+func (c *Client) SetLogger(logger logger.Logger) {
```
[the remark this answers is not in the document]
utku-caglayan: Thanks for catching that, removed as you suggested

client.go
```diff
@@ -107,32 +108,39 @@ func newClient(config Config) (*Client, error) {
...
 	if err != nil {
 		return nil, err
 	}
-	clientLogger := ilogger.NewWithLevel(logLevel)
 	serializationService, err := serialization.NewService(&config.Serialization)
 	if err != nil {
 		return nil, err
 	}
-	clientLogger.Trace(func() string { return fmt.Sprintf("creating new client: %s", name) })
+	clientLogger.Log(logger.TraceLevel, func() string { return fmt.Sprintf("creating new client: %s", name) })
```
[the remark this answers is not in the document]
utku-caglayan: Ty, fixed as suggested

client.go
```diff
@@ -107,32 +108,39 @@ func newClient(config Config) (*Client, error) {
...
 	c.createComponents(&config)
 	return c, nil
 }
 
+func loggerFromConf(config Config) (logger.Logger, error) {
+	if config.Logger.Custom != nil {
+		return config.Logger.Custom, nil
+	}
+	clientLogger := ilogger.NewWithLevel(config.Logger.Level)
+	return clientLogger, nil
```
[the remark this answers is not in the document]
utku-caglayan: Thanks, fixed it.

logger/logger_config.go
utku-caglayan: Agree, removed it as you suggested

internal/logger/default_logger.go
utku-caglayan: Changed this part, please take a look again

internal/logger/default_logger.go
utku-caglayan: Sure, just thought it is common practice to ignore logging error. Added a comment as suggested

internal/logger/default_logger.go
```diff
@@ -44,43 +44,39 @@ const (
...
 
 // New returns a Default Logger with defaultLogLevel.
 func New() *DefaultLogger {
+	numericLevel, _ := logger.GetLogLevel(defaultLogLevel) // defaultLogLevel exists, no err
 	return &DefaultLogger{
 		Logger: log.New(os.Stderr, "", log.LstdFlags),
-		Level:  defaultLogLevel,
+		Level:  numericLevel,
 	}
 }
```
yuce, line 57: This function can just call `NewWithLevel` with the default log level:
```
return NewWithLevel(defaultLogLevel)
```

internal/logger/default_logger.go
```diff
@@ -44,43 +44,39 @@ const (
...
 		Logger: log.New(os.Stderr, "", log.LstdFlags),
-		Level:  loggingLevel,
-	}
+		Level:  numericLevel,
+	}, nil
 }
 
-func (l *DefaultLogger) canLog(level logger.Level) bool {
+func (l *DefaultLogger) Log(level logger.Level, formatter func() string) {
 	numericLevel, err := logger.GetLogLevel(level)
```
yuce, line 71: Let's rename `numericLevel` to something better. How about `wantedLevel`?

logger/logger_config.go
```diff
@@ -25,7 +25,7 @@ type Level string
 const (
 	// OffLevel disables logging.
 	OffLevel Level = "off"
-	// CriticalLevel level. Used for errors that should definitely be noted.
+	// CriticalLevel is used for errors that halts the client.
 	CriticalLevel Level = "critical"
 	// ErrorLevel level. Logs. Used for errors that should definitely be noted.
```
yuce, line 30: Can you use the same document format for other log level doc comments? `ErrorLevel is used ...`, etc.

logger/logger.go
```diff
@@ -23,29 +22,24 @@ const (
...
+	case ErrorLevel:
+		return errorLevel, nil
+	case WarnLevel:
+		return warnLevel, nil
+	case InfoLevel:
+		return infoLevel, nil
+	case DebugLevel:
+		return debugLevel, nil
+	case TraceLevel:
+		return traceLevel, nil
```
yuce, line 41: Does it make sense to invert the order the levels to pay the least cost the common scenario, such as: Trace, Debug, Info, Warn, Error, Critical, Off? (so getting the Trace level is fastest, while getting the Off level is the slowest, which is not used much).

yuce (changes requested):

internal/cluster/failover_service.go
utku-caglayan: removed as discussed

internal/logger/default_logger.go
utku-caglayan: Thanks, will fix that

client.go
```diff
@@ -107,24 +107,23 @@ func newClient(config Config) (*Client, error) {
...
 		name = fmt.Sprintf("hz.client_%d", id)
 	}
-	logLevel, err := ilogger.GetLogLevel(config.Logger.Level)
+	serializationService, err := serialization.NewService(&config.Serialization)
 	if err != nil {
 		return nil, err
 	}
-	clientLogger := ilogger.NewWithLevel(logLevel)
-	serializationService, err := serialization.NewService(&config.Serialization)
+	clientLogger, err := loggerFromConf(config)
```
yuce, line 114: Is there a reason why you've changed the creation order of serialization service and logger?

logger/logger.go
utku-caglayan: https://play.golang.org/p/LcdE381wsK9
I could not find a meaningful difference. Do you think we can prove this? I would love to learn more about this.
usage: go test --benchmark=. benchmark_test.go

source: https://adayinthelifeof.nl/2021/03/04/go-map-vs-switch.html

If we do not see a performance improvement with the suggested ordering approach, I think it would be more readable as is right now. Wdyt?

client.go
utku-caglayan: I wanted logger to be closer to its first usage `clientLogger.Trace(func() string { return fmt.Sprintf("creating new client: %s", name) })`

logger/logger.go
yuce: In the link you've sent it mentions that the switch cases are hold in a linked list, so the cases at the head of the list should be traversed first, hence should be faster to execute (although just a tiny bit). Not sure about the readability. Does ordering the cases in reverse make th switch less readable?

utku-caglayan: Ah okay, I did not read that part.
`go test --bench=. bench_test.go 
goos: darwin
goarch: arm64
BenchmarkForFrequentLogLevels-8                         309533968                3.617 ns/op
BenchmarkForFrequentLogLevelsWithOrderedSwitch-8        330991947                3.628 ns/op
`
Seeing the benchmark results, I assumed it was a jump table. Now I re-run benchmark few times, and it seems this result was a fluke. You are right, I do observe ordered one to be the faster variant. Will fix that, thanks!

internal/logger/default_logger.go
utku-caglayan: Sure 👍

logger/logger.go
utku-caglayan: Great idea!

internal/logger/default_logger.go
```diff
@@ -21,124 +21,114 @@ import (
...
+		Weight: numericLevel,
+	}, nil
 }
 
-func (l *DefaultLogger) Infof(format string, values ...interface{}) {
-	if l.canLogInfo() {
-		s := fmt.Sprintf("INFO : %s", fmt.Sprintf(format, values...))
-		l.Output(logCallDepth, s)
+func (l *DefaultLogger) Log(wantedLevel logger.Weight, formatter func() string) {
+	if l.Weight < wantedLevel {
```
sancar, line 68: Isn't this line is wrong ? 
Lets say that we called 
```
// Infof formats the given string with the given values, if logger level is info or finer.
func (la LogAdaptor) Infof(format string, values ...interface{}) {
	la.Log(logger.WeightInfo, func() string {
		return fmt.Sprintf(format, values...)
	})
}
```
WeightInfo is 400. 
And lets say that the user configured to see the everything which means he set WeightTrace which means 
l.weight is Trace which is 600.

This means that we will not log in this case. But it should have been logged. Right ?

logger/logger.go
```diff
@@ -0,0 +1,47 @@
+package logger
+
+import (
+	"fmt"
+
+	"github.com/hazelcast/hazelcast-go-client/internal/hzerrors"
+)
+
+// Logger is the interface that is used by client for logging.
+// Level is: type Level string
```
sancar, line 10: This documentation seems like out-of-date. 
We should describe what Weight is instead here I guess

sancar (approved):

logger/logger.go
utku-caglayan: Thanks, great catch.

yuce (approved):
Thanks, looks good!

PR #689, opened 2021-11-08
[1.2.0] MultiMap [API-478]

utku-caglayan:


multi_map.go
```diff
@@ -0,0 +1,446 @@
...
+	"time"
+
+	"github.com/hazelcast/hazelcast-go-client/internal/proto"
+	"github.com/hazelcast/hazelcast-go-client/internal/proto/codec"
+	"github.com/hazelcast/hazelcast-go-client/internal/serialization"
+	"github.com/hazelcast/hazelcast-go-client/types"
+)
+
+const (
+//	maxIndexAttributes               = 255
```
utku-caglayan, line 30: @yuce I couldn't come up with a way to handle these. Should we define them separately for MultiMap? Thanks for your help 🙏

common.go
```diff
@@ -0,0 +1,26 @@
+package hazelcast
```
yuce, line 1: Does it make sense to move the functions in this file to `proxy.go`?

common.go
```diff
@@ -0,0 +1,26 @@
...
+	if ctx == nil {
+		return defaultLockID
+	}
+	if lockIDValue := ctx.Value(lockIDKey); lockIDValue == nil {
+		return defaultLockID
+	} else if lid, ok := lockIDValue.(lockID); !ok {
+		return defaultLockID
+	} else {
+		return int64(lid)
+	}
```
yuce, line 25: How about rewriting this block in the following style?
```
lidv := ctx.Value(lockIDKey)
if lockIDValue == nil {
    return defaultLockID
}
lid, ok := lidv.(lockID)
if !ok {
   return defaultLockID
}
return int64(lid)
```

example_test.go
```diff
@@ -125,6 +125,39 @@ func ExampleMap() {
...
+	}
+	// Add multiple values to existing key
+	if err = peopleMap.PutAll(ctx, "jane", "smith", "mason"); err != nil {
+		log.Fatal(err)
+	}
+	values, err := peopleMap.Get(ctx, "jane")
+	if err != nil {
+		log.Fatal(err)
+	}
+	// ["smith", "mason", "doe"] order of values are not may not be preserved
```
yuce, line 155: ...order of values may not be preserved

internal/it/multimap.go
```diff
@@ -0,0 +1,86 @@
...
+	hz "github.com/hazelcast/hazelcast-go-client"
+)
+
+func MultiMapTester(t *testing.T, f func(t *testing.T, m *hz.MultiMap)) {
+	MultiMapTesterWithConfig(t, nil, f)
+}
+
+func MultiMapTesterWithConfig(t *testing.T, configCallback func(*hz.Config), f func(t *testing.T, m *hz.MultiMap)) {
+	makeMapName := func() string {
+		return NewUniqueObjectName("map")
```
yuce, line 34: `multi-map`

multi_map_it_test.go
```diff
@@ -0,0 +1,399 @@
...
+
+	hz "github.com/hazelcast/hazelcast-go-client"
+	"github.com/hazelcast/hazelcast-go-client/internal/it"
+	"github.com/hazelcast/hazelcast-go-client/types"
+)
+
+func TestMultiMap_Put(t *testing.T) {
+	it.MultiMapTester(t, func(t *testing.T, m *hz.MultiMap) {
+		targetValue := "value1"
+		assert.False(t, !it.MustBool(m.Put(context.Background(), "key", targetValue)), "multi-map put failed")
```
yuce, line 37: Let's create and store the context in a variable, to make the code more clear (unless separate context is required):
```
ctx := context.Background()
```
Since we are testing `m.Put` it would be better to separate these into several lines, so the usage is more apparent, like:
```
increased, err := m.Put(ctx, "key", targetValue)
if err != nil {
    t.Fatal(err)
}
if !assert.True(t, increased, "multi-map put failed")
```

multi_map_it_test.go
```diff
@@ -0,0 +1,399 @@
...
+		tmp := it.MustValue(m.Get(context.Background(), "key"))
+		it.AssertEquals(t, []interface{}{targetValue}, tmp)
+
+		targetValue2 := "value2"
+		assert.False(t, !it.MustBool(m.Put(context.Background(), "key", targetValue2)), "multi-map put failed")
```
yuce, line 42: Similar comment as above.

multi_map_it_test.go
```diff
@@ -0,0 +1,399 @@
...
+		tmp = it.MustValue(m.Get(context.Background(), "key"))
```
yuce, line 43: `tmp` is not a nice variable name, how about `value`?

multi_map_it_test.go
```diff
@@ -0,0 +1,399 @@
...
+		assert.ElementsMatch(t, []interface{}{targetValue, targetValue2}, tmp)
+	})
+}
+
+//todo check if this putAll api addresses the need
```
yuce, line 48: Is that TODO a leftover?

multi_map_it_test.go
```diff
@@ -0,0 +1,399 @@
...
+func TestMultiMap_PutAll(t *testing.T) {
+	it.MultiMapTester(t, func(t *testing.T, m *hz.MultiMap) {
+		existingValue := "v4"
+		assert.False(t, !it.MustBool(m.Put(context.Background(), "k1", existingValue)), "multimap put failed")
```
yuce, line 52: Why do we negate `it.MustBool` and use `assert.False` here? Wouldn't it be more straightforward just to use `assert.True`?

multi_map_it_test.go
```diff
@@ -0,0 +1,399 @@
...
+		value := it.MustValue(m.Get(context.Background(), "k1"))
+		it.AssertEquals(t, []interface{}{existingValue}, value)
+		it.Must(m.PutAll(context.Background(), "k1", "v1", "v2", "v3"))
```
yuce, line 55: The same comments as the test above, about separating the operation to its own line. It would be great if you could update other tests as well.

multi_map_it_test.go
```diff
@@ -0,0 +1,399 @@
...
+			types.NewEntry("k3", "v3"),
+		}
+		for _, target := range targets {
+			assert.False(t, !it.MustBool(m.Put(context.Background(), target.Key, target.Value)), "multi-map put failed")
+		}
+		if entries, err := m.GetEntrySet(context.Background()); err != nil {
+			t.Fatal(err)
+		} else if !entriesEqualUnordered(targets, entries) {
+			t.Fatalf("target: %#v != %#v", targets, entries)
+		}
```
yuce, line 146: Let's try to ditch the `else`.

multi_map_it_test.go
```diff
@@ -0,0 +1,399 @@
...
+					intValue = v[0].(int64)
+				}
+				intValue++
+				if _, err := cm.Remove(lockCtx, key); err != nil {
+					panic(err)
+				}
+				if successful, err := cm.Put(lockCtx, key, intValue); err != nil {
+					panic(err)
+				} else if !successful {
+					panic("operation multi-map put failed")
```
yuce, line 179: Let's try to ditch the `else`. Likewise for other tests.

proxy_multi_map.go
```diff
@@ -0,0 +1,515 @@
...
+const (
+//	maxIndexAttributes               = 255
+//	defaultLockID                    = 0
+//	lockIDKey          lockIDKeyType = "__hz_lock_id"
+//	leaseUnset                       = -1
+)
+
+//type lockID int64
+//type lockIDKeyType string
+
```
yuce, line 38: Forgotten commented lines

proxy_multi_map.go
```diff
@@ -0,0 +1,515 @@
...
+In the example below, a listener configuration for added and updated entries is created.
+Entries only with key "somekey":
+
+	entryListenerConfig := hazelcast.MultiMapEntryListenerConfig{
+		Key: "somekey",
+		IncludeValue: true,
+	}
+	m, err := client.GetMap(ctx, "somemap")
+	entryListenerConfig.NotifyEntryAdded(true)
+	entryListenerConfig.NotifyEntryUpdated(true)
```
yuce, line 58: Should `entryListenerConfig` lines moved above the `GetMap` line?

proxy_multi_map.go
```diff
@@ -0,0 +1,515 @@
...
+	subscriptionID, err := m.AddEntryListener(ctx, entryListenerConfig, func(event *hazelcast.EntryNotified) {
+		switch event.EventType {
+		case hazelcast.EntryAdded:
+			fmt.Println("Entry Added:", event.Value)
+		case hazelcast.EntryRemoved:
+			fmt.Println("Entry Removed:", event.Value)
+		case hazelcast.EntryUpdated:
+			fmt.Println("Entry Updated:", event.Value)
+		case hazelcast.EntryEvicted:
+			fmt.Println("Entry Remove:", event.Value)
```
yuce, line 71: Evicted

proxy_multi_map.go
```diff
@@ -0,0 +1,515 @@
...
+	return &MultiMap{proxy: p}
+}
+
+// NewLockContext augments the passed parent context with a unique lock ID.
+// If passed context is nil, context.Background is used as the parent context.
+func (m *MultiMap) NewLockContext(ctx context.Context) context.Context {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	// todo this is shared with map, should we use a different lockIDKey?
```
yuce, line 123: No need. The IDs belong to different spaces (although they may be the same)

proxy_multi_map.go
```diff
@@ -0,0 +1,515 @@
...
+	return context.WithValue(ctx, lockIDKey, lockID(m.refIDGen.NextID()))
+}
+
+// AddEntryListener adds a continuous entry listener to this map.
```
yuce, line 127: multi-map.

proxy_multi_map.go
```diff
@@ -0,0 +1,515 @@
...
+func (m *MultiMap) AddEntryListener(ctx context.Context, config MultiMapEntryListenerConfig, handler EntryNotifiedHandler) (types.UUID, error) {
+	//todo does smart field corresponds to "localOnly"?
```
yuce, line 129: That's right. In smart mode, the member does not send the request to another member (so it is local). In unisocket mode, it forwards the request to other members.

proxy_multi_map.go
```diff
@@ -0,0 +1,515 @@
...
+}
+
+// Clear deletes all entries one by one and fires related events
+func (m *MultiMap) Clear(ctx context.Context) error {
+	request := codec.EncodeMultiMapClearRequest(m.name)
+	_, err := m.invokeOnRandomTarget(ctx, request, nil)
+	return err
+}
+
+// ContainsKey returns true if the map contains an entry with the given key
```
yuce, line 140: multi-map also period is missing at the end of the the doc comment. Other doc comments may also be missing it.

proxy_multi_map.go
```diff
@@ -0,0 +1,515 @@
...
+	return context.WithValue(ctx, lockIDKey, lockID(m.refIDGen.NextID()))
+}
+
+// AddEntryListener adds a continuous entry listener to this map.
+func (m *MultiMap) AddEntryListener(ctx context.Context, config MultiMapEntryListenerConfig, handler EntryNotifiedHandler) (types.UUID, error) {
+	//todo does smart field corresponds to "localOnly"?
+	return m.addEntryListener(ctx, config.IncludeValue, m.smart, config.Key, handler)
+}
+
+// Clear deletes all entries one by one and fires related events
```
yuce, line 133: Period is missing at the end of the the doc comment.

proxy_multi_map.go
```diff
@@ -0,0 +1,515 @@
...
+	if keyData, err := m.validateAndSerialize(key); err != nil {
+		return false, err
+	} else {
+		request := codec.EncodeMultiMapContainsKeyRequest(m.name, keyData, lid)
+		if response, err := m.invokeOnKey(ctx, request, keyData); err != nil {
+			return false, err
+		} else {
+			return codec.DecodeMapContainsKeyResponse(response), nil
+		}
+	}
```
yuce, line 152: Let's try to eliminate `else`s in new code.

yuce: Also, let's use `DecodeMultiMapContainsKeyResponse`.

proxy_multi_map.go
```diff
@@ -0,0 +1,515 @@
...
+}
+
+// ContainsValue returns true if the map contains an entry with the given value
```
yuce, line 155: multi-map

proxy_multi_map.go
```diff
@@ -0,0 +1,515 @@
...
+		request := codec.EncodeMultiMapContainsValueRequest(m.name, valueData)
+		if response, err := m.invokeOnRandomTarget(ctx, request, nil); err != nil {
+			return false, err
+		} else {
+			return codec.DecodeMultiMapContainsValueResponse(response), nil
+		}
+	}
+}
+
+// Delete removes the mapping for a key from this map if it is present
```
yuce, line 169: multi-map. Other doc comments may also be using `map` instead of `multi-map`.

yuce (changes requested):

common.go
utku-caglayan: Yes that would be better, thanks

common.go
utku-caglayan: Fixed as suggested

example_test.go
utku-caglayan: thanks for catching that

internal/it/multimap.go
utku-caglayan: Fixed, thanks for catching that

multi_map_it_test.go
utku-caglayan: I agree that would be more readable. I will fix these, thank you

multi_map_it_test.go
utku-caglayan: Sure

multi_map_it_test.go
utku-caglayan: Yes it was, removing this line

multi_map_it_test.go
utku-caglayan: You are right, I guess IDE shortcuts lead to this. Fixing it!

multi_map_it_test.go
utku-caglayan: Fixed as suggested

multi_map_it_test.go
utku-caglayan: Makes sense, thanks

proxy_multi_map.go
utku-caglayan: That makes sense, thanks

proxy_multi_map.go
utku-caglayan: https://github.com/hazelcast/hazelcast-go-client/pull/689#discussion_r744799439

actually I wanted hear your thoughts for these, but some of your comments already addressed it. I am removing these

proxy_multi_map.go
utku-caglayan: Oh big mistake, thanks!

proxy_multi_map.go
utku-caglayan: Added period to all missing docs in the file and proxy_map. Thanks!

yuce (approved):
Thanks, looks good!

PR #691, opened 2021-11-10
[1.2.0] Prevent panic in listener handlers if member not found

yuce:
`clusterService.GetMemberByUUID` may return `nil` if there is no member for the given UUID. In that case, the handler for the listener is called with the zero value of `cluster.MemberInfo`.

### DIFF
--- a/events.go
+++ b/events.go
@@ -62,6 +62,8 @@ const (
 )
 
 // EntryNotified contains information about an entry event.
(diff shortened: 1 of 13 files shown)


utku-caglayan (approved):
Looks great!

sancar (approved):

yuce:
Thanks for the review!

PR #693, opened 2021-11-23
Example for MultiMap [API-478]

utku-caglayan:


examples/multimap/main.go
```diff
@@ -0,0 +1,48 @@
...
+	m, err := client.GetMultiMap(ctx, mapName)
+	if err != nil {
+		log.Fatal(err)
+	}
+	// Populate map
+	if success, err := m.Put(ctx, "key", "value1"); err != nil {
+		log.Fatal(err)
+	} else if !success {
+		log.Fatal("multi-map put operation failed")
+	}
```
yuce, line 32: Let's prefer to not use `else`.

examples/multimap/main.go
```diff
@@ -0,0 +1,48 @@
...
+	if success, err := m.Put(ctx, "key", "value2"); err != nil {
+		log.Fatal(err)
+	} else if !success {
+		log.Fatal("multi-map put operation failed")
+	}
```
yuce, line 37: Let's prefer to not use `else`.

examples/multimap/main.go
```diff
@@ -0,0 +1,48 @@
...
+		log.Fatal(err)
+	}
+	// Get a random map
+	rand.Seed(time.Now().Unix())
+	mapName := fmt.Sprintf("sample-%d", rand.Int())
+	m, err := client.GetMultiMap(ctx, mapName)
+	if err != nil {
+		log.Fatal(err)
+	}
+	// Populate map
```
yuce, line 27: `Populate the map.`

examples/multimap/main.go
```diff
@@ -0,0 +1,48 @@
...
+		log.Fatal("multi-map put operation failed")
+	}
+
+	// Get both values under the same key
+	values, err := m.Get(ctx, "key")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// We see values in an interface slice
```
yuce, line 45: How about something like: `An interface slice contains the values`.

yuce (approved):
Looks good! Just a few things to make it even better.

examples/multimap/main.go
utku-caglayan: Sure also changes map example

emreyigit (approved):

PR #696, opened 2021-12-05
[1.2.0] Event ordering

### LINKED ISSUE #692: Ensure event handlers run in the receive order per partition

utku-caglayan:
Fixes #692

Hazelcast Cluster members, respects the order of the events. Different order guarantees are given for different event types (see [docs](https://docs.hazelcast.com/hazelcast/5.0/events/global-event-configuration) for detail)

Java Client summarizes these guarantees as "Order of the events originated from the same partition is preserved".  For example events which belong to the same key of a map are received(on the clients) in the same order as they created. This is achieved via [StripeExecutor](https://github.com/hazelcast/hazelcast/blob/fb720d9e7f4b7643c31fdb61e9b83b6ca8ecc552/hazelcast/src/main/java/com/hazelcast/internal/util/executor/StripedExecutor.java) implementation which is also mentioned in the link above.

Java Client processes incoming events on a StripedExecutor, by treating event handlers as "stripedRunnable"s with respect to their partitionID. Please check [it](https://github.com/hazelcast/hazelcast/blob/33cd9e1e65f92ce1b466e6c76b0262133a2f22ea/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/listener/ClientListenerServiceImpl.java#L164) for reference.

This PR fixes event processing (execution of event-handlers) logic by respecting the order of events from the same partition.


internal/invocation/invocation_service.go
```diff
@@ -68,15 +69,17 @@ func NewService(
 		groupLostCh:     make(chan *GroupLostEvent),
 		invocations:     map[int64]Invocation{},
 		handler:         handler,
-		eventDispatcher: eventDispacher,
+		eventDispatcher: eventDispatcher,
 		logger:          logger,
 		state:           ready,
+		executor:        newStripeExecutor(5, 100),
```
emreyigit, line 75: How do you decide the size and count of the queues ?

internal/invocation/invocation_service.go
```diff
@@ -183,7 +187,16 @@ func (s *Service) handleClientMessage(msg *proto.ClientMessage) {
 				return fmt.Sprintf("invocation with unknown correlation ID: %d", correlationID)
 			})
 		} else if inv.EventHandler() != nil {
-			go inv.EventHandler()(msg)
+			handler := func() {
+				inv.EventHandler()(msg)
+			}
+			if inv.PartitionID() == -1 {
```
emreyigit, line 193: How is it possible? I couldn't find such case if I don't miss something.

emreyigit:
I think we should add more details to PRs. We experienced similar situation recently at .net side. @ihsandemir suggested it. Otherwise, reviewing becomes frustrating.

internal/invocation/invocation_service.go
utku-caglayan: When an invocation does not have a specific target partition, partitionID is set to "-1". I have read it on Java client, and see it for my self on manual tests, but sadly I cannot point you to an actual doc.

Please check reference [Java implementation](https://github.com/hazelcast/hazelcast/blob/fb720d9e7f4b7643c31fdb61e9b83b6ca8ecc552/hazelcast/src/main/java/com/hazelcast/internal/util/executor/StripedExecutor.java)

internal/invocation/invocation_service.go
utku-caglayan: Thanks for reminding that!

I initially just used default value on Java Client for count, made something up for size. I will check if these are configureable with config file, and make necessary implementation to provide that.
If you also would like to see:
https://github.com/hazelcast/hazelcast/blob/fde1a42f93233858f206f660b048abfa3a7ac51a/hazelcast/src/main/java/com/hazelcast/client/properties/ClientProperty.java#L63

internal/invocation/invocation_service.go
utku-caglayan: Refactored -1 to a constant for readability 👍

internal/invocation/invocation_service.go
emreyigit: Btw, default size looks 1m on java side.

utku-caglayan: Yes you are right! I followed that implementation with https://github.com/hazelcast/hazelcast-go-client/pull/696/commits/c4dc047d11121599b217dcaf4ecb87967d3fb9c9

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,139 @@
...
+	*testing.T
+	previousCallArg int
+}
+
+func (oc *orderChecker) call(arg int) {
+	assert.Equal(oc.T, oc.previousCallArg+1, arg, "order of the tasks is not preserved")
+	oc.previousCallArg = arg
+}
+
+func Test_defaultExecuteFnc(t *testing.T) {
```
yuce, line 24: This one should be renamed if you have renamed `defaultExecuteFnc`,

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/goleak"
+)
+
+type orderChecker struct {
+	*testing.T
+	previousCallArg int
```
yuce, line 16: `previousCallArg` should be protected with atomic, wherever it is accessed, since it is written after initialization and accessed by several goroutines.

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/goleak"
+)
+
+type orderChecker struct {
+	*testing.T
```
yuce, line 15: Embedding `T` in this struct feels wrong to me.

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,139 @@
...
+	previousCallArg int
+}
+
+func (oc *orderChecker) call(arg int) {
+	assert.Equal(oc.T, oc.previousCallArg+1, arg, "order of the tasks is not preserved")
```
yuce, line 20: How about panicing here instead of `assert.Equal`? `T` is supposed to be accessed from the goroutine it was created.

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,139 @@
...
+	oc.previousCallArg = arg
+}
+
+func Test_defaultExecuteFnc(t *testing.T) {
+	oc := &orderChecker{T: t}
+	tasks := make(chan func(), 3)
```
yuce, line 26: Let's set this to the default queue size: `10_000`.

internal/invocation/stripe_executor.go
```diff
@@ -0,0 +1,73 @@
...
+	"sync"
+)
+
+// stripeExecutor executes given "tasks" preserving the order among the ones
+// that are given with the same key
+type stripeExecutor struct {
+	quit            chan struct{}
+	wg              *sync.WaitGroup
+	executeFunction func(queue chan func(), quit chan struct{}, wg *sync.WaitGroup)
+	tasks           []chan func()
```
yuce, line 14: Let's call this field `taskQueues` to make it more clearer.

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,139 @@
...
+			key:           4,
+			expectedIndex: 0,
+		},
+		{
+			queueCount:    1,
+			key:           5,
+			expectedIndex: 0,
+		},
+	}
+	for ind, tt := range tests {
```
yuce, line 83: How about `i` instead of `ind`?

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,139 @@
...
+		t.Run(fmt.Sprintf("QueueCount: %d, Key: %d", tt.queueCount, tt.key), func(t *testing.T) {
+			se := newStripeExecutor(tt.queueCount, 0)
```
yuce, line 85: Why do we set the queue size to 0?

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,139 @@
...
+				assert.FailNow(t, "dispatcher did not dispatch to correct queue")
+			}
+		})
+	}
+}
+
+func Test_serialExecutor_start(t *testing.T) {
+	t.Logf("enabled leak check")
+	defer goleak.VerifyNone(t)
+	t.Run("Functionality test", func(t *testing.T) {
```
yuce, line 102: What's the reason of running the test in `t.Run`?

yuce (changes requested):
Here's some initial comments, will continue to review.

internal/invocation/stripe_executor.go
```diff
@@ -0,0 +1,73 @@
...
+	se.quit = make(chan struct{})
+	se.wg = &sync.WaitGroup{}
+	se.executeFunction = defaultExecuteFnc
+	return se
+}
+
+// start fires up the workers for each queue
+func (se stripeExecutor) start() {
+	se.wg.Add(int(se.queueCount))
+	for ind := range se.tasks {
```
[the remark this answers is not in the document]
utku-caglayan: It is short for "index". Should I use "i"?

cluster/event_config.go
```diff
@@ -0,0 +1,33 @@
...
+
+const (
+	// Default values differ from java impl. Also queue size is calculated differently.
+	// Java Client: queueSize per worker = defaultEventQueueCapacity / defaultEventWorkerCount
+	// Go Client: queueSize per worker = defaultEventQueueCapacity
+	defaultEventQueueCapacity = 10000
+	defaultEventWorkerCount   = 5
+)
+
+type EventConfig struct {
```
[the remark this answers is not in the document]
utku-caglayan: Sure, removing configuration bits

internal/invocation/stripe_executor.go
```diff
@@ -0,0 +1,73 @@
...
+		ind := ind
+		go se.executeFunction(se.tasks[ind], se.quit, se.wg)
+	}
+}
+
+// dispatch sends the handler "task" to the appropriate queue, "tasks"
+// with the same key end up on the same queue
```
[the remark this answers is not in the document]
utku-caglayan: Sure, thanks!

internal/invocation/stripe_executor.go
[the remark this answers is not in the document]
yuce: Yeah, `i` is more common.

internal/invocation/stripe_executor_test.go
utku-caglayan: Oh didn't know that. Thanks, will do as you suggest

internal/invocation/stripe_executor.go
utku-caglayan: Sounds great.

internal/invocation/stripe_executor_test.go
utku-caglayan: I think it should not matter, but changed it anyway with refactors

internal/invocation/stripe_executor_test.go
utku-caglayan: I missed this I guess. Fixed it!

internal/invocation/stripe_executor_test.go
utku-caglayan: Removed as suggested.

internal/invocation/stripe_executor_test.go
utku-caglayan: Of course! My bad, fixed

internal/invocation/stripe_executor.go
```diff
@@ -0,0 +1,73 @@
...
+	se.dispatch(uint32(key), handler)
+}
+
+// stop blocks until all workers are stopped.
+func (se stripeExecutor) stop() {
+	close(se.quit)
+	se.wg.Wait()
+}
+
+func (se stripeExecutor) setExecutorFnc(custom func(queue chan func(), quit chan struct{}, wg *sync.WaitGroup)) {
```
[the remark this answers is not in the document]
utku-caglayan: That would be great!

client_it_test.go
```diff
@@ -615,6 +615,8 @@ func TestClientStartShutdownMemoryLeak(t *testing.T) {
 			config.Logger.Level = logger.TraceLevel
 		}
 		config.Cluster.Unisocket = !smart
+		// TODO we should either decrease our default config values or find another way to process events more efficiently
+		// this will be temporary, fix this before PR merge
```
yuce, line 619: @utku-caglayan Was this comment addressed?

internal/invocation/invocation_service.go
```diff
@@ -34,31 +34,35 @@ const (
 	stopped = 1
 )
 
+const noSpecificPartition = -1
```
yuce, line 37: How about `randomPartition`?

yuce: Or, `noPartition`?

internal/invocation/invocation_service.go
```diff
@@ -183,7 +190,16 @@ func (s *Service) handleClientMessage(msg *proto.ClientMessage) {
 				return fmt.Sprintf("invocation with unknown correlation ID: %d", correlationID)
 			})
 		} else if inv.EventHandler() != nil {
-			go inv.EventHandler()(msg)
+			handler := func() {
```
yuce, line 193: How about making `executor.dispatch` and `executor.dispatchRandom` take a `proto.ClientMessageHandler` insted of `func()` so this extra handler is not required?

yuce (changes requested):

client_it_test.go
utku-caglayan: This is fixed, with out latest default values. However I forgot to remove this command. Thanks for pointing it out.

internal/invocation/invocation_service.go
utku-caglayan: Renamed it as randomPartition 🙏

internal/invocation/invocation_service.go
```diff
@@ -183,7 +190,16 @@ func (s *Service) handleClientMessage(msg *proto.ClientMessage) {
...
+				inv.EventHandler()(msg)
+			}
+			if inv.PartitionID() == noSpecificPartition {
+				// Execute on a random worker
+				s.executor.dispatchRandom(handler)
+				return
+			}
```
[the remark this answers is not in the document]
utku-caglayan: Refactored as suggested

internal/invocation/invocation_service.go
```diff
@@ -183,7 +190,16 @@ func (s *Service) handleClientMessage(msg *proto.ClientMessage) {
...
+			partitionID := uint32(inv.PartitionID())
+			s.executor.dispatch(partitionID, handler)
```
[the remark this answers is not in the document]
utku-caglayan: Refactored as suggested

internal/invocation/stripe_executor.go
```diff
@@ -0,0 +1,89 @@
...
+	se.execFn = defaultExecFn
+	return se
+}
+
+// start fires up the workers for each queue.
+func (se stripeExecutor) start() {
+	se.wg.Add(int(se.queueCount))
+	for i := range se.taskQueues {
+		// copy to temp var
+		index := i
```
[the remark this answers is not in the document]
utku-caglayan: It seems it is not, great catch.

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,142 @@
...
+	oc.previousCallArg = arg
+}
+
+func Test_defaultExecFn(t *testing.T) {
+	var oc orderChecker
+	tasks := make(chan func(), 10_000)
+	for i := 1; i < 4; i++ {
+		tmp := i
+		tasks <- func() {
+			oc.call(tmp)
```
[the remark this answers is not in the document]
utku-caglayan: channel "tasks" accepts values of type `func()`, because of that I am skipping this if its okay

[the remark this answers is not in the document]
yuce: Right, my bad.

internal/invocation/invocation_service.go
emreyigit: I didn't understand, Will you keep it same with java or leave it 100? It looks **Outdated** but I couldn't see any change.

internal/invocation/stripe_executor.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func defaultExecFn(queue chan func(), quit chan struct{}, wg *sync.WaitGroup) {
+	defer wg.Done()
+	for {
+		select {
+		case task := <-queue:
+			task()
+		case <-quit:
+			return
```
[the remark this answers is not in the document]
yuce: It's OK to discard events when the client is shutting down.

internal/invocation/stripe_executor.go
```diff
@@ -0,0 +1,95 @@
...
+// executor represents the function that will run on workers of stripeExecutor.
+type executor func(queue chan func(), quit chan struct{}, wg *sync.WaitGroup)
+
+// stripeExecutor executes given "tasks" preserving the order among the ones that are given with the same key.
+type stripeExecutor struct {
+	quit       chan struct{}
+	wg         *sync.WaitGroup
+	execFn     executor
+	taskQueues []chan func()
+	queueCount int32
```
yuce, line 27: Is there a reason why `queueCount` is `int32` instead of `int`?

internal/invocation/stripe_executor.go
```diff
@@ -0,0 +1,95 @@
...
+
+// newStripeExecutor returns a new stripeExecutor with default configuration.
+func newStripeExecutor() stripeExecutor {
+	// ignore error, default values do not raise error.
+	ex, _ := newStripeExecutorWithConf(defaultEventWorkerCount, defaultEventQueueCapacity)
+	return ex
+}
+
+// newStripeExecutor returns a new stripeExecutor with configured queueCount and queueSize.
+func newStripeExecutorWithConf(queueCount, queueSize int32) (stripeExecutor, error) {
```
yuce, line 38: Let's rename this to `newStripeExecutorWithConfig` to be consistent with the rest of the codebase.

yuce (changes requested):

client_it_test.go
```diff
@@ -140,6 +142,72 @@ func TestClientMemberEvents(t *testing.T) {
...
+		// access it with atomic package
+		count int32
+		// accumulate errors from handlers
+		handlerErr []error
+	)
+	wg.Add(1)
+	it.MustValue(m.AddEntryListener(ctx, config, func(event *hz.EntryNotified) {
+		atomic.AddInt32(&count, 1)
+		key := event.Key.(int64)
+		fmt.Println(key)
```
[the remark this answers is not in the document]
utku-caglayan: Sorry my bad

client_it_test.go
```diff
@@ -140,6 +142,72 @@ func TestClientMemberEvents(t *testing.T) {
...
+	c := it.MustValue(hz.StartNewClientWithConfig(ctx, clientConf)).(*hz.Client)
+	defer c.Shutdown(ctx)
+	ss := it.MustValue(serialization.NewService(&clientConf.Serialization)).(*serialization.Service)
+	// Create test map
+	m := it.MustValue(c.GetMap(ctx, "TestClientEventHandlingOrder")).(*hz.Map)
+	var config hz.MapEntryListenerConfig
+	config.NotifyEntryAdded(true)
+	var (
+		// have 271 partitions by default, starting from 1
+		partitionToEvent = make([]int64, 272)
```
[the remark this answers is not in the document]
utku-caglayan: Sure, will refactor for that

client_it_test.go
```diff
@@ -140,6 +142,72 @@ func TestClientMemberEvents(t *testing.T) {
...
+		handlerErr []error
+	)
+	wg.Add(1)
+	it.MustValue(m.AddEntryListener(ctx, config, func(event *hz.EntryNotified) {
+		atomic.AddInt32(&count, 1)
+		key := event.Key.(int64)
+		fmt.Println(key)
+		pid, err := calcPartitionID(ss, key)
+		if err != nil {
+			handlerErr = append(handlerErr, err)
```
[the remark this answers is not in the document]
utku-caglayan: Ofcourse, thanks for that!

internal/invocation/stripe_executor.go
utku-caglayan: Sure!

internal/invocation/stripe_executor.go
utku-caglayan: Will do a type conversion on invocation.PartitionID.

internal/invocation/stripe_executor.go
utku-caglayan: Thanks 👍

internal/invocation/invocation_service.go
utku-caglayan: @yuce also commented on this after I implemented your suggestions. For now leaving it only default values will be simpler to manage, considering that it already scales it self with the number of the CPU's dynamically.
Codebase is ready for implementing the configuration bits with ease. We thought, let's consider this in the feature if there is some issues to configure manually.
Does that sound good to you too?

internal/invocation/stripe_executor.go
utku-caglayan: I am moving this type conversion to `InvocationService.handleMessage()` since we need to it somewhere, because `invocation.PartitionID()` return "int32".
You are right this is cleaner

internal/invocation/stripe_executor.go
```diff
@@ -0,0 +1,97 @@
...
+// start fires up the workers for each queue.
+func (se stripeExecutor) start() {
+	se.wg.Add(int(se.queueCount))
+	for i := range se.taskQueues {
+		go se.execFn(se.taskQueues[i], se.quit, se.wg)
+	}
+}
+
+// dispatch sends the handler "task" to one of the appropriate taskQueues, "tasks" with the same key end up on the same queue. Returns true if queue is full and could not dispatch
+func (se stripeExecutor) dispatch(key int, task func()) (queueFull bool) {
```
yuce, line 67: I think this function should return `false` when the task cannot be dispatched. That could also help getting rid of the return variable name.

utku-caglayan: It is definitely more idiomatic thanks, addressed it at https://github.com/hazelcast/hazelcast-go-client/pull/696/commits/3bb2d0ff640966a075cfd6b125e5cf03c81abdc6

emreyigit (approved):

client_it_test.go
```diff
@@ -140,6 +143,64 @@ func TestClientMemberEvents(t *testing.T) {
...
+	ctx := context.Background()
+	c := it.MustValue(hz.StartNewClientWithConfig(ctx, conf)).(*hz.Client)
+	defer c.Shutdown(ctx)
+	ss := it.MustValue(serialization.NewService(&conf.Serialization)).(*serialization.Service)
+	// Create test map
+	m := it.MustValue(c.GetMap(ctx, "TestClientEventHandlingOrder")).(*hz.Map)
+	var lc hz.MapEntryListenerConfig
+	lc.NotifyEntryAdded(true)
+	var (
+		// have 271 partitions by default, starting from 1
```
yuce, line 168: Why it starts from 1?

client_it_test.go
```diff
@@ -140,6 +143,64 @@ func TestClientMemberEvents(t *testing.T) {
 	})
 }
 
+func calcPartitionID(ss *serialization.Service, key interface{}) (int32, error) {
```
yuce, line 146: I think this function may be named better. Maybe `calculatePartitionID` as the last resort.

internal/invocation/stripe_executor.go
```diff
@@ -0,0 +1,97 @@
...
+
+// newStripeExecutor returns a new stripeExecutor with default configuration.
+func newStripeExecutor() stripeExecutor {
+	// ignore error, default values do not raise error.
+	ex, _ := newStripeExecutorWithConfig(defaultEventWorkerCount, defaultEventQueueCapacity)
+	return ex
+}
+
+// newStripeExecutor returns a new stripeExecutor with configured queueCount and queueSize.
+func newStripeExecutorWithConfig(queueCount, queueSize int) (stripeExecutor, error) {
```
yuce, line 38: How about returning only the stripeExecutor? `error` is always nil. We can update it if in the future we need to return an error. Also, should it return `*stripeExecutor`?

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+	quit := make(chan struct{})
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go defaultExecFn(tasks, quit, &wg)
+	assert.Eventually(t, func() bool {
+		return oc.previousCallArg == 3
+	}, time.Second, time.Millisecond*200, "execute function could not finish the tasks")
```
yuce, line 43: How about `200*time.Millisecond` instead of `time.Millisecond*200`?

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+	close(quit)
+	go func() {
+		wg.Wait()
+		// Just to see if goroutine finished
+		oc.previousCallArg = 10
+	}()
+	assert.Eventually(t, func() bool {
+		return oc.previousCallArg == 10
+	}, time.Second, 200*time.Millisecond, "execute function did not notify about its finish")
```
yuce, line 52: Maybe something like:`execute function did not complete"?

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+			queueCount:    1,
+			key:           5,
+			expectedIndex: 0,
+		},
+	}
+	for i, tt := range tests {
+		t.Run(fmt.Sprintf("QueueCount: %d, Key: %d", tt.queueCount, tt.key), func(t *testing.T) {
+			se, err := newStripeExecutorWithConfig(tt.queueCount, 10_000)
+			assert.Nil(t, err)
+			tmpHandler := func() {
```
yuce, line 91: How about `task` instead of `tmpHandler`?

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+		return oc.previousCallArg == 10
+	}, time.Second, 200*time.Millisecond, "execute function did not notify about its finish")
+}
+
+func Test_serialExecutor_dispatch(t *testing.T) {
+	tests := []struct {
+		queueCount    int
+		key           int
+		expectedIndex int
+	}{
```
yuce, line 60: Should there be a test case for `key == -1` as well? If it's hard to do that in this function due to random stuff, at the very least we should have a separate test that checks `dispatch` works with negative integers too.

internal/invocation/stripe_executor_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	se := newStripeExecutor()
+	se.start()
+	go func() {
+		for _, task := range tasks {
+			if ok := se.dispatch(task.key, task.handler); !ok {
+				panic("could not dispatch event handler")
+			}
+		}
+	}()
```
yuce, line 152: Can you refactor this, so each task is dispatched by a different goroutine?

client_it_test.go
```diff
@@ -140,6 +143,64 @@ func TestClientMemberEvents(t *testing.T) {
...
+		// wait for all events to be processed
+		wg sync.WaitGroup
+		// access it with atomic package
+		count int32
+	)
+	for i := range partitionToEvent {
+		partitionToEvent[i] = make([]int, 10)
+	}
+	wg.Add(1)
+	it.MustValue(m.AddEntryListener(ctx, lc, func(event *hz.EntryNotified) {
```
yuce, line 179: It would be nicer to extract the function to a variable.

client_it_test.go
```diff
@@ -140,6 +143,64 @@ func TestClientMemberEvents(t *testing.T) {
...
+		}
+		partitionToEvent[pid] = append(partitionToEvent[pid], key)
+		if count == 1000 {
+			// last event processed
+			wg.Done()
+		}
+	}))
+	for i := 1; i <= 1000; i++ {
+		it.MustValue(m.Put(ctx, i, "test"))
+	}
```
yuce, line 195: Could you refactor this so `m.Put`s run concurrently?

yuce (changes requested):

internal/invocation/stripe_executor_test.go
utku-caglayan: List of tasks is created in a way that randomizes tasks but respect the order of them with respect to their keys, just like the situation on clusters. If we dispatch each one in a different goroutine, we cannot guarantee the order among the tasks of the same key.

client_it_test.go
utku-caglayan: I assumed it at the beginning and it seems I was wrong. Thanks a lot for catching that! Fixed at https://github.com/hazelcast/hazelcast-go-client/pull/696/commits/6faa426bc2a39bd8dbb77be78fb1465fee2e194d

internal/invocation/stripe_executor_test.go
utku-caglayan: Fixed at https://github.com/hazelcast/hazelcast-go-client/pull/696/commits/db4dcd61d1d5f15c2887eabffc894e8c33884651

internal/invocation/stripe_executor_test.go
utku-caglayan: Fixed at https://github.com/hazelcast/hazelcast-go-client/pull/696/commits/db4dcd61d1d5f15c2887eabffc894e8c33884651

internal/invocation/stripe_executor_test.go
utku-caglayan: Fixed at https://github.com/hazelcast/hazelcast-go-client/pull/696/commits/db4dcd61d1d5f15c2887eabffc894e8c33884651

client_it_test.go
utku-caglayan: Fixed at https://github.com/hazelcast/hazelcast-go-client/pull/696/commits/db4dcd61d1d5f15c2887eabffc894e8c33884651

internal/invocation/stripe_executor_test.go
utku-caglayan: Great idea! Created a separate test for that at https://github.com/hazelcast/hazelcast-go-client/pull/696/commits/99b797188a95adb57b442993319c2fa084a63e53

internal/invocation/stripe_executor.go
utku-caglayan: Got rid off error at https://github.com/hazelcast/hazelcast-go-client/pull/696/commits/67d42e5abc830c5a9b777d23da3aec8bf841e135

utku-caglayan: Refactored stripeExecutor to pointer semantics as you suggested at https://github.com/hazelcast/hazelcast-go-client/pull/696/commits/f8f778f05671ee4f60aeaabd0f062d74352436e9

client_it_test.go
utku-caglayan: Instead implemented a blackbox one as we discussed https://github.com/hazelcast/hazelcast-go-client/pull/696/commits/df8e738d0154a1476b74f10fbe6c609833e0bfe4

client_it_test.go
utku-caglayan: Sadly, could not come up with a better one. Renamed at https://github.com/hazelcast/hazelcast-go-client/pull/696/commits/5fadb182c34b30145d6eb696303670f96354140f

yuce (approved):
Thanks, looks good!

PR #702, opened 2021-12-23
[1.2.0] [EASY] Set TTL return [API-1081]

### LINKED ISSUE #701: Map.SetTTL does not return a boolean for indicating the existence of a an entry [API-1081]
`Map.SetTTL` should return both a boolean to indicate whether the entry for the given key exists:
* `true`: Entry exists
* `false`: Entry does not exist

Since this is public API, we should introduce a new method.

utku-caglayan:
Fixes #701

This work introduces "SetTTLAffected" functionality to "Map" API. We are introducing this to let users access to the boolean response of the server. We did not want to break the existing api, hence this is created as a new function.

### DIFF
--- a/map_it_test.go
+++ b/map_it_test.go
@@ -1060,7 +1060,7 @@ func TestMap_SetTTL(t *testing.T) {
 func TestMap_SetTTLAffected(t *testing.T) {
 	it.MapTester(t, func(t *testing.T, m *hz.Map) {
 		ctx := context.Background()
-		testcases := []struct {
+		testCases := []struct {
 			key        string
 			isEffected bool
 			expectErr  bool
@@ -1083,10 +1083,10 @@ func TestMap_SetTTLAffected(t *testing.T) {
 		_ = it.MustValue(m.Put(ctx, "happy path", "someValue"))
 		_ = it.MustValue(m.PutWithTTL(ctx, "setTTL on already expired key", "someValue", time.Millisecond))
 		time.Sleep(time.Millisecond)
-		for _, test := range testcases {
-			effected, err := m.SetTTLAffected(ctx, test.key, time.Second)
-			assert.Equal(t, test.expectErr, err != nil)
-			assert.Equal(t, test.isEffected, effected)
+		for _, tc := range testCases {
+			affected, err := m.SetTTLAffected(ctx, tc.key, time.Second)
+			assert.Equal(t, tc.expectErr, err != nil)
+			assert.Equal(t, tc.isEffected, affected)
 		}
 	})
 }
--- a/map_it_test.go
+++ b/map_it_test.go
@@ -1057,7 +1057,7 @@ func TestMap_SetTTL(t *testing.T) {
 	})
 }
 
-func TestMap_SetTTLEffected(t *testing.T) {
+func TestMap_SetTTLAffected(t *testing.T) {
 	it.MapTester(t, func(t *testing.T, m *hz.Map) {
 		ctx := context.Background()
 		testcases := []struct {
@@ -1084,7 +1084,7 @@ func TestMap_SetTTLEffected(t *testing.T) {
 		_ = it.MustValue(m.PutWithTTL(ctx, "setTTL on already expired key", "someValue", time.Millisecond))
 		time.Sleep(time.Millisecond)
 		for _, test := range testcases {
-			effected, err := m.SetTTLEffected(ctx, test.key, time.Second)
+			effected, err := m.SetTTLAffected(ctx, test.key, time.Second)
 			assert.Equal(t, test.expectErr, err != nil)
 			assert.Equal(t, test.isEffected, effected)
 		}
--- a/map_it_test.go
+++ b/map_it_test.go
@@ -1065,23 +1065,26 @@ func TestMap_SetTTLAffected(t *testing.T) {
 			isEffected    bool
 			errorExpected bool
 		}{
+			// happy path
 			{
-				key:        "happy path",
+				key:        "k1",
 				isEffected: true,
 			},
+			// setTTL on non-existing key
 			{
-				key:           "setTTL on non-existing key",
+				key:           "k2",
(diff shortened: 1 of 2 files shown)


map_it_test.go
```diff
@@ -1047,7 +1047,7 @@ func TestMap_SetTTL(t *testing.T) {
 		ctx := context.Background()
 		targetValue := "value"
 		it.Must(m.Set(ctx, "key", targetValue))
-		if err := m.SetTTL(ctx, "key", 20*time.Second); err != nil {
+		if err := m.SetTTL(ctx, "key", 2*time.Second); err != nil {
```
yuce, line 1050: Why the change?

map_it_test.go
```diff
@@ -1057,6 +1057,40 @@ func TestMap_SetTTL(t *testing.T) {
 	})
 }
 
+func TestMap_SetTTLAffected(t *testing.T) {
+	it.MapTester(t, func(t *testing.T, m *hz.Map) {
+		ctx := context.Background()
+		testcases := []struct {
+			key        string
+			isEffected bool
```
yuce, line 1065: Isn't `affected` more appropriate?

map_it_test.go
```diff
@@ -1057,6 +1057,40 @@ func TestMap_SetTTL(t *testing.T) {
...
+				expectErr:  false,
+			},
+			{
+				key:        "setTTL on already expired key",
+				isEffected: false,
+				expectErr:  false,
+			},
+		}
+		_ = it.MustValue(m.Put(ctx, "happy path", "someValue"))
+		_ = it.MustValue(m.PutWithTTL(ctx, "setTTL on already expired key", "someValue", time.Millisecond))
```
yuce, line 1084: `setTTL` => `PutWithTTL`?

map_it_test.go
```diff
@@ -1057,6 +1057,40 @@ func TestMap_SetTTL(t *testing.T) {
...
+		time.Sleep(time.Millisecond)
+		for _, test := range testcases {
+			effected, err := m.SetTTLAffected(ctx, test.key, time.Second)
```
yuce, line 1087: `effected` => `affected`?

map_it_test.go
```diff
@@ -1057,6 +1057,40 @@ func TestMap_SetTTL(t *testing.T) {
 	})
 }
 
+func TestMap_SetTTLAffected(t *testing.T) {
+	it.MapTester(t, func(t *testing.T, m *hz.Map) {
+		ctx := context.Background()
+		testcases := []struct {
```
yuce, line 1063: How about `testCases`?

map_it_test.go
```diff
@@ -1057,6 +1057,40 @@ func TestMap_SetTTL(t *testing.T) {
...
+			{
+				key:        "setTTL on already expired key",
+				isEffected: false,
+				expectErr:  false,
+			},
+		}
+		_ = it.MustValue(m.Put(ctx, "happy path", "someValue"))
+		_ = it.MustValue(m.PutWithTTL(ctx, "setTTL on already expired key", "someValue", time.Millisecond))
+		time.Sleep(time.Millisecond)
+		for _, test := range testcases {
```
yuce, line 1086: How about `tc` instead of `test`?

map_it_test.go
```diff
@@ -1057,6 +1057,40 @@ func TestMap_SetTTL(t *testing.T) {
 	})
 }
 
+func TestMap_SetTTLAffected(t *testing.T) {
+	it.MapTester(t, func(t *testing.T, m *hz.Map) {
+		ctx := context.Background()
+		testcases := []struct {
+			key        string
+			isEffected bool
+			expectErr  bool
```
yuce, line 1066: `Err` prefixed identifiers are usually `error`s. How about renaming that to something like `failureExpected`?

yuce (changes requested):
Looks good, expect a few minor things.

map_it_test.go
utku-caglayan: I think this takes too much time on test pipeline. Am I mistaken?

map_it_test.go
utku-caglayan: It is just to populate the map. Actual test case is in the for loop body.
If you mean something else can you explain?

map_it_test.go
utku-caglayan: Since we are asserting for an expected error, I do not think renaming it as "failure" is a good practice. Sadly I do not have a reference to defend this point. If I can find an example for this on standard library, I will share it

utku-caglayan: https://github.com/golang/go/blob/master/src/crypto/x509/name_constraints_test.go#L46
This is an example from std lib.

map_it_test.go
yuce: 2 seconds is not enough for the TTL behaviour. That will end up with a flaky test.

map_it_test.go
yuce: You are not asserting for an expected error, but the existence of an error. The field in the link you mention is a string field (presumably the contents of an error message). How about `errorExpected` then?

map_it_test.go
yuce: OK, that key is a bit confusing. How about renaming it something else, maybe something random?

map_it_test.go
utku-caglayan: Got it, I was not aware of that. Asking just for curiosity, is this a documented behavior?
I would love to check if there is a document for these in the future, before making such changes.

map_it_test.go
utku-caglayan: That sounds good to me, thanks

map_it_test.go
utku-caglayan: Thanks, fixed as suggested!

yuce (approved):
Thanks, looks good!

alisengul53 (approved):
Looks good to me

PR #708, opened 2021-12-29
[1.2.0] Ignore download errors in rc.sh for the SQL jar

yuce:
SQL is not available for Hz < 4.1, so there is not a jar for 4.0.x line. This PR prevents the rc.sh to exit with an error if the SQL jar is not found.

### DIFF
--- a/rc.sh
+++ b/rc.sh
@@ -49,13 +49,24 @@ download () {
   local repo=$1
   local jar_path=$2
   local artifact=$3
+  # Set this to anything in the caller to ignore the download error
+  local ignore_err=${4:-}
+  local err
+  local output
   if [ -f "$jar_path" ]; then
       log_info "$jar_path already exists, skipping download."
   else
       log_info "Downloading: $jar_path ($artifact) from: $repo"
-      mvn -q dependency:get -DrepoUrl=$repo -Dartifact=$artifact -Ddest="$jar_path"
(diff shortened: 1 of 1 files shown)


utku-caglayan (approved):
LGTM, but I could not understand why coverage dropped. Do you have any idea?

yuce:
No idea, but codecov doesn't seem to measure the coverage accurately. In local it's about 81%

alisengul53:
Actually we are measuring the coverage and uploading it to codecov. I guess somehow it doesn't show result file correctly. But I will check it in detail.

rc.sh
```diff
@@ -124,6 +135,7 @@ startRC () {
   # Download Hazelcast Community jars
   downloadTests
   downloadHazelcast
+  # Download the SQL jar
   downloadSQL
```
alisengul53, line 139: Just a minor suggestion;
We can also have a server version check here. No need to try downloading sql jar for server version less then `4.1`

alisengul53 (approved):
Looks good to me

rc.sh
yuce: We will probably use the same "if something is not available, then ignore it" logic in the future for other optional components. It is possible to check the server version, but I think the current way is more robust.

yuce:
@utku-caglayan @alisengul53 Thanks for the review!

PR #718, opened 2022-01-05
Prevent panic when queries run concurrently

yuce:
Moved client shutdown code to db close, instead of conn close to prevent a panic when queries run concurrently.

### DIFF
--- a/sql/driver/driver_it_test.go
+++ b/sql/driver/driver_it_test.go
@@ -24,6 +24,7 @@ import (
 	"math/big"
 	"net/url"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


sql/driver/driver_it_test.go
```diff
@@ -417,6 +420,40 @@ func TestSetSSLConfig(t *testing.T) {
...
+		wg := &sync.WaitGroup{}
+		wg.Add(cnt)
+		for i := 0; i < cnt; i++ {
+			go func(wg *sync.WaitGroup) {
+				var k int64
+				var v string
+				res := db.QueryRow(fmt.Sprintf(`SELECT __key, this from "%s" LIMIT 1`, mapName))
+				if err := res.Scan(&k, &v); err != nil {
+					panic(err)
+				}
```
utku-caglayan, line 449: We can assert "k" and "v" here

utku-caglayan (approved):
Looks great! Minor suggestion

sql/driver/driver_it_test.go
yuce: We could do that, but we already test the scanning of correct values  in other tests. This one just makes sure that `QueryRow` can run concurrently without panics.

alisengul53 (approved):

PR #720, opened 2022-01-05
Include usage for map::executeOnKeys [API-912]

utku-caglayan:
Modifies executeOnKey example to include executeOnKeys usage

### DIFF
--- a/examples/map/execute_on_key/main.go
+++ b/examples/map/execute_on_key/main.go
@@ -20,30 +20,41 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	// Populate map
+	// Populate map.
 	if _, err = m.Put(ctx, "key1", "value"); err != nil {
 		log.Fatal(err)
 	}
 	if _, err = m.Put(ctx, "key2", "value"); err != nil {
 		log.Fatal(err)
 	}
-	// only change value corresponding to key1
-	finalVal, err := m.ExecuteOnKey(ctx, &IdentifiedEntryProcessor{value: "test"}, "key1")
+	// Only change value corresponding to "key1".
+	prevVal, err := m.ExecuteOnKey(ctx, &IdentifiedEntryProcessor{value: "testOnKey"}, "key1")
 	if err != nil {
 		log.Fatal(err)
 	}
-	fmt.Println("previous value:", finalVal)
+	fmt.Println("previous value:", prevVal)
+	// Observe value corresponding to "key1" changed, "key2" remains unchanged.
+	checkValue(ctx, m, "key1")
+	checkValue(ctx, m, "key2")
+	// Change both entries.
+	updatedTo, err := m.ExecuteOnKeys(ctx, &IdentifiedEntryProcessor{value: "testOnKeys"}, "key1", "key2")
+	if err != nil {
+		log.Fatal(err)
+	}
+	// ExecuteOnKeys returns updated results.
+	fmt.Println(updatedTo)
 	// check the values
-	value, err := m.Get(ctx, "key1")
-	if err != nil {
-		log.Fatal(err)
-	}
-	fmt.Println("key1:", value)
-	value, err = m.Get(ctx, "key2")
-	if err != nil {
-		log.Fatal(err)
-	}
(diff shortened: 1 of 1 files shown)


examples/map/execute_on_key/main.go
```diff
@@ -20,30 +20,41 @@ func main() {
...
+	fmt.Println(updatedTo)
+	// check the values
+	checkValue(ctx, m, "key1")
+	checkValue(ctx, m, "key2")
+	// Shutdown client
+	client.Shutdown(ctx)
+}
+
+// Retrieve and print value corresponding to "key" from map "m"
+func checkValue(ctx context.Context, m *hazelcast.Map, key string) {
```
yuce, line 54: No need to use context for this example, let's just use `context.TODO()` to make it simpler.

examples/map/execute_on_key/main.go
```diff
@@ -20,30 +20,41 @@ func main() {
...
 	}
-	fmt.Println("previous value:", finalVal)
-	// check the values
-	value, err := m.Get(ctx, "key1")
+	fmt.Println("previous value:", prevVal)
+	// Observe value corresponding to "key1" changed, "key2" remains unchanged.
+	checkValue(ctx, m, "key1")
+	checkValue(ctx, m, "key2")
+	// Change both entries.
+	updatedTo, err := m.ExecuteOnKeys(ctx, &IdentifiedEntryProcessor{value: "testOnKeys"}, "key1", "key2")
```
yuce, line 40: How about `newValue` instead of `updatedTo`?

yuce (approved):
Just two trivial things.

examples/map/execute_on_key/main.go
utku-caglayan: Thanks, makes sense. Fixed at https://github.com/hazelcast/hazelcast-go-client/pull/720/commits/014319c28f64f58430797e915b99c3bf348410dc

examples/map/execute_on_key/main.go
utku-caglayan: Fixed at https://github.com/hazelcast/hazelcast-go-client/pull/720/commits/014319c28f64f58430797e915b99c3bf348410dc

yuce (approved):

examples/map/execute_on_key/main.go
```diff
@@ -20,30 +20,41 @@ func main() {
...
+	fmt.Println(newValues)
+	// check the values
+	checkValue(m, "key1")
+	checkValue(m, "key2")
+	// Shutdown client
+	client.Shutdown(ctx)
+}
+
+// Retrieve and print value corresponding to "key" from map "m"
+func checkValue(m *hazelcast.Map, key string) {
```
nevzatseferoglu, line 54: Based on my reading, using `log.Fatal()` outside of the main is a bad practice. We can return `err` than check it in the main context.

examples/map/execute_on_key/main.go
```diff
@@ -20,30 +20,41 @@ func main() {
...
+	value, err := m.Get(context.TODO(), key)
 	if err != nil {
 		log.Fatal(err)
```
nevzatseferoglu, line 57: Based on my reading, using `log.Fatal()` outside of the main is a bad practice. We can return `err` than check it in the main context.
Here: https://stackoverflow.com/a/33890104/10564442

nevzatseferoglu (commented):
Looks great. Maybe, we can leave a tiny note about server side configuration of serialization, such as docs or guide links or something to make it more clear.

examples/map/execute_on_key/main.go
utku-caglayan: I believe examples and tests are similar in this regard. You can check usages of it in standard library. You can see [createTestDir](https://go.dev/src/text/template/examplefiles_test.go) helper function as an example.
It is used as a helper function to examples.

utku-caglayan:
> Looks great. Maybe, we can leave a tiny note about server side configuration of serialization, such as docs or guide links or something to make it more clear.

We provide directions for that to the user as a part of this example in "entry_processor.go".

nevzatseferoglu (approved):

PR #722, opened 2022-01-06
[1.2.0] MultiMap Code Sample: Event Listener [API-1109]

nevzatseferoglu:
Sample scenario:
- User adds an entry listener (for add and delete).
- User completes put/(remove entry) operations
- User observers handler works as expected
- User removes the handler
- AddEntryListener
- RemoveEntryListener

### DIFF
--- a/examples/multimap/event-listener/main.go
+++ b/examples/multimap/event-listener/main.go
@@ -98,7 +98,7 @@ func ObserveListenerOnKey(ctx context.Context, m *hz.MultiMap, myHandler func(*h
 			panic(err)
 		}
 	}
-	// To observe `clear` event, lets clear our map as well.
+	// Trigger a clear event on given MultiMap as well.
 	if err := m.Clear(ctx); err != nil {
 		panic(err)
 	}
--- a/examples/multimap/event-listener/main.go
+++ b/examples/multimap/event-listener/main.go
@@ -1,0 +1,103 @@
+package main
+
+import (
+	"context"
+	"log"
+
+	hz "github.com/hazelcast/hazelcast-go-client"
+	"github.com/hazelcast/hazelcast-go-client/types"
+)
+
+func main() {
+	ctx := context.TODO()
+
+	client, err := hz.StartNewClient(ctx)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	m, err := client.GetMultiMap(ctx, "my-multimap")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	myHandler := func(event *hz.EntryNotified) {
+		switch event.EventType {
+		case hz.EntryAdded:
+			log.Printf("Added: (key: %v, value: %v).\n", event.Key, event.Value)
+		case hz.EntryRemoved:
+			log.Printf("Removed: (key: %v, value: %v).\n", event.Key, event.OldValue)
+		case hz.EntryAllCleared:
+			log.Printf("Multimap was cleared.\n")
+		}
+	}
+
+	myAwesomeKey := "my-awesome-key"
+
+	listenerConfigs := [...]hz.MultiMapEntryListenerConfig{
+		{IncludeValue: true},
+		{IncludeValue: true, Key: myAwesomeKey},
+	}
+
+	listenerConfigs[0].NotifyEntryAdded(true)
+	listenerConfigs[0].NotifyEntryRemoved(true)
+
+	subscriptionID1, err := m.AddEntryListener(ctx, listenerConfigs[0], myHandler)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	myEntries1 := []types.Entry{
+		{Key: "my-key", Value: "my-value1"},
+		{Key: "my-key", Value: "my-value2"},
+		{Key: "my-key", Value: "my-value3"},
+	}
+
+	for _, entry := range myEntries1 {
+		_, err := m.Put(ctx, entry.Key, entry.Value)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+
+	for _, entry := range myEntries1 {
+		_, err := m.Remove(ctx, entry.Key)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+
+	if err := m.RemoveEntryListener(ctx, subscriptionID1); err != nil {
+		log.Fatal(err)
+	}
+
+	listenerConfigs[1].NotifyEntryAllCleared(true)
+	listenerConfigs[1].NotifyEntryAdded(true)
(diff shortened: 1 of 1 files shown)


examples/multimap/event-listener/main.go
```diff
@@ -0,0 +1,103 @@
...
+		case hz.EntryRemoved:
+			log.Printf("Multimap: %s, (key: %v, value: %v) was removed.\n", event.MapName, event.Key, event.OldValue)
+		case hz.EntryAllCleared:
+			log.Printf("Multimap: %s was cleared.\n", event.MapName)
+		}
+	}
+
+	myAwesomeKey := "my-awesome-key"
+
+	listenerConfigs := [...]hz.MultiMapEntryListenerConfig{
```
utku-caglayan, line 37: Instead of creating an array of listener config, can we define them close to their usage? Since there are only two, I feel like it would be simpler without an array.

examples/multimap/event-listener/main.go
```diff
@@ -0,0 +1,103 @@
...
+		{Key: "my-key", Value: "my-value3"},
+	}
+
+	for _, entry := range myEntries1 {
+		_, err := m.Put(ctx, entry.Key, entry.Value)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+
```
utku-caglayan, line 62: Instead of these additional newlines, maybe we can direct the user with a comment to check to logs to see processed events. Wdyt?

examples/multimap/event-listener/main.go
```diff
@@ -0,0 +1,103 @@
...
+	listenerConfigs[0].NotifyEntryAdded(true)
+	listenerConfigs[0].NotifyEntryRemoved(true)
+
+	subscriptionID1, err := m.AddEntryListener(ctx, listenerConfigs[0], myHandler)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	myEntries1 := []types.Entry{
```
utku-caglayan, line 50: It would be great if we can refactor this variable to "entries". You can also re-assign other entries instead of defining myEntries2.

examples/multimap/event-listener/main.go
```diff
@@ -0,0 +1,103 @@
...
+
+	listenerConfigs := [...]hz.MultiMapEntryListenerConfig{
+		{IncludeValue: true},
+		{IncludeValue: true, Key: myAwesomeKey},
+	}
+
+	listenerConfigs[0].NotifyEntryAdded(true)
+	listenerConfigs[0].NotifyEntryRemoved(true)
+
+	subscriptionID1, err := m.AddEntryListener(ctx, listenerConfigs[0], myHandler)
```
utku-caglayan, line 45: How about renaming this variable as subscriptionID or sID in short? We can also reuse this variable instead of defining subscriptionID2.

examples/multimap/event-listener/main.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	listenerConfigs[1].NotifyEntryAllCleared(true)
+	listenerConfigs[1].NotifyEntryAdded(true)
+	subscriptionID2, err := m.AddEntryListener(ctx, listenerConfigs[1], myHandler)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	err = m.Clear(ctx)
```
utku-caglayan, line 81:
```suggestion
	if err = m.Clear(ctx); err != nil {
```
We can handle this error on the same line

utku-caglayan (changes requested):
Thanks for the sample! I noted few minor changes

examples/multimap/event-listener/main.go
```diff
@@ -0,0 +1,119 @@
...
+	"context"
+	"log"
+
+	hz "github.com/hazelcast/hazelcast-go-client"
+	"github.com/hazelcast/hazelcast-go-client/types"
+)
+
+func main() {
+	ctx := context.TODO()
+
```
yuce, line 29: Can you use the blank lines for comments that explains the line below?

nevzatseferoglu:
`put` operation was correctly notified and handled in the correct order. But `remove` operation was not correctly handled and does not harmonize with remove order call. Therefore, I noted this comment for both examples.
https://github.com/hazelcast/hazelcast-go-client/blob/429a7b95fc999058c746da6933ce9b7f54645824/examples/multimap/event-listener/main.go#L109

Output of `ObserveListenerIncludeValueOnly`:
```
Multimap: my-multimap, (key: my-key, value: my-value1) was added.
Multimap: my-multimap, (key: my-key, value: my-value2) was added.
Multimap: my-multimap, (key: my-key, value: my-value3) was added.
Multimap: my-multimap, (key: my-key, value: my-value2) was removed.
Multimap: my-multimap, (key: my-key, value: my-value3) was removed.
Multimap: my-multimap, (key: my-key, value: my-value1) was removed.
```

examples/multimap/event-listener/main.go
```diff
@@ -0,0 +1,153 @@
...
+		{Key: "my-dummy-key", Value: "my-dummy-value"},
+	}
+	// Put my observation entries to the MultiMap
+	for _, entry := range myEntries {
+		_, err := m.Put(ctx, entry.Key, entry.Value)
+		if err != nil {
+			return err
+		}
+	}
+	// To observe `clear` event, lets clear our map as well.
```
utku-caglayan, line 102: How about "Triggers a clear event."?

examples/multimap/event-listener/main.go
```diff
@@ -0,0 +1,153 @@
...
+	myAwesomeKey := "my-awesome-key"
+	// Let's create a config for our listeners with our key
+	listenerConfig := hz.MultiMapEntryListenerConfig{IncludeValue: true, Key: myAwesomeKey}
+	// Add desired notification through config
+	listenerConfig.NotifyEntryAllCleared(true)
+	listenerConfig.NotifyEntryAdded(true)
+	// Add our continuous entry listener to the MultiMap that we requested in main
+	subscriptionID, err := m.AddEntryListener(ctx, listenerConfig, myHandler)
+	if err != nil {
+		return err
```
utku-caglayan, line 85: I believe we can panic here. Wdyt?

utku-caglayan (changes requested):
Thanks! A few improvement points to discuss.

examples/multimap/event-listener/main.go
nevzatseferoglu: Panics are also makes sense to me because it does not ruin defer statements call schedule at the end of the call stack.

utku-caglayan (approved):
Thanks for the sample! LGTM

utku-caglayan (approved):
Thanks for the samples! Looks great.

yuce (approved):
Looks good!

PR #724, opened 2022-01-07
Implement CheckNamedValue for Conn

yuce:
This PR implements `CheckNamedValue` on `driver.Conn` so any value supported by Hazelcast is supported (in contrast with the limited set of types database/sql provides).

### DIFF
--- a/internal/it/map.go
+++ b/internal/it/map.go
@@ -18,6 +18,7 @@ package it
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"go.uber.org/goleak"
@@ -87,4 +88,16 @@ func GetClientMapWithConfig(mapName string, config *hz.Config) (*hz.Client, *hz.
 	} else {
 		return client, m
 	}
+}
+
+func MapSetOnServer(clusterID string, mapName string, key, value string) *Response {
+	script := fmt.Sprintf(`
+		var map = instance_0.getMap("%s");
+        map.set(%s, %s);
+	`, mapName, key, value)
+	resp, err := rc.ExecuteOnController(context.Background(), clusterID, script, Lang_JAVASCRIPT)
+	if err != nil {
+		panic(fmt.Errorf("executing on controller: %w", err))
+	}
+	return resp
 }
--- a/internal/serialization/default_serializer.go
+++ b/internal/serialization/default_serializer.go
@@ -503,10 +503,6 @@ func (JavaArrayListSerializer) Read(input serialization.DataInput) interface{} {
 
 func (JavaArrayListSerializer) Write(o serialization.DataOutput, i interface{}) {
 	v := i.([]interface{})
-	if v == nil {
-		o.WriteInt32(nilArrayLength)
-		return
-	}
 	length := len(v)
 	o.WriteInt32(int32(length))
 	for j := 0; j < length; j++ {
--- a/internal/serialization/default_serializer.go
+++ b/internal/serialization/default_serializer.go
@@ -467,6 +467,25 @@ func (JavaLinkedListSerializer) Write(output serialization.DataOutput, i interfa
 	// no-op
 }
 
+type JavaArraySerializer struct{}
+
+func (JavaArraySerializer) ID() int32 {
+	return TypeJavaArray
+}
+
+func (JavaArraySerializer) Read(input serialization.DataInput) interface{} {
+	count := int(input.ReadInt32())
+	res := make([]interface{}, count)
+	for i := 0; i < count; i++ {
+		res[i] = input.ReadObject()
+	}
+	return res
+}
+
+func (JavaArraySerializer) Write(output serialization.DataOutput, i interface{}) {
+	// no-op
+}
+
 type JavaArrayListSerializer struct{}
 
 func (JavaArrayListSerializer) ID() int32 {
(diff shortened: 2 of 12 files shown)


utku-caglayan (approved):
Looks good!

sql/driver/driver_it_test.go
```diff
@@ -267,7 +270,11 @@ func TestSQLWithPortableData(t *testing.T) {
 			DoubleValue:   12.789,
```
srknzl, line 270: Here if I change rec and dec to be like this:

```go
                var dec types.Decimal
		rec := &Record{
			VarcharValue:  "hello",
			TinyIntValue:  -128,
			SmallIntValue: 32767,
			IntegerValue:  -27,
			BigIntValue:   38,
			BoolValue:     true,
			RealValue:     -5.32,
			DoubleValue:   12.789,
			DecimalValue:  dec,
		}

```

 and change Record struct to be:

```go

type Record struct {
	DecimalValue  types.Decimal
	NullValue     interface{}
	VarcharValue  string
	DoubleValue   float64
	BigIntValue   int64
	RealValue     float32
	IntegerValue  int32
	SmallIntValue int16
	TinyIntValue  int8
	BoolValue     bool
}
```


It fails with `runtime error: invalid memory address or nil pointer dereference` after named value checkers are run. Is this expected? Should not our type checker validate nil values?  The problem comes from the fact that dec is not nil, but its field unscaledValue is nil.

srknzl:
Looks good except one comment I left.

sql/driver/driver_it_test.go
yuce: We check `nil` values but `dec` is not a pointer, so it cannot be `nil`. The problem is, zero value of `types.Decimal` contains a `nil` `*big.Int`. I've updated the `*big.Int` serializer to handle that case by defaulting to `0` when a big int is `nil` at: https://github.com/hazelcast/hazelcast-go-client/pull/724/commits/3551354b914296e21d8d562eb03799936b631d47

I've also noticed some of the serializers were missing, added them along the way

internal/serialization/serialization_improvements_test.go
```diff
@@ -119,11 +120,31 @@ func TestSerializationImprovements(t *testing.T) {
...
+			{
+				input:  []interface{}{"foo", int64(22)},
+				name:   "JavaArrayList",
+				target: []interface{}{"foo", int64(22)},
+			},
 			{
 				input:  big.NewInt(-10_000_000),
 				name:   "JavaBigInteger",
 				target: big.NewInt(-10_000_000),
 			},
```
srknzl, line 132: Can we add this to here as well?
```go
{
	input:  []interface{}{},
	name:   "JavaArrayList-Empty",
	target: []interface{}{},
},
```

 As far as I see this fails due to a bug in array list serializer. When the slice's length is 0 we should not write -1 to the length. You can check java's ArrayListStreamSerializer. We were using NULL_ARRAY_LENGTH in serialization before 2019. It is changed in this pr: https://github.com/hazelcast/hazelcast/pull/15371/files
You might have checked python and take implementation from there. Python as well as Nodejs serialization code seems a bit out of date regarding these changes. Maybe we should sync them with server soon.

srknzl (commented):
Looks good but still I need more changes(mostly because due to new changes of yours )

internal/it/map.go
```diff
@@ -88,3 +89,15 @@ func GetClientMapWithConfig(mapName string, config *hz.Config) (*hz.Client, *hz.
 		return client, m
 	}
 }
+
+func MapSetOnServer(clusterID string, mapName string, key, value string) *Response {
```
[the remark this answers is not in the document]
yuce: It's used in `internal/serialization/server_it_test.go`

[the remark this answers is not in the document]
srknzl: Is there a file like that in this pr? Can you check again that you committed the file?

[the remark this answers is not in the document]
yuce: Oops, added them at: https://github.com/hazelcast/hazelcast-go-client/pull/724/commits/91eb5435498000f5bc44e1f75710a42848167921

internal/serialization/default_serializer.go
```diff
@@ -482,8 +501,17 @@ func (JavaArrayListSerializer) Read(input serialization.DataInput) interface{} {
...
 }
 
-func (JavaArrayListSerializer) Write(output serialization.DataOutput, i interface{}) {
-	// no-op
+func (JavaArrayListSerializer) Write(o serialization.DataOutput, i interface{}) {
+	v := i.([]interface{})
+	if v == nil {
+		o.WriteInt32(nilArrayLength)
+		return
+	}
```
srknzl, line 509: Can v be nil? If it is nil, it will be serialized with null serializer isn't it? We don't have such logic in ArrayListSerializer in Java side.

internal/serialization/serialization_improvements_test.go
yuce: You're right. Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/724/commits/c331dcb2b8fff769657f40b733f9b2111336676b

srknzl (approved):
Thank you!

yuce:
@srknzl @utku-caglayan Thanks for the review!

PR #727, opened 2022-01-18
[1.2.0] Add missing MultiMap::TryLockWithLease usage to the example [API-1107]

utku-caglayan:
add missing MultiMap::TryLockWithLease usage to the example


yuce (approved):

nevzatseferoglu:
https://github.com/hazelcast/hazelcast-go-client/blob/355ef3fac6d044b912e1cf41288cd2bc2cedbf2f/examples/multimap/lock_trylock/main.go#L49 - Prints emhasize that previous operation was `TryLockWith`, but actually it is `TryLock`.

I couldn't see any other thing. But current `TryLockWithLeaseAndTimeout` and `TryLockWithLease` does work nothing on the specified key and cannot show their timeout and lease skills on it. Maybe, It could be more clear if there is a scenario which `TryLockWithLeaseAndTimeout` and `TryLockWithLease` compete each other in different routines to lock the key by using timeout and lease arguments. Just an idea.

nevzatseferoglu (approved):
Looks good!

PR #729, opened 2022-01-18
[1.2.0] Fix SSL mode for some tests

yuce:
SSL mode (turned on by `ENABLE_SSL=1`) didn't work for some of the tests. Fixed the tests so all tests work with the SSL Mode:

```
HZ_VERSION=5.0 ENABLE_SSL=1 make test-all
```

### DIFF
--- a/internal/it/serialization.go
+++ b/internal/it/serialization.go
@@ -25,8 +25,5 @@ import (
 func SerializationTester(t *testing.T, f func(t *testing.T, config hazelcast.Config, clusterID, mapName string)) {
 	ensureRemoteController(true)
 	mapName := NewUniqueObjectName("map")
-	config := defaultTestCluster.DefaultConfig()
-	// SSL is never used with serialization tests.
-	config.Cluster.Network.SSL.Enabled = false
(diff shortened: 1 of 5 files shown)


internal/it/serialization.go
```diff
@@ -25,5 +25,8 @@ import (
 func SerializationTester(t *testing.T, f func(t *testing.T, config hazelcast.Config, clusterID, mapName string)) {
 	ensureRemoteController(true)
 	mapName := NewUniqueObjectName("map")
+	config := defaultTestCluster.DefaultConfig()
+	// SSL is never used with serialization tests.
+	config.Cluster.Network.SSL.Enabled = false
 	f(t, defaultTestCluster.DefaultConfig(), defaultTestCluster.ClusterID, mapName)
```
srknzl, line 31: should not we use config here?

```suggestion
	f(t, config, defaultTestCluster.ClusterID, mapName)
```

Also, does making SSL enabled break some tests? If not, it may not be needed to be disabled.

srknzl (approved):

utku-caglayan (approved):

yuce:
Thanks for the reviews!

PR #731, opened 2022-01-20
[1.2.0] [EASY] Refactor Test Skipper [API-1138]

yuce:
* Refactored the test skipper, so it can be tested better.
* Fixed the multiple condition behavior. All conditions must be satisfied to skip a test.
* More tests, documentation.
* Added `arch` key to `SkipIf`.


internal/it/skip.go
```diff
@@ -0,0 +1,327 @@
...
+
+func mustAtoi(s string) int {
+	n, err := strconv.Atoi(s)
+	if err != nil {
+		panic(fmt.Errorf("could not parse int %s: %w", s, err))
+	}
+	return n
+}
+
+func checkEquality(left, operator, right, key string) bool {
```
utku-caglayan, line 305: Just a suggestion, case-insensitive comparison can be easier to use.

utku-caglayan (approved):
Looks great!

internal/it/skip.go
yuce: We panic if a key (like `hz`, `version`) not found, so the user cannot make a mistake. I think let's keep it so the user cannot write something like `VeRsiON < 5.1` :)

emreyigit (approved):

yuce:
@utku-caglayan @emreyigit Thanks for the review!

PR #732, opened 2022-01-23
[1.2.0] [EASY] Respect the context in query result after the first page.

yuce:
When iterating over the pages of results, the context is not respected after the initial page. As reported by @utku-caglayan,  that's especially problematic in streaming use case, when there is no new row in the stream.

This PR fixes that by updating `QueryResult` constructor to spawn a goroutine which cancels the query when the context is canceled.

### DIFF
--- a/internal/sql/driver/result.go
+++ b/internal/sql/driver/result.go
@@ -17,6 +17,7 @@
 package driver
 
 import (
+	"context"
 	"database/sql/driver"
 	"fmt"
 	"io"
@@ -39,6 +40,7 @@ type QueryResult struct {
 	page             *sql.Page
 	ss               *SQLService
 	conn             *icluster.Connection
+	doneCh           chan struct{}
 	metadata         sql.RowMetadata
 	queryID          sql.QueryID
 	cursorBufferSize int32
@@ -47,16 +49,27 @@ type QueryResult struct {
 }
 
 // NewQueryResult creates a new QueryResult.
-func NewQueryResult(qid sql.QueryID, md sql.RowMetadata, page *sql.Page, ss *SQLService, conn *icluster.Connection, cbs int32) (*QueryResult, error) {
+func NewQueryResult(ctx context.Context, qid sql.QueryID, md sql.RowMetadata, page *sql.Page, ss *SQLService, conn *icluster.Connection, cbs int32) (*QueryResult, error) {
+	doneCh := make(chan struct{})
 	qr := &QueryResult{
 		queryID:          qid,
 		metadata:         md,
 		page:             page,
 		ss:               ss,
 		conn:             conn,
+		doneCh:           doneCh,
 		cursorBufferSize: cbs,
 		index:            0,
 	}
+	go func() {
+		select {
+		case <-ctx.Done():
+			// ignoring the error here, since there is nothing to do on error.
+			_ = qr.Close()
+		case <-doneCh:
+			return
+		}
+	}()
 	return qr, nil
 }
 
@@ -75,6 +88,7 @@ func (r *QueryResult) Columns() []string {
 // It implements database/sql/Rows interface.
 func (r *QueryResult) Close() error {
 	if atomic.CompareAndSwapInt32(&r.state, open, closed) {
+		close(r.doneCh)
 		if err := r.ss.closeQuery(r.queryID, r.conn); err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


utku-caglayan (approved):
Thanks for such a quick fix, looks great.

sql/driver/driver_it_test.go
```diff
@@ -460,6 +460,35 @@ func TestSetSSLConfig(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		defer rows.Close()
+		for rows.Next() {
+			// do nothing
+		}
+		toc := time.Now()
+		took := toc.Sub(tic).Milliseconds()
+		if took >= 2000 {
+			// the time should be less than 200 milliseconds (two pages of results), which proves that the query is canceled after the first page.
```
srknzl, line 486:
```suggestion
			// the time should be less than 2000 milliseconds (two pages of results), which proves that the query is canceled after the first page.
```

sql/driver/driver_it_test.go
```diff
@@ -460,6 +460,35 @@ func TestSetSSLConfig(t *testing.T) {
 	}
 }
 
+func TestContextCancelAfterFirstPage(t *testing.T) {
```
srknzl, line 463: If we test context cancellation is working with manual cancellation will it always work in other context cancellation variants such as timeout? 

If so, an easier and more reliable approach might be to cancel the context manually in the for loop after the first row. I am assuming the first page will contain just one row since server generates one record in a second. 

The reason I suggested this approach is the fact that initiating a query, sending the request to server and receiving and parsing the response from server can take more than 1100ms time in some cases and we can cancel the result before the first page is received.  I don't think time passed being less than 2000ms proves the otherwise. Time passed can be 1900ms and the context cancellation can happen at 1100ms and first page can be received at 1700ms.

internal/sql/driver/result.go
```diff
@@ -47,16 +49,28 @@ type QueryResult struct {
...
 		cursorBufferSize: cbs,
 		index:            0,
 	}
+	// the following goroutine cancels the query when the context is canceled.
+	go func() {
+		select {
+		case <-ctx.Done():
+			// ignoring the error here, since there is nothing to do on error.
+			_ = qr.Close()
+		case <-doneCh:
```
srknzl, line 70: when we close doneCh we execute this case right?  and `return` shuts down this goroutine

yuce: That's right.

sql/driver/driver_it_test.go
yuce: Yes, manual cancellation and timeout cancellation uses the same mechanism.

That makes sense. Updated the test at: https://github.com/hazelcast/hazelcast-go-client/pull/732/commits/67f639ffa9e7b56914698f1beee58080daea9e8f

sql/driver/driver_it_test.go
```diff
@@ -442,6 +442,53 @@ func TestSQLQueryWithQueryTimeout(t *testing.T) {
...
+		This test ensures context cancellation works for results after the first page (fetch operation).
+		generate_stream function generates N rows per second. Since the context times out after 1100 ms, it should be canceled right after getting the first number.
+	*/
+	it.SkipIf(t, "hz < 5.0")
+	it.SQLTester(t, func(t *testing.T, client *hz.Client, config *hz.Config, m *hz.Map, mapName string) {
+		db := driver.Open(*config)
+		defer db.Close()
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		tic := time.Now()
```
srknzl, line 474: we can now get rid of the timing stuff right?

sql/driver/driver_it_test.go
```diff
@@ -442,6 +442,53 @@ func TestSQLQueryWithQueryTimeout(t *testing.T) {
...
+		rows, err := db.QueryContext(ctx, "select * from table(generate_stream(1))")
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer rows.Close()
+		for rows.Next() {
+			cancel()
```
srknzl, line 481: This call to cancel happens after the first row is ready, right?

sql/driver/driver_it_test.go
yuce: No, we still want to be sure that this test takes about 1 sec.

sql/driver/driver_it_test.go
yuce: That's right. The first page is retrieved together with the `QueryContext` call above. To check, changing the loop as follows shows the first row:
```
		for rows.Next() {
			var v int64
			if err := rows.Scan(&v); err != nil {
				t.Fatal(err)
			}
			fmt.Println(v)
			cancel()
		}
```

srknzl (approved):

sql/driver/driver_it_test.go
srknzl: aha ok. no need to do that, thanks.

yuce:
@srknzl @utku-caglayan Thanks for the reviews!

PR #734, opened 2022-01-25
[1.2.0] Add valueCount to MultiMap [API-1139] [API-1180]

nevzatseferoglu:

PR #735, opened 2022-01-26
Conflict with documentation on methods which use `connectToObject` [API-1180]

nevzatseferoglu:
According to documentation;
- `m.Get(...)` should return `nil`, if the given key does not exist on the multi-map.
-  But actual implementation is different, because of underling implementation of [convertToObjects](https://github.com/hazelcast/hazelcast-go-client/blob/b0d78a4eb0a1996271bdefb7a794e43bee65f5b1/proxy.go#L308). It creates a slice as `decodedValues := make([]interface{}, len(values))`, even if given `values` is zero, it creates an object and end up becoming non-nil.

### DIFF
--- a/multi_map_it_test.go
+++ b/multi_map_it_test.go
@@ -619,3 +619,27 @@ func TestMultiMap_MultiMapEntryListener(t *testing.T) {
 		}
 	})
 }
+
+func TestMultiMap_GetObservation(t *testing.T) {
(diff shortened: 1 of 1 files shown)


yuce:
Closing, since this test is moved to https://github.com/hazelcast/hazelcast-go-client/pull/734

PR #739, opened 2022-01-31
[1.2.0] Accessing `ClusterID` of connected cluster [API-1141]

nevzatseferoglu:


client_internals_it_test.go
```diff
@@ -149,3 +152,53 @@ func (h *riggedInvocationHandler) Invoke(inv invocation.Invocation) (int64, erro
...
+		})
+		c := it.MustClient(hz.StartNewClientWithConfig(ctx, config))
+		defer func(ctx context.Context, c *hz.Client) {
+			err := c.Shutdown(ctx)
+			if err != nil {
+				t.Error("client should had shutdown properly")
+			}
+		}(ctx, c)
+		ci := hz.NewClientInternal(c)
+		prevClusterId := ci.ClusterID()
```
utku-caglayan, line 188: Do you think we should assert this against types.UUID{}?

client_internals_it_test.go
```diff
@@ -149,3 +152,53 @@ func (h *riggedInvocationHandler) Invoke(inv invocation.Invocation) (int64, erro
 	}
 	return h.Handler.Invoke(inv)
 }
+
+func TestClusterID(t *testing.T) {
+	it.SkipIf(t, "oss")
+	clientTester(t, func(t *testing.T, smart bool) {
+		ctx := context.Background()
+		cls1 := it.StartNewClusterWithOptions("clusterId-test-cluster1", 15701, it.MemberCount())
```
utku-caglayan, line 160: Can we register a defer call for cls1.Shutdown()? I am not sure if this has any side effects if cluster is already closed, however there are multiple code paths until we reach line
```cls1.Shutdown()``` and it would be great that we do not leak this cluster if something goes wrong.

File: internal/cluster/connection_manager.go [hunk 3]
```diff
@@ -265,6 +265,17 @@ func (m *ConnectionManager) checkClusterIDChanged() {
 	}
 }
 
+// GetClusterID returns the id of the connected cluster at that moment.
+func (m *ConnectionManager) GetClusterID() types.UUID {
+	m.clusterIDMu.Lock()
+	defer m.clusterIDMu.Unlock()
+	clusterID := m.clusterID
+	if clusterID == nil {
+		return types.UUID{}
```
utku-caglayan, line 274: This looks like an erroneous case. If this indeed describes a fault, what adding a boolean to return values to indicate an error is observed?

utku-caglayan (commented):
Thanks! I noted a few points to discuss

internal/cluster/connection_manager.go
yuce: This is not an erroneous case. `clusterID` is set to `nil` At `ConnectionManager.reset`.

client_internals_it_test.go
nevzatseferoglu: You are right, thank you. Fixed at [49d4f5b](https://github.com/hazelcast/hazelcast-go-client/pull/739/commits/49d4f5ba7721a379b2db599d9d0df6c6c7e4f4b5).

client_internals_it_test.go
nevzatseferoglu: I guess, in this test scenario we made sure that `prevClusterId` is not nil or empty because of the statement, 
`c := it.MustClient(hz.StartNewClientWithConfig(ctx, config))`. If the client is started with no error, It is obvious that there is a ClusterID.

internal/cluster/connection_manager.go
nevzatseferoglu: Returning an empty UUID for the inconsistent case of `ClusterID` does more make sense to me. Therefore, I didn't change the current state.

yuce: @nevzatseferoglu Could you update this method's documentation to reflect the current behavior of returning  `types.UUID{}` if the client is not connected to a cluster?

nevzatseferoglu: Fixed at [ca6ca5c](https://github.com/hazelcast/hazelcast-go-client/pull/739/commits/ca6ca5c1ee38fffe75305bb1268ead9fe0c0e882).

utku-caglayan (dismissed):
LGTM!

yuce:
@nevzatseferoglu Thanks, let's keep this PR open until v1.2.0 is released. We'll merge it after that.

yuce (approved):

utku-caglayan (approved):

PR #744, opened 2022-02-28
Add docs badge to readme

utku-caglayan:
Adds following documentation badge shortcut to readme
[Go Reference](https://pkg.go.dev/github.com/hazelcast/hazelcast-go-client)


yuce (approved):
Nice idea!

nevzatseferoglu (approved):
If you want, we can add more badges such as; `coverage`, `version`, `dependencies` etc. I suppose that there is no privacy concern on these information.

utku-caglayan:
> If you want, we can add more badges such as; `coverage`, `version`, `dependencies` etc. I suppose that there is no privacy concern on these informations.

Sounds good, if you have seen any, can you share one as an example?

yuce:
Let's merge this PR and discuss about other badges later.

PR #745, opened 2022-03-02
Removed MultiMap AddEntryListener method

yuce:


utku-caglayan (approved):
Looks good!

srknzl:
Any justification?

yuce:
@srknzl We're changing the listener API for Map and MultiMap as you know. We'll rework `AddEntryListener` for MultiMap and add it back in v1.3.0

yuce:
MultiMap is slated to be released in v1.2.0, so there is no concern about backward compatibility.

srknzl (approved):

Issue #740, opened 2022-02-15
Flaky test: TestClientEventHandlingOrder [API-1221]


nevzatseferoglu:
See: https://github.com/hazelcast/hazelcast-go-client/runs/5199820170?check_suite_focus=true#step:8:315

utku-caglayan:
Fixed with mentioned PR, closing this issue.

PR #749, opened 2022-03-07
Refactor Event Handling Order Test with correct assumptions [API-1221]

utku-caglayan:
`TestClientEventHandlingOrder` test case is written with an incorrect assumption, resulting it to fail time to time.

Previous assumption: Map entry events received by client are ordered based on partitionID that entries belong.
Current(fix) assumption: Map entry events received by client are ordered based on entry key.

This fix, addresses this assumption change by refactoring mentioned test.

### DIFF
--- a/client_it_test.go
+++ b/client_it_test.go
@@ -23,7 +23,6 @@ import (
 	"log"
 	"reflect"
 	"runtime"
-	"sort"
 	"sync"
 	"sync/atomic"
 	"testing"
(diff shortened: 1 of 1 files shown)


client_it_test.go
```diff
@@ -208,42 +207,42 @@ func TestClientEventHandlingOrder(t *testing.T) {
 	// Create custom cluster, and client from it
 	cls := it.StartNewClusterWithOptions("event-order-test-cluster", 15701, it.MemberCount())
 	defer cls.Shutdown()
-	conf := cls.DefaultConfig()
+	cnfg := cls.DefaultConfig()
```
yuce, line 210: How about `cfg` instead of `cnfg` just to be consistent with other tests?

client_it_test.go
```diff
@@ -208,42 +207,42 @@ func TestClientEventHandlingOrder(t *testing.T) {
...
+	for i := 0; i < iterationCount; i++ {
+		it.Must(m.Set(ctx, "sameKey", i))
+		it.MustValue(m.Remove(ctx, "sameKey"))
 	}
 	wg.Wait()
-	for _, keys := range partitionToEvent {
-		if !sort.IntsAreSorted(keys) {
-			t.Fatalf("events are not processed in order, event keys:\n%v\n", keys)
-		}
+	assert.Equal(t, 0, len(journal)%2)
```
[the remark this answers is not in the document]
srknzl: I wrote `eventCount * 2` firstly, now I fixed as `eventCount`

client_it_test.go
utku-caglayan: Thanks, fixed at https://github.com/hazelcast/hazelcast-go-client/pull/749/commits/3ff8826cf8dafa7914a11a5c3b203f6b6c43dd72

client_it_test.go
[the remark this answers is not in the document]
utku-caglayan: Good idea, added at https://github.com/hazelcast/hazelcast-go-client/pull/749/commits/3ff8826cf8dafa7914a11a5c3b203f6b6c43dd72

srknzl (dismissed):

client_it_test.go
```diff
@@ -208,42 +207,42 @@ func TestClientEventHandlingOrder(t *testing.T) {
...
-	conf := cls.DefaultConfig()
+	cfg := cls.DefaultConfig()
 	ctx := context.Background()
-	c := it.MustValue(hz.StartNewClientWithConfig(ctx, conf)).(*hz.Client)
+	c := it.MustValue(hz.StartNewClientWithConfig(ctx, cfg)).(*hz.Client)
 	defer c.Shutdown(ctx)
-	ss := it.MustValue(serialization.NewService(&conf.Serialization)).(*serialization.Service)
 	// Create test map
-	m := it.MustValue(c.GetMap(ctx, "TestClientEventHandlingOrder")).(*hz.Map)
+	m := it.MustValue(c.GetMap(ctx, "my-map")).(*hz.Map)
```
yuce, line 215: Could you use `NewUniqueObjectName` to generate the map name to make sure there's no conflict with other test runs? Something like:
```
ls := "smart"
if !smart {
	ls = "unisocket"
}
mapName := NewUniqueObjectName("map", ls)
```

yuce (changes requested):
Looks good, other than a minor thing.

client_it_test.go
utku-caglayan: Thanks for pointing that out! Used "it.MapTester" utility for the test at https://github.com/hazelcast/hazelcast-go-client/commit/a7fc2045715f84ad7e4a13b2dbc74d77857b3590

utku-caglayan:
> Don't you need to sync access to `journal`?

Since we are operating on the same keys, we expect that events will be handled in a synchronous manner. Considering that, do you think it is still needed?

yuce:
Yeah, I think it's better to add that.

yuce (approved):
Looks good!

utku-caglayan:
fixes #740

PR #750, opened 2022-03-09
Introduce new map listener APIs for IMap API-1231

srknzl:
See https://hazelcast.atlassian.net/wiki/spaces/HZC/pages/3997106177/Go+Client+Map+Event+Callback+Refactor+Technical+Design+Document


client_internals_it_test.go
```diff
@@ -39,15 +39,19 @@ import (
 func TestListenersAfterClientDisconnected(t *testing.T) {
 	t.Run("MemberHostname_ClientIP", func(t *testing.T) {
 		testListenersAfterClientDisconnected(t, "localhost", "127.0.0.1", 46501)
+		testListenersAfterClientDisconnectedWithSpecificEventHandler(t, "localhost", "127.0.0.1", 46501)
```
yuce, line 42: Let's make these separate tests, something like:
```
t.Run("MemberHostname_ClientIP_new_listener_interface", func(t *testing.T) {
    testListenersAfterClientDisconnectedWithSpecificEventHandler(t, "localhost", "127.0.0.1", 46501)
})
```

example_test.go
```diff
@@ -242,6 +242,54 @@ func ExampleMap_AddEntryListener() {
 	if err := m.RemoveEntryListener(ctx, subscriptionID); err != nil {
 		log.Fatal(err)
 	}
+
+	// You can also register event specific listeners.
```
yuce, line 246: Could you remove the old way of adding listeners from the example?

map_it_test.go
```diff
@@ -843,6 +905,39 @@ func TestMap_EntryNotifiedEventWithPredicate(t *testing.T) {
...
+		}
+		t.Logf("TestMap_EntryNotifiedEventWithPredicateWithSpecificHandler subscriptionID: %s", subID)
+		for i := 0; i < int(totalCallCount); i++ {
+			key := fmt.Sprintf("key-%d", i)
+			value := &it.SamplePortable{A: "foo", B: int32(i)}
+			it.MustValue(m.Put(context.Background(), key, value))
+		}
+		it.Eventually(t, func() bool {
+			cc := atomic.LoadInt32(&callCount)
+			t.Logf("call count target: %d, current: %d", totalCallCount, cc)
```
yuce, line 935: Could you add this log line to other similar tests as well?

proxy_map.go
```diff
@@ -1235,58 +1307,135 @@ func (as attributeSet) Attrs() []string {
...
+func (c *MapEntryListenerConfig) SetEntryInvalidatedListener(handler EntryNotifiedHandler) {
+	c.entryInvalidatedListener = handler
+}
+
+// SetEntryLoadedListener adds a listener to entry loaded events to the hazelcast.MapEntryListenerConfig.
+func (c *MapEntryListenerConfig) SetEntryLoadedListener(handler EntryNotifiedHandler) {
+	c.entryLoadedListener = handler
+}
+
+func (m *MapEntryListenerConfig) EventSpecificHandlersConfigured() bool {
```
yuce, line 1376: Let's unexport `EventSpecificHandlersConfigured`

yuce, line 1376: Let's unexport `EventSpecificHandlersConfigured`

yuce (changes requested):
Looks good, a few ideas...

map_it_test.go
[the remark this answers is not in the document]
srknzl: Should I change the older tests as well if they need a change?

srknzl:
I sent a [commit](https://github.com/hazelcast/hazelcast-go-client/pull/750/commits/b994d9316e8bff64e981025fc622b6e9b9672a98) to resolve all your comments unless you want [this](https://github.com/hazelcast/hazelcast-go-client/pull/750#discussion_r825918208) for existing tests as well.

srknzl:
Added @emreyigit as the second reviewer due to https://hazelcast.atlassian.net/wiki/spaces/HZC/pages/3871145993/Team+Process#Code-review-process

map_it_test.go
```diff
@@ -870,6 +969,36 @@ func TestMap_EntryNotifiedEventToKeyAndPredicate(t *testing.T) {
...
+			IncludeValue: true,
+			Key:          "k1",
+			Predicate:    predicate.Equal("A", "foo"),
+		}
+		listenerConfig.SetEntryAddedListener(handler)
+		if _, err := m.AddEntryListener(context.Background(), listenerConfig); err != nil {
+			t.Fatal(err)
+		}
+		it.MustValue(m.Put(context.Background(), "k1", &it.SamplePortable{A: "foo", B: 10}))
+		it.MustValue(m.Put(context.Background(), "k1", &it.SamplePortable{A: "bar", B: 10}))
```
emreyigit, line 992: I think that will trigger `Updated` not `Added` event right? That is why `callCount` is 1.

srknzl: Yes exactly

proxy_map.go
```diff
@@ -122,8 +122,80 @@ func (m *Map) NewLockContext(ctx context.Context) context.Context {
 }
 
 // AddEntryListener adds a continuous entry listener to this map.
-func (m *Map) AddEntryListener(ctx context.Context, config MapEntryListenerConfig, handler EntryNotifiedHandler) (types.UUID, error) {
-	return m.addEntryListener(ctx, config.flags, config.IncludeValue, config.Key, config.Predicate, handler)
+func (m *Map) AddEntryListener(ctx context.Context, config MapEntryListenerConfig, handler ...EntryNotifiedHandler) (types.UUID, error) {
```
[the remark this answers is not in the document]
srknzl: `handler` was a slice but we changed the API altogether. Basically, it was made a variadic parameter to allow both the old usage and the new one(in which we don't pass any handlers). You can check out the new API in the TDD. https://hazelcast.atlassian.net/wiki/spaces/HZC/pages/3997106177/Go+Client+Map+Event+Listener+Refactor+TDD

proxy_map.go
```diff
@@ -121,6 +122,35 @@ func (m *Map) NewLockContext(ctx context.Context) context.Context {
...
+	EntryExpired     func(event *EntryNotified)
+	EntryAllEvicted  func(event *EntryNotified)
+	EntryAllCleared  func(event *EntryNotified)
+	EntryMerged      func(event *EntryNotified)
+	EntryInvalidated func(event *EntryNotified)
+	EntryLoaded      func(event *EntryNotified)
+}
+
+func (m *Map) AddListener(ctx context.Context, listener MapListener, includeValue bool) (types.UUID, error) {
+	return types.NewUUID(), errors.New("s")
```
yuce, line 138: re: `errors.New("s")` is that a placeholder for a proper error?

srknzl: Yes, The implementation is not done as you can clearly see.

map_it_test.go
```diff
@@ -789,11 +794,61 @@ func TestMap_EntryNotifiedEvent(t *testing.T) {
 	})
 }
 
+func TestMap_EntryNotifiedEventWithAddListener(t *testing.T) {
+	it.MapTester(t, func(t *testing.T, m *hz.Map) {
+		const totalCallCount = int32(100)
+		callCount := int32(0)
+		subscriptionID, err := m.AddListener(context.Background(), hz.MapListener{
+			EntryAdded: func(event *hz.EntryNotified) {
+				if event.EventType == hz.EntryAdded {
```
yuce, line 803: How about calling `t.Fatalf("unexpected event type: %v", event.EventType)` if `EventType!= hz.EntryAdded`? That would get rid of the `else` block. Same comment for other similar code.

map_it_test.go
```diff
@@ -789,11 +794,61 @@ func TestMap_EntryNotifiedEvent(t *testing.T) {
...
+		for i := 0; i < int(totalCallCount); i++ {
+			key := fmt.Sprintf("key-%d", i)
+			value := fmt.Sprintf("value-%d", i)
+			it.MustValue(m.Put(context.Background(), key, value))
+		}
+		it.Eventually(t, func() bool {
+			return atomic.LoadInt32(&callCount) == totalCallCount
+		})
+		atomic.StoreInt32(&callCount, 0)
+
```
yuce, line 822: I guess a comment here would be more appropriate here than a  blank line.

map_it_test.go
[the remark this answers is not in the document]
yuce: That could be useful.

yuce (commented):
A few more ideas

example_test.go
```diff
@@ -209,19 +209,19 @@ func ExampleMap_AddEntryListener() {
...
-			fmt.Println("Entry Updated:", event.Value)
+			fmt.Println("Entry Updated:", event.Key)
 		// this is an entry evicted event
 		case hazelcast.EntryEvicted:
-			fmt.Println("Entry Remove:", event.Value)
+			fmt.Println("Entry Evicted:", event.Key)
 		// this is an entry loaded event
 		case hazelcast.EntryLoaded:
-			fmt.Println("Entry Loaded:", event.Value)
+			fmt.Println("Entry Loaded:", event.Key)
```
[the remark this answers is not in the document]
srknzl: Regarding the evicted, it was wrong, I fixed it. Regarding the conversion from Value to Key, I think printing the key is more reasonable to indicate that the entry with that key is evicted. Two entries' values can be the same but their key cannot be the same.

map_it_test.go
srknzl: A matter of styling I guess but ok I will do that

map_it_test.go
srknzl: Well, I went over this test again and did not understand why I put some entries to the map after removing the listener. Also, even if we do that the counter should not increase. I did not know how the test passes. Can you enlighten me if you understand the logic? If it is also weird to you I will remove this part:

```js
                // Clear the map to get entry added events
		err = m.Clear(context.Background())
		if err != nil {
			t.Fatal(err)
		}
		for i := 0; i < int(totalCallCount); i++ {
			key := fmt.Sprintf("key-%d", i)
			value := fmt.Sprintf("value-%d", i)
			it.MustValue(m.Put(context.Background(), key, value))
		}
		cc := atomic.LoadInt32(&callCount)
		if !assert.Equal(t, int32(0), cc) {
			t.Fatalf("call count target: %d, current: %d", totalCallCount, cc)
		}
```

map_it_test.go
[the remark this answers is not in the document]
srknzl: Ok changed other tests as well.

map_it_test.go
yuce: It's for making sure after removing the listener put operations do not fire events (hence remove listener is successful). What you have here seems correct to me, the counter doesn't increase and the test passes. But this line is misleading:
```
t.Fatalf("call count target: %d, current: %d", totalCallCount, cc)
```
It should be:
```
t.Fatalf("call count target: %d, current: %d", 0, cc)
```

proxy_map.go
```diff
@@ -950,7 +1052,39 @@ func (m *Map) addIndex(ctx context.Context, indexConfig types.IndexConfig) error
...
+//	}
+//	subscriptionID := types.NewUUID()
+//	addRequest := m.makeListenerRequest(keyData, predicateData, flags, includeValue)
+//	listenerHandler := func(msg *proto.ClientMessage) {
+//		m.makeListenerDecoder(msg, keyData, predicateData, m.mapListenerEventHandler(listener, methodName))
+//	}
+//	removeRequest := codec.EncodeMapRemoveEntryListenerRequest(m.name, subscriptionID)
+//	err = m.listenerBinder.Add(ctx, subscriptionID, addRequest, removeRequest, listenerHandler)
+//	return subscriptionID, err
+//}
```
yuce, line 1077: Did you forget to delete these lines?

proxy_map.go
```diff
@@ -950,7 +1052,39 @@ func (m *Map) addIndex(ctx context.Context, indexConfig types.IndexConfig) error
...
+
+func (m *Map) addEntryListener(
+	ctx context.Context,
+	flags int32,
+	includeValue bool,
+	key interface{},
+	predicate predicate.Predicate,
+	handler EntryNotifiedHandler,
+	methodName string,
```
yuce, line 1086: I don't like having to pass `methodName` just to put it in an error message. How about removing that and changing the error logging code to something like:
```
p.logger.Errorf("preparing entry notified event: %w", err)
```

srknzl: Changed it in fa35423

proxy_map.go
srknzl: Deleted in fa35423

proxy_map.go
```diff
@@ -38,43 +38,36 @@ Map is a distributed map.
...
-In the example below, a listener configuration for added and updated entries is created.
-Entries only with key "somekey" and matching to predicate year > 2000 are considered:
+To listen events of a map, you can use the AddListener, AddListenerWithKey, AddListenerWithPredicate and AddListenerWithPredicateAndKey methods.
+The first method adds a listener to the map's all events. The others filter the events depending on a key and/or a predicate.
+In all methods you specify whether you want to include value in the event or not.
+
+You can pass a MapListener struct to these methods to add handlers to different event types. You can add different handlers to different event types
+with a single MapListener struct. If you don't specify a handler in for an event type in MapListener struct, there will be no handler for that event.
+In the example below, a listener for added and updated entry events is created. Entries only with key "somekey" and matching to predicate year > 2000
+are considered:
```
yuce, line 50: Could you merge these lines to make them a single line?

proxy_map.go
```diff
@@ -38,43 +38,36 @@ Map is a distributed map.
...
-	entryListenerConfig.NotifyEntryAdded(true)
-	entryListenerConfig.NotifyEntryUpdated(true)
-	m, err := client.GetMap(ctx, "somemap")
-
-After creating the configuration, the second step is adding an event listener and a handler to act on received events:
 
-	subscriptionID, err := m.AddEntryListener(ctx, entryListenerConfig, func(event *hazelcast.EntryNotified) {
-		switch event.EventType {
-		case hazelcast.EntryAdded:
+	m, err := client.GetMap(ctx, "somemap")
```
yuce, line 58: Could you add a line to indicate error checking is missing, maybe something like:
```
// error checking is omitted.
```
Probably after also the `subscriptionID, err ...` line below.

proxy_map.go
```diff
@@ -121,7 +114,111 @@ func (m *Map) NewLockContext(ctx context.Context) context.Context {
...
+	}
+	return flags
+}
+
+func (m *Map) mapListenerEventHandler(listener MapListener) EntryNotifiedHandler {
+	return func(event *EntryNotified) {
+		switch event.EventType {
+		case EntryAdded:
+			listener.EntryAdded(event)
+			break
```
yuce, line 166: The `break`s in switch are not necessary.

proxy_map.go
```diff
@@ -121,7 +114,111 @@ func (m *Map) NewLockContext(ctx context.Context) context.Context {
...
+		case EntryLoaded:
+			listener.EntryLoaded(event)
+			break
+		case EntryAllCleared:
+			listener.MapCleared(event)
+			break
+		case EntryAllEvicted:
+			listener.MapEvicted(event)
+			break
+		}
```
yuce, line 191: Does it make sense to add a `default` branch here with a warning? Maybe something like:
```
m.logger.Warnf("unknown map event type: %d", event.EventType)
```

proxy_map.go
```diff
@@ -950,7 +1052,14 @@ func (m *Map) addIndex(ctx context.Context, indexConfig types.IndexConfig) error
...
 
-func (m *Map) addEntryListener(ctx context.Context, flags int32, includeValue bool, key interface{}, predicate predicate.Predicate, handler EntryNotifiedHandler) (types.UUID, error) {
+func (m *Map) addEntryListener(
+	ctx context.Context,
+	flags int32,
+	includeValue bool,
+	key interface{},
+	predicate predicate.Predicate,
+	handler EntryNotifiedHandler,
+) (types.UUID, error) {
```
yuce, line 1062: Does the IDE format each parameter on its own line? If not, it's OK to have the method signature on a single line.

yuce (changes requested):

proxy_map.go
srknzl: I guess I did it for readability. Reverting.

proxy.go
```diff
@@ -410,23 +410,38 @@ func (p *proxy) decodeEntryNotified(binKey, binValue, binOldValue, binMergingVal
...
 func (p *proxy) makeEntryNotifiedListenerHandler(handler EntryNotifiedHandler) entryNotifiedHandler {
-	return func(
-		binKey, binValue, binOldValue, binMergingValue iserialization.Data,
+	return func(binKey, binValue, binOldValue, binMergingValue iserialization.Data,
 		binEventType int32,
 		binUUID types.UUID,
-		affectedEntries int32) {
-		key, value, oldValue, mergingValue, err := p.decodeEntryNotified(binKey, binValue, binOldValue, binMergingValue)
+		affectedEntries int32,
+	) {
```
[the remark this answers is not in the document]
srknzl: Done in f24f3a5c5bc06be225ef32efe9dab80ae4141e46

proxy.go
```diff
@@ -410,23 +410,38 @@ func (p *proxy) decodeEntryNotified(binKey, binValue, binOldValue, binMergingVal
...
+		handler(event)
+	}
+}
+
+func (p *proxy) prepareEntryNotifiedEvent(
+	binKey, binValue, binOldValue, binMergingValue iserialization.Data,
+	binEventType int32,
+	binUUID types.UUID,
+	affectedEntries int32,
+) (*EntryNotified, error) {
```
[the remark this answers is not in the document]
srknzl: Done in f24f3a5c5bc06be225ef32efe9dab80ae4141e46

yuce (approved):
Looks good!

PR #753, opened 2022-03-19
Proxy manager shutdown [API-1259]

nevzatseferoglu:
When shutdown happens, already assigned proxies are destroyed and each destroy sends a signal to the server.

**Side effect**:
`ClientInternal` has to return a current proxy map to be tested.


internal/client/client.go
```diff
@@ -86,6 +86,22 @@ func (c *Config) Validate() error {
...
+	for _, f := range m {
+		err := f(ctx)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 type Client struct {
```
utku-caglayan, line 105: Can you run `make check` to see and fix alignment issues?

proxy_manager.go
```diff
@@ -222,3 +222,68 @@ func (m *proxyManager) getFlakeIDGeneratorConfig(name string) FlakeIDGeneratorCo
...
+		return fmt.Errorf("given proxy argument is nil")
+	}
+	errHandler := func(dsName string, i int, err error) error {
+		return fmt.Errorf("given %s proxy cannot be destroyed, last index: %d, %w", dsName, i, err)
+	}
+	for i, p := range proxyArgs {
+		if p == nil {
+			return fmt.Errorf("given %d. proxy argument is nil", i)
+		}
+		switch ds := p.(type) {
```
utku-caglayan, line 237: Do we need this type switch? I am not sure if we need to know the actual type. What about using a type assertion to an interface like
```
`type Destroyer interface {
    Destroy(...)...
    }
```

PR #757, opened 2022-03-29
Workflow for nightly build of Go client 

alisengul53:


.github/workflows/nightly_tests.yaml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          go-version: ${{ matrix.go_version }}
+          
+      - name: Checkout code
+        uses: actions/checkout@v2
+          
+      - name: Test
+        run: |
+          ./rc.sh start
+          env MEMBER_COUNT=3 make test-cover
```
yuce, line 32: It would be better to specify Hazelcast versions explicitly:
```
env HZ_VERSION=5.1 ./rc.sh start
env HZ_VERSION=5.1 MEMBER_COUNT=3 make test-cover
```

alisengul53: Added.
It is the same in the coverage report workflow, we should change it there I think but before changing it we can consider to use always latest release of the server. WDYT?

yuce (approved):

.github/workflows/nightly_tests.yaml
yuce: Thanks. We should change that as well.

utku-caglayan (approved):
This will be great thanks!

PR #759, opened 2022-04-01
SQL Driver Context Cancellation

utku-caglayan:
Currently fetch calls on sql.driver does not respect context cancellation that is passed earlier with `db.QueryContext`.

This work introduces a test that fails on that situation, and fixes regarding it.

### DIFF
--- a/sql_it_test.go
+++ b/sql_it_test.go
@@ -476,23 +476,40 @@ func sqlQueryWithCursorBufferSizeTest(t *testing.T) {
 
 func sqlStatementWithQueryTimeoutTest(t *testing.T) {
 	it.SkipIf(t, "hz < 5.0")
-	stmt := sql.NewStatement("select v from table(generate_stream(1))")
-	stmt.SetQueryTimeout(3 * time.Second)
-	it.Must(stmt.SetCursorBufferSize(2))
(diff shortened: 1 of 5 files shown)


nevzatseferoglu (approved):
LGTM!

sql_it_test.go
```diff
@@ -476,23 +476,40 @@ func sqlQueryWithCursorBufferSizeTest(t *testing.T) {
...
+			query: "select v from table(generate_stream(1))",
+		},
+		{
+			name:  "streaming query with no data, waiting for data/pages",
+			query: "select v from table(generate_stream(0))",
+		},
+	}
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			stmt := sql.NewStatement("select v from table(generate_stream(1))")
```
yuce, line 494: Shouldn't the statement created with `tc.query`?

yuce (dismissed):
Looks good, just a minor thing.

sql_it_test.go
utku-caglayan: Thanks for catching that https://github.com/hazelcast/hazelcast-go-client/pull/759/commits/0d0cbf7d9127c46262c759755dae517e1e19ecbb

yuce (approved):
Looks good!

PR #765, opened 2022-04-08
Restored serialization behaviour for time.Time values

yuce:
Previously, we used a heuristic to map `time.Time` values to the correct reference implementation type. This PR reverses that change, and uses the v1.1.1 behaviour.

### DIFF
--- a/client_internals_it_test.go
+++ b/client_internals_it_test.go
@@ -238,14 +238,6 @@ func TestClientInternal_OrderedMembers(t *testing.T) {
 	it.Eventually(t, func() bool {
 		return sameMembers(targetUUIDs, ci.OrderedMembers())
 	})
-	types.NewUUID()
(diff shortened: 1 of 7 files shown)


utku-caglayan (approved):
LGTM!

client_internals_it_test.go
```diff
@@ -238,14 +238,6 @@ func TestClientInternal_OrderedMembers(t *testing.T) {
 	it.Eventually(t, func() bool {
 		return sameMembers(targetUUIDs, ci.OrderedMembers())
 	})
-	types.NewUUID()
```
emreyigit, line 241: Is this related to the PR?

yuce: No, this is an accidental leftover from an earlier PR.

emreyigit (approved):

PR #768, opened 2022-04-13
Removed SQL Driver

yuce:


utku-caglayan (approved):
Looks good although would love to see this on the public API

Issue #774, opened 2022-05-04
Map partitions not getting distributed across all nodes


halfvector:
Go version: 1.18
Hazelcast Go Client version: 1.2.0
Hazelcast server version: 5.1.1
Number of clients: 1
Cluster size: 3

#### Expected behaviour
Even distribution of load across all nodes when doing a Put or Set on a Map.

<img width="407" alt="image" src="https://user-images.githubusercontent.com/1817604/166586779-6bebd030-be97-4ac7-a3bb-f73263a59a8d.png">

#### Actual behaviour
One node in the cluster gets every command dispatched to it.

<img width="405" alt="image" src="https://user-images.githubusercontent.com/1817604/166586896-01918d2d-013d-4a46-af9d-babc179e9b04.png">

The bug seems to be in the way ClusterView events are handled, where the UUID->Partitions map is decoded incorrectly, resulting in one UUID getting assigned to ALL partitions.

Easy way to observe this is by logging from `ConnectionManager.GetConnectionForPartition()` like this:
`m.logger.Infof("Partition ownership: partition_id=%v / uuid=%v", partitionID, uuid.String())`

It will show that every partition belongs to the same node.

PR #775, opened 2022-05-04
Fix all partitions being assigned to one node

halfvector:
Fixes https://github.com/hazelcast/hazelcast-go-client/issues/774

### Before:
3-node cluster, showing uneven activity by go-client:
image

3-node cluster, showing uneven Network and CPU utilization:
image

### After:
3-node cluster, showing even activity by go-client:
image

3-node cluster, showing even Network and CPU utilization:
image

Summary of root cause:
1. `DecodeListMultiFrameWithListInteger()` returns a single slice of same partitions for all nodes, instead of a slice of slices one for each node
2. `DecodeEntryListUUIDListInteger()` returns a n:m mapping of all uuids to all partitions, it should be a 1:k mapping of node to just its partitions.
3. `partitionTable.Update()` always assigns the last uuid in the pairs to all partitions, it should assign the correct node uuid to each partition

### DIFF
--- a/internal/cluster/partition_service.go
+++ b/internal/cluster/partition_service.go
@@ -118,12 +118,10 @@ func (p *partitionTable) Update(pairs []proto.Pair, version int32, connectionID 
 	}
 	newPartitions := map[int32]types.UUID{}
 	for _, pair := range pairs {
-		uuids := pair.Key.([]types.UUID)
+		uuid := pair.Key.(types.UUID)
 		ids := pair.Value.([]int32)
-		for _, uuid := range uuids {
-			for _, id := range ids {
-				newPartitions[id] = uuid
-			}
+		for _, id := range ids {
+			newPartitions[id] = uuid
 		}
 	}
 	if reflect.DeepEqual(p.partitions, newPartitions) {
--- a/internal/proto/codec/builtin.go
+++ b/internal/proto/codec/builtin.go
@@ -632,7 +632,11 @@ func DecodeListMultiFrameForData(frameIterator *proto.ForwardFrameIterator) []is
 }
 
 func DecodeListMultiFrameWithListInteger(frameIterator *proto.ForwardFrameIterator) [][]int32 {
-	result := make([][]int32, 0)
+	var result [][]int32
+	DecodeListMultiFrame(frameIterator, func(fi *proto.ForwardFrameIterator) {
+		result = append(result, DecodeListInteger(fi))
+	})
+	return result
 	frameIterator.Next()
 	for !CodecUtil.NextFrameIsDataStructureEndFrame(frameIterator) {
 		result = append(result, DecodeListInteger(frameIterator))
--- a/internal/proto/codec/builtin.go
+++ b/internal/proto/codec/builtin.go
@@ -366,7 +366,7 @@ func DecodeEntryListUUIDListInteger(frameIterator *proto.ForwardFrameIterator) [
 	keySize := len(keys)
 	result := make([]proto.Pair, keySize)
 	for i := 0; i < keySize; i++ {
-		result[i] = proto.NewPair(keys, values)
+		result[i] = proto.NewPair(keys[i], values[i])
 	}
 	return result
 }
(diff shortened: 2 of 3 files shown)


internal/proto/codec/builtin_test.go
```diff
@@ -586,10 +586,30 @@ func TestEntryListUUIDListIntegerCodec_Decode(t *testing.T) {
...
+	// when
+	result := DecodeEntryListUUIDListInteger(clientMessage.FrameIterator())
+
+	// then
+	assert.Equal(t, len(result), 2)
+	assert.Equal(t, result[0].Key.(types.UUID).String(), entries[0].Key.(types.UUID).String())
+	assert.EqualValues(t, result[0].Value.([]int32), entries[0].Value)
+	assert.Equal(t, result[1].Key.(types.UUID).String(), entries[1].Key.(types.UUID).String())
+	assert.EqualValues(t, result[1].Value.([]int32), entries[1].Value)
+}
```
halfvector, line 611: New test to cover the scenario of a multi-node cluster providing a map of partitions.

internal/cluster/partition_service.go
```diff
@@ -118,12 +118,10 @@ func (p *partitionTable) Update(pairs []proto.Pair, version int32, connectionID
 	}
 	newPartitions := map[int32]types.UUID{}
 	for _, pair := range pairs {
-		uuids := pair.Key.([]types.UUID)
+		uuid := pair.Key.(types.UUID)
 		ids := pair.Value.([]int32)
-		for _, uuid := range uuids {
-			for _, id := range ids {
-				newPartitions[id] = uuid
-			}
```
halfvector, line 126: This is where the last node-uuid would get assigned to all the partitions. Because each pair was given all 271 partitions instead of just the sub-set for the specific node.

yuce (dismissed):
This is a pretty big mistake on our part. Thanks for this PR. It looks great!

internal/proto/codec/builtin.go
```diff
@@ -631,11 +631,11 @@ func DecodeListMultiFrameForData(frameIterator *proto.ForwardFrameIterator) []is
 	return result
 }
 
-func DecodeListMultiFrameWithListInteger(frameIterator *proto.ForwardFrameIterator) []int32 {
-	result := make([]int32, 0)
+func DecodeListMultiFrameWithListInteger(frameIterator *proto.ForwardFrameIterator) [][]int32 {
+	result := make([][]int32, 0)
```
utku-caglayan, line 635: Great fix! It would better to use var initializer for default construction and also to use the existing `DecodeListMultiFrame` functionality iterate the frames.
```suggestion
	var result [][]int32
	DecodeListMultiFrame(frameIterator, func(fi *proto.ForwardFrameIterator) {
		result = append(result, DecodeListInteger(fi))
	})
	return result
```

internal/proto/codec/builtin_test.go
utku-caglayan: Thanks for this test! Do you think we can merge it the test above this one? It would be better to have a single one instead of separate ones.

utku-caglayan (changes requested):
Thanks a lot for identifying the issue and your contribution. I have attached few comments to discuss.

internal/proto/codec/builtin_test.go
halfvector: My pleasure!

From what I've seen in the code-base, each test ensures a single behavior. Are you sure you want me to break that pattern and merge two distinct test cases into a single function? (I'm happy to follow any convention, just double checking)

utku-caglayan: Oh you are right! I would say let's keep it as is. Resolving this

internal/proto/codec/builtin.go
utku-caglayan: Sorry, I believe I misplaced the suggestion. Can you refactor the method to eliminate unreachable code?

utku-caglayan (approved):
Looks great! Thanks.

yuce (approved):

PR #781, opened 2022-05-30
[TRIVIAL] Make ClientType and ClientVersion variables

yuce:
This PR makes `internal.ClientType` a variable, so it can be changed by a build flag.such as:
```
go build -v -ldflags="-X 'github.com/hazelcast/hazelcast-go-client/internal.ClientType=CLC' -X 'github.com/hazelcast/hazelcast-go-client/internal.ClientVersion=1.0.0-beta1'"

```

### DIFF
--- a/client.go
+++ b/client.go
@@ -35,7 +35,8 @@ import (
 )
 
 const (
-	ClientVersion = internal.ClientVersion
+	// ClientVersion is the semantic versioning compatible client version.
+	ClientVersion = internal.CurrentClientVersion
 )
 
 // StartNewClient creates and starts a new client with the default configuration.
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -35,7 +35,8 @@ import (
 )
 
 const (
-	ClientVersion = internal.ClientVersion
+	// ClientVersion is the semantic versioning compatible client version.
+	ClientVersion = internal.CurrentClientVersion
```
srknzl, line 39: Why don't we use `internal.ClientVersion` but use `internal.CurrentClientVersion` which cannot be changed via command line options?

yuce: Variables cannot be assigned to constants. And we have to use a variable in order to be able to change the version at build time.

srknzl (approved):

client.go
srknzl: What is the reason of having `ClientVersion` in client.go that is not used ?

srknzl: Ah ok it is public API to be used by the user. Ok

PR #784, opened 2022-06-01
Added back the SQL driver

yuce:
We've removed the SQL driver prior to v1.2.0 release. This PR adds back the support. Note that, the code in this PR adds previously approved changes back, there is no new code.


srknzl (approved):
Then, it is a no-brainer approval for me

PR #788, opened 2022-06-09
[DON'T MERGE] Fix 674 add ringbuffer support

nitram509:
* This PR just covers the Ringbuffer support.
* Recommendations from @srknzl and @yuce implemented
* codec implementation supported by @yuce 
* the filtering in ReadMany() is not supported, since proper codecs are missing - please comment if this is required.

Any feedback is welcome.

I'm reachable also via Hazelcast Community Slack as 'nitram509'.


srknzl:
@yuce Should not we implement filter support as well? Let's check what is missing in the codecs

yuce:
There is no missing codecs, but the filter is any value that implements `IdentifiedDataSerializable`. There should be a Java class on the server side for the filter. Our tests include a sample class that which can be used as the filter. Here's an example from the Python client tests: https://github.com/hazelcast/hazelcast-python-client/blob/master/tests/integration/backward_compatible/proxy/ringbuffer_test.py#L273

srknzl:
>  There is no missing codecs, but the filter is any value that implements IdentifiedDataSerializable. There should be a Java class on the server side for the filter. Our tests include a sample class that which can be used as the filter. Here's an example from the Python client tests: https://github.com/hazelcast/hazelcast-python-client/blob/master/tests/integration/backward_compatible/proxy/ringbuffer_test.py#L273



Can't be a portable or a compact as well?

nitram509:
@yuce @srknzl I do not fully understand how the prefix filter classes in Hazelcast are serialized, but it seems to me special to Hazelcast and does not match ProtoBuf, TLV, or other protocols I'm aware of.

Since the Python implementation is also not yet complete (see https://github.com/hazelcast/hazelcast-python-client/blob/18326a76f1c38564a4dab4332cc194c4e83ab483/hazelcast/serialization/api.py#L476)
would you be fine with agile-like splitting this feature as well?
This means, that this PR covers basic Ringbuffer support, and a follow-up ticket could cover advanced filtering.
What do you think?

nitram509:
Meanwhile, I did some research on the Python client and the Java client.

I must admit, I'm a bit confused. In the Python client, theres an embedded class (see Yüce's link above),
which contains a PrefixFilter with ClassID=14.
In the Java client, there's a PrefixFilter as well, but with ClassID=4.
Also, the Java client contains a StartsWithStringFilter, with ClassID=14.

About the mismatch I'm confused.
Also, I wonder, why In all 3 cases, these filters are only present in the TEST code.
I would assume the filter classes would be perfectly re-usable.
Why are they not present in the regular src content folder?

Overall, it seems to me, there's no simple way forward.
And I would love to have at least basic Ringbuffer support in the Go client.
May I propose to keep the scope of this PR "small"?

That said, I realized, I missed writing some example source file ... I will add such quickly.
Are there other guidelines I should follow to support in reviewing this PR?

yuce:
@nitram509 The filters can only be written in Java, since they are executed on the server-side. But it's OK to leave the tests out in this PR. We will add it in another one. Thanks a lot for your contribution.

examples/ringbuffer/main.go
```diff
@@ -0,0 +1,58 @@
...
+	// Add an item to the Ringbuffer
+	sequence, err := rb.Add(ctx, "item 1", hazelcast.OverflowPolicyOverwrite)
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Printf("Added item 1 with sequence number=%d\n", sequence)
+
+	// Get an item by a known sequence number
+	item, err := rb.ReadOne(ctx, sequence)
+	if err != nil {
```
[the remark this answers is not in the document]
utku-caglayan: An example that checks https://github.com/hazelcast/hazelcast-go-client/blob/master/hzerrors/errors.go#L77 this would b great!

internal/it/ring_buffer.go
```diff
@@ -0,0 +1,82 @@
...
+	"testing"
+
+	"go.uber.org/goleak"
+
+	hz "github.com/hazelcast/hazelcast-go-client"
+)
+
+func RingbufferTester(t *testing.T, f func(t *testing.T, rb *hz.Ringbuffer)) {
+	makeListName := func() string {
+		return NewUniqueObjectName("list")
```
utku-caglayan, line 30:
```suggestion
		return NewUniqueObjectName("ringbuffer")
```

internal/it/ring_buffer.go
```diff
@@ -0,0 +1,82 @@
...
+		config := defaultTestCluster.DefaultConfig()
+		if cbCallback != nil {
+			cbCallback(&config)
+		}
+		config.Cluster.Unisocket = !smart
+		client, rb = getClientRingbufferWithConfig(ringBufferName(), &config)
+		defer func() {
+			ctx := context.Background()
+			if err := rb.Destroy(ctx); err != nil {
+				t.Logf("test warning, could not destroy list: %s", err.Error())
```
utku-caglayan, line 55:
```suggestion
				t.Logf("test warning, could not destroy ringbuffer: %s", err.Error())
```

internal/it/ring_buffer.go
```diff
@@ -0,0 +1,82 @@
...
+	"context"
+	"testing"
+
+	"go.uber.org/goleak"
+
+	hz "github.com/hazelcast/hazelcast-go-client"
+)
+
+func RingbufferTester(t *testing.T, f func(t *testing.T, rb *hz.Ringbuffer)) {
+	makeListName := func() string {
```
utku-caglayan, line 29:
```suggestion
	makeRingbufferName := func() string {
```

proxy_ring_buffer.go
```diff
@@ -0,0 +1,266 @@
...
+
+// A Ringbuffer is a data structure where the content is stored in a ring-like
+// structure. A Ringbuffer has a fixed capacity, so it won't grow beyond
+// that capacity and endanger the stability of the system. If that capacity
+// is exceeded, the oldest item in the Ringbuffer is overwritten.
+//
+// For details see https://docs.hazelcast.com/imdg/latest/data-structures/ringbuffer
+type Ringbuffer struct {
+	*proxy
+	partitionID int32
```
utku-caglayan, line 35: It would be great to store the capacity of Ringbuffer as in Java client. 

https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/client/impl/proxy/ClientRingbufferProxy.java#L68

It makes a "lazy-set" for that field

proxy_ring_buffer.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		return ReadResultSetSequenceUnavailable, err
+	}
+	return codec.DecodeRingbufferAddAllResponse(response), nil
+}
+
+// ReadOne
+// Reads one item from the Ringbuffer. If the sequence is one beyond the current tail, this call blocks until an
+// item is added. This method is not destructive unlike e.g. a queue.take. So the same item can be read by multiple
+// readers, or it can be read multiple times by the same reader. Currently, it isn't possible to control how long this
```
utku-caglayan, line 143: I would love to discuss this point. The context parameter provides a way to pass a value on how much it will block. Context cancelation and timeout will result in an error, but it is doable for the user.

Do you guys agree? If yes, how can we document this 🤔

proxy_ring_buffer.go
```diff
@@ -0,0 +1,266 @@
...
+
+// ReadCount Number of items that have been read before filtering.
+func (rrs *ReadResultSet) ReadCount() int32 {
+	return rrs.readCount
+}
+
+// Get one item from List of items that have been read.
+func (rrs *ReadResultSet) Get(index int) (interface{}, error) {
+	if index < 0 || int32(index) >= rrs.readCount {
+		return nil, errors.New(fmt.Sprintf("index out of range [%d] with length %d", index, rrs.readCount))
```
utku-caglayan, line 245:
```suggestion
		return nil, hzcerrors.NewIllegalArgumentError(fmt.Sprintf("index out of range [%d] with length %d", index, rrs.readCount), nil)
```
We should return IllegalArgumentError to comply with the Java client
https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/ReadResultSetImpl.java#L143

utku-caglayan (changes requested):
Thanks for the collaboration! Leaving some notes to discuss.

examples/ringbuffer/main.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	// Add an item to the Ringbuffer
+	sequence, err := rb.Add(ctx, "item 1", hazelcast.OverflowPolicyOverwrite)
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Printf("Added item 1 with sequence number=%d\n", sequence)
+
+	// Get an item by a known sequence number
+	item, err := rb.ReadOne(ctx, sequence)
```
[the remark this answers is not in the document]
nitram509: will do.

examples/ringbuffer/main.go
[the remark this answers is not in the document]
nitram509: will do.

internal/it/ring_buffer.go
nitram509: Good catch!

internal/it/ring_buffer.go
nitram509: Good Catch.

examples/ringbuffer/main.go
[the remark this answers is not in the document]
nitram509: Done.

examples/ringbuffer/main.go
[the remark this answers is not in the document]
nitram509: added in the example.

proxy_ring_buffer.go
nitram509: will do.

nitram509: Added as suggested.

proxy_ring_buffer.go
nitram509: Handling timeouts explicitly I consider best practice ... I do suggest adding an example.

nitram509: I did add some example code
2bdad54129f947b5b8a7b4d2c80d7a185ac69878
What do you think?

utku-caglayan: Thanks for the example! I think it looks good but I'm not sure if we should support it with context since other clients do not have such a mechanism yet. I would love to hear other reviewers' thoughts as well on this. If that's okay for you, can you leave this conversation as "unresolved" for now?

proxy_ring_buffer.go
```diff
@@ -0,0 +1,266 @@
...
+// this id is not the sequence of the item you are about to publish but from a previously published item. So it can't be used
+// to find that item.
+//
+// Add returns the sequence number of the added item. You can read the added item using this number.
+func (rb *Ringbuffer) Add(ctx context.Context, item interface{}, overflowPolicy OverflowPolicy) (sequence int64, err error) {
+	elementData, err := rb.validateAndSerialize(item)
+	if err != nil {
+		return ReadResultSetSequenceUnavailable, err
+	}
+	request := codec.EncodeRingbufferAddRequest(rb.name, int32(overflowPolicy), elementData)
```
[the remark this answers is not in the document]
srknzl: How do we validate enum like values in go client? We can check if `overflowPolicy` is in range [0, 1], but we need to update the valid range as well if a new overflowPolicy is added in the future.

internal/it/ring_buffer.go
```diff
@@ -0,0 +1,82 @@
...
+	"go.uber.org/goleak"
+
+	hz "github.com/hazelcast/hazelcast-go-client"
+)
+
+func RingbufferTester(t *testing.T, f func(t *testing.T, rb *hz.Ringbuffer)) {
+	makeRingbufferName := func() string {
+		return NewUniqueObjectName("list")
+	}
+	RingbufferTesterWithConfigAndName(t, makeListName, nil, f)
```
srknzl, line 32:
```suggestion
	RingbufferTesterWithConfigAndName(t, makeRingbufferName, nil, f)
```

proxy_ring_buffer.go
srknzl: > The context parameter provides a way to pass a value on how much it will block. Context cancelation and timeout will result in an error, but it is doable for the user.

What do you mean by an error? So it is possible to set timeouts and cancel the API call in go by default ? Well, I did not understand the problem. I don't think there will be any harm of supporting them with context because we have context (as far as I know) in all async APIs in the go client

proxy_ring_buffer.go
```diff
@@ -0,0 +1,271 @@
...
+// ReadCount Number of items that have been read before filtering.
+func (rrs *ReadResultSet) ReadCount() int32 {
+	return rrs.readCount
+}
+
+// Get one item from List of items that have been read.
+func (rrs *ReadResultSet) Get(index int) (interface{}, error) {
+	if index < 0 || int32(index) >= rrs.readCount {
+		return nil, hzcerrors.NewIllegalArgumentError(fmt.Sprintf("index out of range [%d] with length %d", index, rrs.readCount), nil)
+	}
```
srknzl, line 251: What about extracting this logic into a method like Java code? Also you return NewIllegalArgumentError in this snippet whereas you return an error with errors.New below

internal/it/ring_buffer.go
nitram509: Done, see 62089238940f24f70e8c300cf5f265c2e3a91542

proxy_ring_buffer.go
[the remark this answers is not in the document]
nitram509: Go does not support ENUMs as we know from Java.
I did choose for the closest, what is typically used in Go: a type + some constants.
See https://github.com/nitram509/hazelcast-go-client/blob/62089238940f24f70e8c300cf5f265c2e3a91542/proxy_ring_buffer.go#L66

This approach guarantees type safety and thus the compiler checks for values.
Also, this approach can be used in the client code generators. ... similar to enums, such Go snippets can be created.
Then future extensions would be safe.

@utku-caglayan on your concrete example (line 110), in the method signature, is using the type already ... so we have some safety at compile time.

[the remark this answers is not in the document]
srknzl: I did not understand the type safety provided here. This is valid code: 

```
sequence, err := rb.Add(ctx, "item 1", 123)
```

proxy_ring_buffer.go
srknzl: Well, actually `ReadResultSet` and its `get(int index)` method are public as well. I think it is ok to return a `IllegalArgumentError` here as a validation error. If you extract these checks into a single method, regardless of the error type, they will be consistent.

proxy_ring_buffer.go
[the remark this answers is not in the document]
nitram509: Oops, you're right. I was fallen under the wrong assumption of type safety here,
but I realized type conversion happens :/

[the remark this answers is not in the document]
nitram509: Serkan, thanks for pointing out the missing type-safety.
On a more general level...

Would you consider following the idea of using type safety, by e.g. using interfaces?
This way, one could not provide an arbitrary int value.

Or would you rather consider explicit range checks and no further implementations to enforce type checks?

I could provide an example if you want?

proxy_ring_buffer.go
nitram509: Will provide some.

nitram509: Done. See 255eed35975866dac4defbee0434c423c6dce367

proxy_ring_buffer.go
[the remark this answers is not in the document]
srknzl: I think @yuce will know what to do better than me. I don't know what we do in the client code and not so familiar with golang as well.

examples/ringbuffer/main.go
```diff
@@ -0,0 +1,90 @@
...
+		log.Fatal(err)
+	}
+	// Get a random Ringbuffer
+	rand.Seed(time.Now().Unix())
+	queueName := fmt.Sprintf("sample-%d", rand.Int())
+	rb, err := client.GetRingbuffer(ctx, queueName)
+	if err != nil {
+		log.Fatal(err)
+	}
+
```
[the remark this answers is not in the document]
nitram509: Done.
See 1342a57f9015edaa470f487a6daec44ea8e6fe2d as well.

proxy_ring_buffer.go
```diff
@@ -0,0 +1,266 @@
...
+
+func newRingbuffer(p *proxy) (*Ringbuffer, error) {
+	if partitionID, err := p.stringToPartitionID(p.name); err != nil {
+		return nil, err
+	} else {
+		return &Ringbuffer{proxy: p, partitionID: partitionID}, nil
+	}
+}
+
+// Add an item to the tail of the Ringbuffer. If there is space in the Ringbuffer, the call
```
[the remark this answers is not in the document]
nitram509: Can we resolve this comment?

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+// For details see https://docs.hazelcast.com/hazelcast/latest/data-structures/ringbuffer
+type Ringbuffer struct {
+	*proxy
+	partitionID int32
+	capacity    int64
+}
+
+// OverflowPolicy
+// Using this OverflowPolicy one can control the behavior what should to be done
+// when an item is about to be added to the RingBuffer, but there is {@code 0}
```
srknzl, line 42: Will this {@code 0} work with godoc?  Also make `RingBuffer`s Ringbuffer for consistency. There are 3 of them in this file. Also there are 3 `ringbuffer`s

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+	return codec.DecodeRingbufferAddResponse(response), nil
+}
+
+// AddAll
+// Adds all the items of a collection to the tail of the Ringbuffer. A addAll is likely to outperform multiple calls
+// to add(Object) due to better io utilization and a reduced number of executed operations. If the batch is empty,
+// the call is ignored. When the collection is not empty, the content is copied into a different data-structure.
+// This means that: after this call completes, the collection can be re-used. the collection doesn't need to be serializable.
+// If the collection is larger than the capacity of the Ringbuffer, then the items that were written first will be
+// overwritten. Therefor this call will not block. The items are inserted in the order of the Iterator of the collection.
```
srknzl, line 126:
```suggestion
// overwritten. Therefore this call will not block. The items are inserted in the order of the Iterator of the collection.
```

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+// If an addAll is executed concurrently with an add or addAll, no guarantee is given that items are contiguous.
+// The result of the future contains the sequenceId of the last written item.
```
srknzl, line 128: Why did you remove the spaces between paragraphs here? I think Java's version looks better: 

image

Also there is no future in this API. So the sentence `The result of the future contains the sequenceId of the last written item.` needs to be changed

proxy_ring_buffer.go
srknzl, line 126: > The items are inserted in the order of the Iterator of the collection

This looks like a Java specific sentence. In general can you refactor whole comment so that it does not mention a collection but a slice maybe?

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+		return ReadResultSetSequenceUnavailable, err
+	}
+	return codec.DecodeRingbufferAddAllResponse(response), nil
+}
+
+// ReadOne
+// Reads one item from the Ringbuffer. If the sequence is one beyond the current tail, this call blocks until an
+// item is added. This method is not destructive unlike e.g. a queue.take. So the same item can be read by multiple
+// readers, or it can be read multiple times by the same reader. Currently, it isn't possible to control how long this
+// call is going to block. In the future we could add e.g. tryReadOne(long sequence, long timeout, TimeUnit unit).
```
srknzl, line 146: `tryReadOne(long sequence, long timeout, TimeUnit unit)` part is Java specific. Let's remove the sentence or change the signature to be go compatible.

srknzl (commented):
I will continue reviewing, left some comments

proxy_ring_buffer.go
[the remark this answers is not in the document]
yuce: Since this value is directly sent to the server-side, it would be better to check it. I think it's enough to check whether the policy is either `OverflowPolicyOverwrite` or `OverflowPolicyFail`, which are the only valid values. If they are not valid, `ErrIllegalArgument` may be returned, something like:
```
if int(overflowPolicy) != 0 || int(overflowPolicy) != 1 {
	msg := fmt.Sprintf("invalid overflow policy: %d", overflowPolicy)
	return 0, hzcerrors.NewIllegalArgumentError(msg, nil)	
}
```
I had to cast the `overflowPolicy` to `int`, otherwise GoLand is trying to simplify the condition, and replaces it with `true`.

proxy_ring_buffer.go
yuce: Go client supports context cancellation and timeouts natively, which is different from other clients. I think we can just remove the "Currently, it isn't possible to control how long this  call is going to block. In the future we could add e.g. tryReadOne(long sequence, long timeout, TimeUnit unit)." part from the documentation.

examples/ringbuffer/main.go
```diff
@@ -0,0 +1,87 @@
...
+		log.Fatal(err)
+	}
+	// Get a random Ringbuffer
+	rand.Seed(time.Now().Unix())
+	queueName := fmt.Sprintf("sample-%d", rand.Int())
+	rb, err := client.GetRingbuffer(ctx, queueName)
+	if err != nil {
+		log.Fatal(err)
+	}
+	// *** Example adding and retrieving single items
```
yuce, line 45: I guess 3 asterisks were left by mistake?

internal/it/ring_buffer.go
```diff
@@ -0,0 +1,82 @@
...
+	"context"
+	"testing"
+
+	"go.uber.org/goleak"
+
+	hz "github.com/hazelcast/hazelcast-go-client"
+)
+
+func RingbufferTester(t *testing.T, f func(t *testing.T, rb *hz.Ringbuffer)) {
+	name := func() string {
```
yuce, line 29: How about `makeName` instead of `name`?

internal/it/ring_buffer.go
```diff
@@ -0,0 +1,82 @@
...
+		})
+	}
+}
+
+func getClientRingbufferWithConfig(name string, config *hz.Config) (*hz.Client, *hz.Ringbuffer) {
+	client := getDefaultClient(config)
+	if l, err := client.GetRingbuffer(context.Background(), name); err != nil {
+		panic(err)
+	} else {
+		return client, l
```
yuce, line 80: Could you convert this code to:
```
l, err := client.GetRingbuffer(context.Background(), name)
if err != nil {
		panic(err)
}
return client, l
```

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+ * limitations under the License.
+ */
+
+package hazelcast
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	hzcerrors "github.com/hazelcast/hazelcast-go-client/internal/hzerrors"
```
yuce, line 23: By convention, we add prefix `i` to the internal import aliases (just like the iserialization import below). So this import would become:
```
ihzerrors "github.com/hazelcast/hazelcast-go-client/internal/hzerrors"
```

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/hazelcast/hazelcast-go-client/internal/proto/codec"
+	iserialization "github.com/hazelcast/hazelcast-go-client/internal/serialization"
+)
+
+// A Ringbuffer is a data structure where the content is stored in a ring-like
+// structure. A Ringbuffer has a fixed capacity, so it won't grow beyond
+// that capacity and endanger the stability of the system. If that capacity
+// is exceeded, the oldest item in the Ringbuffer is overwritten.
```
yuce, line 31: We try to put each sentence on its own line, so each line ends with a period. Would it be possible for you to adjust this documentation and others for that convention?

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+// Overflowing happens when a time-to-live is set and the oldest item in
+// the RingBuffer (the head) is not old enough to expire.
+type OverflowPolicy int
+type ReadResultSet struct {
+	rb        *Ringbuffer
+	readCount int32
+	items     []iserialization.Data
+	itemSeqs  []int64
+	nextSeq   int64
+}
```
yuce, line 54: Could you move `ReadResultSet` to somewhere before or after `OverflowPolicy`, so `OverflowPolicy` type and constants are consecutive?

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+	// since it doesn't trap the thread due to backoff.
+	//
+	// Example: if there is a time-to-live of 30 seconds, the buffer is full
+	// and the oldest item in the ring has been placed a second ago, then there
+	// are 29 seconds remaining for that item. Using this policy you are not
+	// going to overwrite that item for the next 29 seconds.
+	OverflowPolicyFail OverflowPolicy = 1
+
+	// ReadResultSetSequenceUnavailable is used when error happened
+	ReadResultSetSequenceUnavailable int64 = -1
```
yuce, line 83: Could you move this constant to its own block next to `ReadResultSet` type?

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+)
+
+func newRingbuffer(p *proxy) (*Ringbuffer, error) {
+	partitionID, err := p.stringToPartitionID(p.name)
+	if err != nil {
+		return nil, err
+	}
+	return &Ringbuffer{proxy: p, partitionID: partitionID, capacity: ReadResultSetSequenceUnavailable}, nil
```
yuce, line 91: Could you layout this so each field is in its own line? Like:
```
rb := &Ringbuffer{
    proxy: p,
    partitionID: partitionID,
...
}
return rb, nil
```

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+// ReadOne
+// Reads one item from the Ringbuffer. If the sequence is one beyond the current tail, this call blocks until an
+// item is added. This method is not destructive unlike e.g. a queue.take. So the same item can be read by multiple
+// readers, or it can be read multiple times by the same reader. Currently, it isn't possible to control how long this
+// call is going to block. In the future we could add e.g. tryReadOne(long sequence, long timeout, TimeUnit unit).
+func (rb *Ringbuffer) ReadOne(ctx context.Context, sequence int64) (interface{}, error) {
+	request := codec.EncodeRingbufferReadOneRequest(rb.name, sequence)
+	response, err := rb.invokeOnPartition(ctx, request, rb.partitionID)
+	if err != nil {
+		return ReadResultSetSequenceUnavailable, err
```
yuce, line 151: How about `return nil, err` on this line? Same comment for other lines which have `return ReadResultSetSequenceUnavailable, err`

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+// than the maxCount, these items are returned. So it could be the number of items read is smaller than the maxCount.
+// If there are fewer items available than minCount, then this call blacks. Reading a batch of items is likely to
+// perform better because less overhead is involved. A filter can be provided to only select items that need to be read.
+// If the filter is null, all items are read. If the filter is not null, only items where the filter function returns
+// true are returned. Using filters is a good way to prevent getting items that are of no value to the receiver.
+// This reduces the amount of IO and the number of operations being executed, and can result in a significant performance improvement.
+func (rb *Ringbuffer) ReadMany(ctx context.Context, startSequence int64, minCount int32, maxCount int32, filter interface{}) (ReadResultSet, error) {
+	if filter != nil {
+		return ReadResultSet{}, errors.New("filter functions are not yet supported in th Go client, please use 'nil' for now")
+	}
```
yuce, line 226: This is not correct. Could you remove this check?

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+	return rrs.rb.convertToObject(rrs.items[index])
+}
+
+// GetSequence one sequence number from List of sequence numbers for the items that have been read.
+func (rrs *ReadResultSet) GetSequence(index int) (int64, error) {
+	if err := rrs.validateIndexInRange(index); err != nil {
+		return ReadResultSetSequenceUnavailable, err
+	}
+	return rrs.itemSeqs[index], nil
```
yuce, line 260: Could you range check the `index` and return an illegal argument error with something like: `ihzerrors.NewIllegalArgumentError(fmt.Sprintf("index %d not in range 0 to %d", index, rrs.Size()))`

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+func (rrs *ReadResultSet) GetNextSequenceToReadFrom() int64 {
+	return rrs.nextSeq
+}
+
+func (rrs *ReadResultSet) validateIndexInRange(index int) error {
+	if index < 0 || int32(index) >= rrs.readCount {
+		return hzcerrors.NewIllegalArgumentError(fmt.Sprintf("index out of range [%d] with length %d", index, rrs.readCount), nil)
+	}
+	return nil
+}
```
yuce, line 278: You can use `check.WithinRangeInt32(index, 0, rrs.readCount - 1)` instead.

ring_buffer_it_test.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func TestRingbuffer_ReadMany_Get_with_startSequence(t *testing.T) {
+	it.RingbufferTester(t, func(t *testing.T, rb *hz.Ringbuffer) {
+		rb.AddAll(context.Background(), hz.OverflowPolicyOverwrite, "x", "0", "1", "2", "y", "z")
+
+		rs, err := rb.ReadMany(context.Background(), 1, 3, 3, nil)
+		assert.NoError(t, err)
+
+		item, _ := rs.Get(0)
```
yuce, line 120: It's better not to ignore the errors. You can use `it.MustValue`, like:
```
item := it.MustValue(rs.Get(0))
```
Similar comment to test code where the error is ignored.

yuce (changes requested):
Thanks again for your contribution! Here are some ideas...

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+	request := codec.EncodeRingbufferRemainingCapacityRequest(rb.name)
+	response, err := rb.invokeOnPartition(ctx, request, rb.partitionID)
+	if err != nil {
+		return ReadResultSetSequenceUnavailable, err
+	}
+	return codec.DecodeRingbufferRemainingCapacityResponse(response), nil
+}
+
+// ReadMany
+// Reads a batch of items from the Ringbuffer. If the number of available items after the first read item is smaller
```
srknzl, line 216: Similar comment about whitespacing in apidocumentation. I think in the IDE it does not look good in when it is one paragraph:

image

If this is required in go, ignore me

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+	return ReadResultSet{
+		rb:        rb,
+		readCount: readCount,
+		items:     items,
+		itemSeqs:  itemSeqs,
+		nextSeq:   nextSeq,
+	}, nil
+}
+
+// ReadCount Number of items that have been read before filtering.
```
srknzl, line 242: Again, I think we should write the whole version here.

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+func (rrs *ReadResultSet) ReadCount() int32 {
+	return rrs.readCount
+}
+
+// Get one item from List of items that have been read.
+func (rrs *ReadResultSet) Get(index int) (interface{}, error) {
+	if err := rrs.validateIndexInRange(index); err != nil {
+		return nil, err
+	}
+	return rrs.rb.convertToObject(rrs.items[index])
```
srknzl, line 252: In TDD we discussed lazy deserialization and we don't want any lazy deserialization for now. So, I guess you can deserialize all the items in `ReadMany()` cc @yuce @utku-caglayan

proxy_ring_buffer.go
srknzl: It is already done in `validateIndexInRange` but it should use `size` not `readCount`

srknzl: Also there is another check in Java side here:

`return seqs.length > index ? seqs[index] : -1;`

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+
+// GetSequence one sequence number from List of sequence numbers for the items that have been read.
+func (rrs *ReadResultSet) GetSequence(index int) (int64, error) {
+	if err := rrs.validateIndexInRange(index); err != nil {
+		return ReadResultSetSequenceUnavailable, err
+	}
+	return rrs.itemSeqs[index], nil
+}
+
+// Size the total size of
```
srknzl, line 263: of what?

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+func (rrs *ReadResultSet) Size() int {
+	return len(rrs.items)
+}
+
+// GetNextSequenceToReadFrom sequence number of the item following the last read item.
```
srknzl, line 268: I prefer more details here.

srknzl (commented):
Some more comments. Only the test file left to review for me

ring_buffer_it_test.go
```diff
@@ -0,0 +1,187 @@
...
+		assert.Error(t, err)
+	})
+}
+
+func TestRingbuffer_AddAll(t *testing.T) {
+	it.RingbufferTester(t, func(t *testing.T, rb *hz.Ringbuffer) {
+		_, err := rb.AddAll(context.Background(), hz.OverflowPolicyOverwrite, "foo", "bar")
+		assert.NoError(t, err)
+
+		foo, err := rb.ReadOne(context.Background(), 0)
```
srknzl, line 53: Can you add tailSequence related lines here as well?

```java
public void addAll() throws Exception {
        Future<Long> f = clientRingbuffer.addAllAsync(asList("foo", "bar"), OVERWRITE).toCompletableFuture();
        Long result = f.get();

        assertEquals(new Long(serverRingbuffer.tailSequence()), result);
        assertEquals("foo", serverRingbuffer.readOne(0));
        assertEquals("bar", serverRingbuffer.readOne(1));
        assertEquals(0, serverRingbuffer.headSequence());
        assertEquals(1, serverRingbuffer.tailSequence());
    }
```

ring_buffer_it_test.go
```diff
@@ -0,0 +1,187 @@
...
+		assert.NoError(t, err)
+		assert.GreaterOrEqual(t, capacity, int64(9999), "There should be a capacity of at least 9999 items.")
+	})
+}
+
+func TestRingbuffer_RemainingCapacity(t *testing.T) {
+	it.RingbufferTester(t, func(t *testing.T, rb *hz.Ringbuffer) {
+		capacity, err := rb.RemainingCapacity(context.Background())
+		assert.NoError(t, err)
+		assert.GreaterOrEqual(t, capacity, int64(9999), "There should be a remaining capacity of at least 9999 items.")
```
srknzl, line 85: Where does this number come from?  Same for capacity. What is the default capacity?

ring_buffer_it_test.go
```diff
@@ -0,0 +1,187 @@
...
+		_, err = rs.GetSequence(-1)
+		assert.Error(t, err)
+
+		_, err = rs.GetSequence(3)
+		assert.Error(t, err)
+
+		_, err = rs.GetSequence(2)
+		assert.NoError(t, err)
+	})
+}
```
srknzl, line 187: We can also write some more tests in RingbufferTest.java like 

* readManyAsync_maxCount
* readOne_whenHitsStale_shouldNotBeBlocked
* readManyAsync_whenStartSequenceIsJustBeyondTailSequence and readManyAsync_whenStartSequenceIsJustBeyondTailSequence using context timeout
* readManyAsync_whenStartSequenceIsEqualToTailSequence
* readManyAsync_whenStartSequenceIsNegative

srknzl (changes requested):

ring_buffer_it_test.go
nitram509: Done. See f82945d3a0b32b6cf18f016f0fec68e5bf368675

ring_buffer_it_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"github.com/stretchr/testify/assert"
+
+	hz "github.com/hazelcast/hazelcast-go-client"
+	"github.com/hazelcast/hazelcast-go-client/internal/it"
+)
+
+func TestRingbuffer_Add(t *testing.T) {
+	it.RingbufferTester(t, func(t *testing.T, rb *hz.Ringbuffer) {
+		sequence, err := rb.Add(context.Background(), "foo", hz.OverflowPolicyOverwrite)
+		assert.NoError(t, err)
```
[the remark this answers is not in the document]
nitram509: Great suggestion.
Replace allmost all occurences with it.MustValue()
See d8a9903c30945bf91fff78a4c5fbda668c942bc3

ring_buffer_it_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	hz "github.com/hazelcast/hazelcast-go-client"
+	"github.com/hazelcast/hazelcast-go-client/internal/it"
+)
+
+func TestRingbuffer_Add(t *testing.T) {
```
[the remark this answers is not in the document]
nitram509: Just adding my 2$cent here ... the Go code seems to have a convention "$topic_$MethodName" where $MethodName is the method under test. I guess we should keep this in favor of this Go code convention, no?

proxy_ring_buffer.go
nitram509: Done.
see c35d9321fd0ac04117ea9fa620e81f205bcaefb5

proxy_ring_buffer.go
nitram509: Made consisten naming Ringbuffer.
ef100407591a499e28efa1f4619f5b9cd61fee6c

proxy_ring_buffer.go
nitram509: HTML is not supported in Godocs, hence I dropped them.
Also, I did copy almost all comments in ```proxy_ring_buffer.go``` from https://github.com/hazelcast/hazelcast-client-protocol/blob/master/protocol-definitions/Ringbuffer.yaml
So I guess, this the wording there should be updated as well.

nitram509: I did drop the 'future' comment.
See 8aff861649746e6dd66f80828011542e2d8f73d5

nitram509: @srknzl  Is there anything else you would like to see/changed?

proxy_ring_buffer.go
nitram509: Done.
ff4bf6521c968ebf1ce850a4ef29a818c7c156e7

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+// Adds all the items of a collection to the tail of the Ringbuffer. A addAll is likely to outperform multiple calls
+// to add(Object) due to better io utilization and a reduced number of executed operations. If the batch is empty,
+// the call is ignored. When the collection is not empty, the content is copied into a different data-structure.
+// This means that: after this call completes, the collection can be re-used. the collection doesn't need to be serializable.
+// If the collection is larger than the capacity of the Ringbuffer, then the items that were written first will be
+// overwritten. Therefor this call will not block. The items are inserted in the order of the Iterator of the collection.
+// If an addAll is executed concurrently with an add or addAll, no guarantee is given that items are contiguous.
+// The result of the future contains the sequenceId of the last written item.
+func (rb *Ringbuffer) AddAll(ctx context.Context, overflowPolicy OverflowPolicy, items ...interface{}) (int64, error) {
+	elementData, err := rb.validateAndSerializeValues(items)
```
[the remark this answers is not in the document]
nitram509: Done.
1d9ef635ca7ba19bb522a59c57987e1003ad5a72

proxy_ring_buffer.go
nitram509: Done.
1ae142f8e57787f00df65b6c9e4888d75ec7b7a6

examples/ringbuffer/main.go
nitram509: removed.
e6aa7636c521723bfac7767ddc9eb65ffa139f79

internal/it/ring_buffer.go
nitram509: sounds better.
c2324b27434bf80e44fbaf7d5bab6e20ad5c490e

internal/it/ring_buffer.go
nitram509: 🤔 If I would do so, then it would be inconsistent with many other implementations in the same package/folder.
This seems to be worth a separate PR, because it would be a more broad refactoring, no?

nitram509: Are you fine with leaving as is?

proxy_ring_buffer.go
nitram509: Sure, done.
see c197bec31d867b265efa486072ff4fed1a9db6cf

nitram509: I took the freedom to change it in some more places, which did not follow that convention.

proxy_ring_buffer.go
nitram509: Done.
9aacf8e0b5617979552f848faa1f61b660b73f4a

proxy_ring_buffer.go
nitram509: You're right, I did remove this check.
7dca10b3d95a035e433a37c03420d56552ad1d73

proxy_ring_buffer.go
nitram509: Done.
I did simplified further and inlined the single line check, as it did add no further value IMHO
fe1b3c0d4b5d741111ae54a31d568fdbb7f8a506

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+	request := codec.EncodeRingbufferAddRequest(rb.name, int32(overflowPolicy), elementData)
+	response, err := rb.invokeOnPartition(ctx, request, rb.partitionID)
+	if err != nil {
+		return ReadResultSetSequenceUnavailable, err
+	}
+	return codec.DecodeRingbufferAddResponse(response), nil
+}
+
+// AddAll
+// Adds all the items of a collection to the tail of the Ringbuffer. A addAll is likely to outperform multiple calls
```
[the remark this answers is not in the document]
nitram509: Do you mean to align all the comments to golang style?
They then will read somewhat "odd" IMHO ... but following your guidance here.

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+	request := codec.EncodeRingbufferSizeRequest(rb.name)
+	response, err := rb.invokeOnPartition(ctx, request, rb.partitionID)
+	if err != nil {
+		return ReadResultSetSequenceUnavailable, err
+	}
+	return codec.DecodeRingbufferSizeResponse(response), nil
+}
+
+// TailSequence returns the sequence of the tail. The tail is the side of the ringbuffer where the items are added to.
+// The initial value of the tail is -1.
```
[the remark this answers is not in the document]
nitram509: 🤷 its straight from https://github.com/hazelcast/hazelcast-client-protocol/blob/60185af4a914b0a4742b9b83e1d6ff9f27762057/protocol-definitions/Ringbuffer.yaml#L32

How to handle this?
Would you consider/aggree updating the client protocol docs first?

proxy_ring_buffer.go
```diff
@@ -0,0 +1,278 @@
...
+	request := codec.EncodeRingbufferHeadSequenceRequest(rb.name)
+	response, err := rb.invokeOnPartition(ctx, request, rb.partitionID)
+	if err != nil {
+		return ReadResultSetSequenceUnavailable, err
+	}
+	return codec.DecodeRingbufferHeadSequenceResponse(response), nil
+}
+
+// RemainingCapacity
+// Returns the remaining capacity of the Ringbuffer. The returned value could be stale as soon as it is returned.
```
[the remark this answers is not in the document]
nitram509: 🤷 its straight from https://github.com/hazelcast/hazelcast-client-protocol/blob/60185af4a914b0a4742b9b83e1d6ff9f27762057/protocol-definitions/Ringbuffer.yaml#L32

How to handle this?
Would you consider/agree updating the client protocol docs first?

proxy_ring_buffer.go
nitram509: 🤷 I did copy the documentation from
https://github.com/hazelcast/hazelcast-client-protocol/blob/60185af4a914b0a4742b9b83e1d6ff9f27762057/protocol-definitions/Ringbuffer.yaml#L244

Would it make sense to patch the reference first?

(see also Slack).

proxy_ring_buffer.go
nitram509: 🤷 May I ask, what is the whole version, as the text originates from the client protocol repo?

proxy_ring_buffer.go
nitram509: I'm sorry, but I feel lost, in what was discussed in "TDD".
Is this already decided (and I shall change it) OR an open question?

proxy_ring_buffer.go
nitram509: Done.
See 0b4e5e0f33ea22aad6490a179cf8f238206d5bc2

proxy_ring_buffer.go
nitram509: Done. 
777e545034f3ae268e0f588f07c78fa536a6a207

proxy_ring_buffer.go
nitram509: 🤷 I'm not that much of a Hazelcast expert. May I ask for some guidance, please?

proxy_ring_buffer.go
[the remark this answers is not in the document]
srknzl: Not necessarily to golang style. As I realized, there are comments in style number 1 and number 2. We can select one style and folow that. However, you can skip this as well as we will update comments later.

proxy_ring_buffer.go
srknzl: Yeah it is decided. You can deserialize all items when you create the ReadResultSet and pass them to ReadResultSet. Then Get() will return the item directly.

ring_buffer_it_test.go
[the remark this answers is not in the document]
srknzl: Ok we can rename the tests later on as well. Let's don't bother with naming.

internal/it/ring_buffer.go
utku-caglayan: That's definitely a valid concern @nitram509 thanks for pointing that out!  We plan to refactor them as well but didn't get a chance to do that yet. Please ignore those "non-idiomatic" instances

proxy_ring_buffer.go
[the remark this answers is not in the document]
nitram509: Done.
b3039d7fd09ac1ad17ac9a8a079a59d05a229544

proxy_ring_buffer.go
nevzatseferoglu: I could not found the statement of "Go client supports context cancellation and timeouts natively" in `pkg doc` or in official repository. We are only saying that;
`Go context support for all distributed data structures.`. User can assumes that we are also supporting `context.WithDeadline` based on this statement.

I think context related edge cases and usage should be explained in more verbose way, especially for the public API.

nevzatseferoglu: > Go client supports context cancellation and timeouts natively, which is different from other clients. I think we can just remove the "Currently, it isn't possible to control how long this call is going to block. In the future we could add e.g. tryReadOne(long sequence, long timeout, TimeUnit unit)." part from the documentation.

Since there is no additional error type other than the `Cancelled` and `DeadlineExceeded`, can't we say that the public API calls support context usage in any sense without denoting the cancellation and timeout ? Because, user can use `WithDeadline` and it is quite similar to `WithTimeout` and produces same `DeadlineExceeded` error.

proxy_ring_buffer.go
nitram509: Done.
See 6e4837544829edc4dd39c191be46b65cf76cfbea

proxy_ring_buffer.go
nitram509: Done.
see bfcc1f0c6a4a407763d6e8da177a70df47395eae

proxy_ring_buffer.go
nitram509: Done. see 604359f2b718a2d71cd19e1852ce0c5c83b4408b

proxy_ring_buffer.go
nitram509: Done.
see 131f287231745e43d060a6bdc58078f9a1cc1dac

nitram509: And 05d7904de418415d8a016d7e12551b0a6c5d334c

proxy_ring_buffer.go
nitram509: For the ReadOne it makes sense and I changed it - see 172728c477ae3d0b93505108dd130f01c8afeef5
For other places, where return type is int64, one can't use nil, but must be ReadResultSetSequenceUnavailable instead

ring_buffer_it_test.go
nitram509: Done.
4130060823bf58a1344681792dc0f5d38e424d33

ring_buffer_it_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+		actualItem, err := rb.ReadOne(context.Background(), 0)
+		assert.NoError(t, err)
+		assert.Equal(t, "foo", actualItem)
+	})
+}
+
+func TestRingbuffer_AddNilElement(t *testing.T) {
+	it.RingbufferTester(t, func(t *testing.T, rb *hz.Ringbuffer) {
+		_, err := rb.Add(context.Background(), nil, hz.OverflowPolicyFail)
```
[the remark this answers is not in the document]
nitram509: It seems simply adding $capacity number of items does not trigger a fail.
How should I prepare a test for that?
May I ask for some guidance, please?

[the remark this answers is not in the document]
srknzl: I am only talking about overflow policy validation. I don't know why your case fails though.

proxy_ring_buffer.go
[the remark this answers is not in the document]
nitram509: Done.
99fa52328e90a44e5ef1fcaade10b769e00fd608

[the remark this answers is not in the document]
nitram509: @yuce please allow me some feedback:
```if int(overflowPolicy) != 0 || int(overflowPolicy) != 1 {```
will always return true (bool arethmetics).
The correct phrase is
```if overflowPolicy != OverflowPolicyFail && overflowPolicy != OverflowPolicyOverwrite```

File: proxy_ring_buffer.go [hunk 46]
```diff
@@ -0,0 +1,278 @@
...
+// If there are fewer items available than minCount, then this call blacks. Reading a batch of items is likely to
+// perform better because less overhead is involved. A filter can be provided to only select items that need to be read.
+// If the filter is null, all items are read. If the filter is not null, only items where the filter function returns
+// true are returned. Using filters is a good way to prevent getting items that are of no value to the receiver.
+// This reduces the amount of IO and the number of operations being executed, and can result in a significant performance improvement.
+func (rb *Ringbuffer) ReadMany(ctx context.Context, startSequence int64, minCount int32, maxCount int32, filter interface{}) (ReadResultSet, error) {
+	if filter != nil {
+		return ReadResultSet{}, errors.New("filter functions are not yet supported in th Go client, please use 'nil' for now")
+	}
+	request := codec.EncodeRingbufferReadManyRequest(rb.name, startSequence, minCount, maxCount, nil)
```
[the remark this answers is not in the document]
nitram509: @srknzl 

I did implement the defensive checks for "if xxx <0 ...".
See 4ed59c73d180f7f1a0c71272d8fe2fb047b2658a

I hesitate to implement the checks about capacity because this is something the server will do anyway.
if we implement this client-side, this requires extra lookups (yes cached, I know).
When there are any plans of e.g. dynamic resizing, these checks on client-side are difficult anyhow.
What I'm trying to say: I do propose to not implement them, as they are beyond the API's data contract and so should be implemented in one place only.

Would you consider agreeing, leaving in just the non-negative tests.

ring_buffer_it_test.go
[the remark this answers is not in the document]
nitram509: Tests implemented
09234496aa5f8b4ac3f744404bb4dc8d4609035f

ring_buffer_it_test.go
nitram509: I did add tests for non-negative checks.
0615f3197d41ca5d33e90725dd62916dcb6b48ed

implementing other tests seems strange to me, since we would test server-side logic, nothing that is implemented in the Go client.

proxy_ring_buffer.go
nitram509: @nevzatseferoglu @yuce @srknzl 
Since Go supports timeouts out of the box, may I propose, that we simply remove this from the documentation?
I took the freedom to commit it already, please do veto if you intent otherwise.
See bdf427baf5a29bae7beb9e3c76136f9dcf857f1d

proxy_ring_buffer.go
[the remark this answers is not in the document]
srknzl: I don't know why Java client has it but can you add it for now. We can discuss why we have it in our team internally. If we will remove it, it should be removed from Java first.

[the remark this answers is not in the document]
nitram509: Done
d15128b24b9e59278ddd608920811529820f3a93

yuce (approved):

yuce:
@nitram509 Thanks a lot for your contribution. We'll move this PR to a feature branch to apply documentation / test related changes. Here's the redirect PR: https://github.com/hazelcast/hazelcast-go-client/pull/813

utku-caglayan (approved):
Thanks for contributing!

yuce:
Closing in order to redirect: https://github.com/hazelcast/hazelcast-go-client/pull/813

PR #793, opened 2022-06-21
Fixes SQL driver tests to work with SSL

yuce:
SQL driver tests didn't work with SSL, since some parts of those tests only used a DSN which doesn't support SSL options. Converted problematic tests to use `driver.Open` instead.


utku-caglayan (approved):
Looks great, thanks!

nevzatseferoglu (approved):
Thanks.

PR #795, opened 2022-06-21
Fix SQL leak

### LINKED ISSUE #794: SQL Query leaks goroutines if not closed explicitly
Hazelcast Go Client version: v1.2.0 - master
Hazelcast server version: 5.1

#### Expected behaviour

Consuming all rows of the result of a query should terminate the goroutine that checks the context at https://github.com/hazelcast/hazelcast-go-client/blob/c6bf18636bdda15b42d751ff47694dd3ed2461b8/internal/sql/driver/result.go#L70

#### Actual behaviour

The goroutine leaks, since the `done` channel is never closed because the state is incorrectly set to `closed` when all rows in the result is consumed. See: https://github.com/hazelcast/hazelcast-go-client/blob/c6bf18636bdda15b42d751ff47694dd3ed2461b8/internal/sql/driver/result.go#L123

yuce:
Fixes #794

### DIFF
--- a/.github/workflows/nightly_tests.yaml
+++ b/.github/workflows/nightly_tests.yaml
@@ -30,7 +30,7 @@ jobs:
         env:
           HAZELCAST_ENTERPRISE_KEY: ${{ secrets.HAZELCAST_ENTERPRISE_KEY }}
           HZ_VERSION: 5.1
-          SSL_ENABLED: 1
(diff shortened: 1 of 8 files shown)


utku-caglayan (approved):
Great catch, thanks!

nevzatseferoglu (approved):
LGTM!

PR #806, opened 2022-06-24
NearCache Implementation Part I [API-1352]

yuce:
This PR contains:

* The base implementation and necessary data types for Map near cache.
* Test infrastructure for near cache.
* Local map stats (contains only near cache stats)
* Implementation and tests for the following Map methods:
    * Get
    * Remove
    * RemoveIfSame
    * Delete
    * Put
    * PutWIthTTL (implementation only)
    * PutWithMaxIdle
    * PutWithTTLAndMaxIdle (implementation only)
    * Set
    * SetWithTTL
    * SetWIthTTLAndMaxIdle
    * TryRemove
    * TryRemoveWithTimeout
    * TryPut
    * ContainsKey (implementation only)


config.go
```diff
@@ -72,15 +72,15 @@ func (c *Config) AddMembershipListener(handler cluster.MembershipStateChangeHand
 	return id
 }
 
-// AddNearCacheConfig adds a near cache configuration.
-func (c *Config) AddNearCacheConfig(cfg nearcache.Config) {
+// AddNearCache adds a near cache configuration.
+func (c *Config) AddNearCache(cfg nearcache.Config) {
```
srknzl, line 76: This is changing the public API and do not conform to TDD.

sql/driver/driver_it_test.go
```diff
@@ -654,23 +652,6 @@ func populateMap(m *hz.Map, count int, keyFn, valueFn func(i int) interface{}) (
...
-	q.Add("unisocket", strconv.FormatBool(config.Cluster.Unisocket))
-	q.Add("log", string(ll))
-	return fmt.Sprintf("hz://%s?%s", config.Cluster.Network.Addresses[0], q.Encode())
-}
-
-func mustDB(db *sql.DB, err error) *sql.DB {
-	it.Must(err)
-	return db
-}
-
```
srknzl, line 673: Why this change is included in this pr? What is the motivation of the change?

types/uuid.go
```diff
@@ -24,7 +24,6 @@ import (
 )
```
srknzl, line 24: Changes in this file seems unrelated to me.

srknzl (commented):
Will continue to review.

nearcache/config.go
```diff
@@ -37,10 +37,10 @@ type Config struct {
 	// Name is the name of this Near Cache configuration.
 	// If the name is not specified, it is set to "default".
 	Name string
-	// EvictionConfig is the optional eviction configuration for the Near Cache.
-	EvictionConfig EvictionConfig
```
srknzl, line 41: Why do you change public API that is not conforming to TDD?

nearcache_map.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func newNearCacheMap(nc *nearCache, ncc *nearcache.Config, ss *serialization.Service) (nearCacheMap, error) {
+	ncm := nearCacheMap{
+		nc:  nc,
+		ncc: ncc,
+		ss:  ss,
+	}
+	// the only valid local policy on the client side is invalidate.
+	ncm.registerInvalidationListener()
```
srknzl, line 42: In Java we don't check local policy but isInvalidateOnChange(), so why do you mention local policy in the comment? 
see https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/client/map/impl/nearcache/NearCachedClientMapProxy.java#L94

Will you check it in registerInvalidationListener?

nearcache_map.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+	}
+	// toNearCacheKey returns the raw key if SerializeKeys is not true.
+	if ncc.SerializeKeys {
+		ncm.toNearCacheKey = func(key interface{}) (interface{}, error) {
+			data, err := ss.ToData(key)
+			if err != nil {
+				return nil, err
+			}
+			// byte slices cannot be map keys
```
srknzl, line 55: What does this comment mean? I think users can use `[]byte` as key in a non-nearcache IMap. can't they?

nearcache_map.go
```diff
@@ -0,0 +1,265 @@
...
+		return false, err
+	}
+	if ok {
+		return cached != nil, nil
+	}
+	keyData, err := m.validateAndSerialize(key)
+	if err != nil {
+		return false, err
+	}
+	return m.containsKeyFromRemote(ctx, keyData)
```
srknzl, line 90: Does it makes sense to add a pointer to the map proxy to nearCacheMap and introduce a new method called something like `containsKeyInternal` and call it here instead of copying the code to each method. It will be hard to update. 

So,

ContainsKey in proxy_map.go:

```go
// ContainsKey returns true if the map contains an entry with the given key.
func (m *Map) ContainsKey(ctx context.Context, key interface{}) (bool, error) {
	if m.hasNearCache {
		return m.ncm.ContainsKey(ctx, key, m)
	}
	return m.containsInternal(ctx, key)
}

// ContainsKey returns true if the map contains an entry with the given key.
func (m *Map) containsInternal(ctx context.Context, key interface{}) (bool, error) {
        keyData, err := m.validateAndSerialize(key)
	if err != nil {
		return false, err
	}
	return m.containsKeyFromRemote(ctx, keyData)
}
``` 

ContainsKey in nearcache_map.go:

```go
func (ncm *nearCacheMap) ContainsKey(ctx context.Context, key interface{}, m *Map) (found bool, err error) {
	key, err = ncm.toNearCacheKey(key)
	if err != nil {
		return false, err
	}
	cached, ok, err := ncm.getCachedValue(key, false)
	if err != nil {
		return false, err
	}
	if ok {
		return cached != nil, nil
	}
	return ncm.m.containsKeyInternal(ctx, key)
}
```

Similar for other methods.

srknzl (changes requested):

nearcache_map.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	return m.deleteFromRemote(ctx, keyData)
+}
+
+func (ncm *nearCacheMap) Get(ctx context.Context, m *Map, key interface{}) (interface{}, error) {
+	key, err := ncm.toNearCacheKey(key)
+	if err != nil {
+		return nil, err
+	}
+	cached, found, err := ncm.getCachedValue(key, false)
```
srknzl, line 111: Should not this be true? https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/client/map/impl/nearcache/NearCachedClientMapProxy.java#L117

```suggestion
	cached, found, err := ncm.getCachedValue(key, true)
```

nearcache_map.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		return false, err
+	}
+	if ok {
+		return cached != nil, nil
+	}
+	keyData, err := m.validateAndSerialize(key)
+	if err != nil {
+		return false, err
+	}
```
srknzl, line 89: We can put serialization logic into `containsKeyFromRemote` as well.

nearcache_map.go
```diff
@@ -0,0 +1,265 @@
...
+	key, err := ncm.toNearCacheKey(key)
+	if err != nil {
+		return err
+	}
+	defer ncm.nc.Invalidate(key)
+	keyData, err := m.validateAndSerialize(key)
+	if err != nil {
+		return err
+	}
+	return m.deleteFromRemote(ctx, keyData)
```
srknzl, line 103: We can put serialization logic into `deleteFromRemote` as well.

nearcache_map.go
```diff
@@ -0,0 +1,265 @@
...
+	key, err := ncm.toNearCacheKey(key)
+	if err != nil {
+		return nil, err
+	}
+	defer ncm.nc.Invalidate(key)
+	keyData, valueData, err := m.validateAndSerialize2(key, value)
+	if err != nil {
+		return false, err
+	}
+	return m.putWithTTLFromRemote(ctx, keyData, valueData, ttl)
```
srknzl, line 138: Why do we return false if err is not nil? It should be nil I guess. Also why not use `m.putWithTTL`?

nearcache_map.go
```diff
@@ -0,0 +1,265 @@
...
+	key, err := ncm.toNearCacheKey(key)
+	if err != nil {
+		return nil, err
+	}
+	defer ncm.nc.Invalidate(key)
+	keyData, valueData, err := m.validateAndSerialize2(key, value)
+	if err != nil {
+		return false, err
+	}
+	return m.putWithMaxIdleFromRemote(ctx, keyData, valueData, ttl, maxIdle)
```
srknzl, line 151: If you refactor `putWithMaxIdle` (move the if to `PutWithMaxIdle`) you can use `m.putWithMaxIdle` here directly. (code reuse)

nearcache_map.go
```diff
@@ -0,0 +1,265 @@
...
+func (ncm *nearCacheMap) Remove(ctx context.Context, m *Map, key interface{}) (interface{}, error) {
+	key, err := ncm.toNearCacheKey(key)
+	if err != nil {
+		return false, err
+	}
+	defer ncm.nc.Invalidate(key)
+	keyData, err := m.validateAndSerialize(key)
+	if err != nil {
+		return nil, err
+	}
```
srknzl, line 163: You can move serialization logic to `removeFromRemote` to have just one instance of the same code snippet that is used in this file and proxy_map.go

nearcache_map.go
```diff
@@ -0,0 +1,265 @@
...
+func (ncm *nearCacheMap) RemoveIfSame(ctx context.Context, m *Map, key interface{}, value interface{}) (bool, error) {
+	key, err := ncm.toNearCacheKey(key)
+	if err != nil {
+		return false, err
+	}
+	defer ncm.nc.Invalidate(key)
+	keyData, valueData, err := m.validateAndSerialize2(key, value)
+	if err != nil {
+		return false, err
+	}
```
srknzl, line 176: Similarly, you can move serialization logic to removeIfSameFromRemote

nearcache_map.go
```diff
@@ -0,0 +1,265 @@
...
+func (ncm *nearCacheMap) Set(ctx context.Context, m *Map, key, value interface{}, ttl int64) error {
+	key, err := ncm.toNearCacheKey(key)
+	if err != nil {
+		return err
+	}
+	defer ncm.nc.Invalidate(key)
+	keyData, valueData, err := m.validateAndSerialize2(key, value)
+	if err != nil {
+		return err
+	}
```
srknzl, line 189: By moving if hasNearCache check to `m.Set` you can move this serialization code to `setFromRemote`

nearcache_map.go
```diff
@@ -0,0 +1,265 @@
...
+func (ncm *nearCacheMap) SetWithTTLAndMaxIdle(ctx context.Context, m *Map, key, value interface{}, ttl time.Duration, maxIdle time.Duration) error {
+	key, err := ncm.toNearCacheKey(key)
+	if err != nil {
+		return err
+	}
+	defer ncm.nc.Invalidate(key)
+	keyData, valueData, err := m.validateAndSerialize2(key, value)
+	if err != nil {
+		return err
+	}
```
srknzl, line 202: You can move this serialization logic to `setWithTTLAndMaxIdleFromRemote`

nearcache_map.go
```diff
@@ -0,0 +1,265 @@
...
+func (ncm *nearCacheMap) TryRemove(ctx context.Context, m *Map, key interface{}, timeout int64) (interface{}, error) {
+	key, err := ncm.toNearCacheKey(key)
+	if err != nil {
+		return false, err
+	}
+	defer ncm.nc.Invalidate(key)
+	keyData, err := m.validateAndSerialize(key)
+	if err != nil {
+		return nil, err
+	}
```
srknzl, line 215: You can move this code to `tryRemoveFromRemote`

config.go
yuce: Yes, I will update the public API TDD and notify the reviewers.

sql/driver/driver_it_test.go
yuce: It removes dead code we forgot to remove in another merged PR.

nearcache_map.go
yuce: These `XFromRemote` methods are used by both `Map` and `nearCacheMap`. By not serializing the key eagerly, we don't pay the serialization cost if the key is already found in the near cache.

nearcache_map.go
yuce: `m.Put` and `m.PutWithTTL` already calls this method if the near cache is enabled. If there is an error, the other returned value is not used, so it's not very important what value is there. But conventionally, the zero-value is returned.

nearcache_map.go
yuce: The same reply for other similar comments.

proxy_map.go
```diff
@@ -100,6 +101,8 @@ You can pass any context.Context to any Map function, but in that case lock owne
 */
 type Map struct {
 	*proxy
+	ncm          nearCacheMap
```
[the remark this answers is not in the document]
yuce: `ncm` is only set when there is a near cache for that map, so putting `hasNearCache` in`nearCacheMap` would be a little strange.

I think `nearCacheMap` is not too bad. Let's keep it if you don't have strong opinion about it.

types/uuid.go
yuce: UUID is used in the `nearCacheRecord`. Having the `text` of the UUID would increase the memory usage a lot. So I've removed it.

nearcache_map.go
srknzl: So, is the comment meaningless here? Can you remove it?

nearcache_map.go
```diff
@@ -0,0 +1,261 @@
...
+	"fmt"
+	"time"
+
+	"github.com/hazelcast/hazelcast-go-client/internal/serialization"
+	"github.com/hazelcast/hazelcast-go-client/nearcache"
+)
+
+type nearCacheMap struct {
+	nc             *nearCache
+	ncc            *nearcache.Config
```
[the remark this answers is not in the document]
yuce: That field wasn't necessary, removed at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/008e4e649ab1ac678e53d3b9653f2875384ed17b

nearcache_map.go
yuce: OK, you are right. I've changed the methods of the nearCacheMap methods before, so it seems I didn't realize there was an optimization opportunity in the latest revision. Refactored at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/a5b6adc0848f374f5049cf9a75bcb2aacccf9699

nearcache_map.go
yuce: Refactored at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/a5b6adc0848f374f5049cf9a75bcb2aacccf9699

nearcache_map.go
yuce: Refactored at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/a5b6adc0848f374f5049cf9a75bcb2aacccf9699

nearcache_map.go
yuce: Refactored at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/a5b6adc0848f374f5049cf9a75bcb2aacccf9699

nearcache_map.go
yuce: Refactored at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/a5b6adc0848f374f5049cf9a75bcb2aacccf9699

nearcache_map.go
yuce: Refactored at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/a5b6adc0848f374f5049cf9a75bcb2aacccf9699

nearcache_map.go
yuce: Refactored at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/a5b6adc0848f374f5049cf9a75bcb2aacccf9699

nearcache_map.go
yuce: Refactored at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/a5b6adc0848f374f5049cf9a75bcb2aacccf9699

nearcache.go
```diff
@@ -0,0 +1,689 @@
...
+func (n nearCacheValueStoreAdapter) ConvertValue(value interface{}) (interface{}, error) {
+	data, ok := value.(serialization.Data)
+	if !ok {
+		return value, nil
+	}
+	return n.ss.ToObject(data)
+}
+
+func (n nearCacheValueStoreAdapter) GetRecordStorageMemoryCost(rec *nearCacheRecord) int64 {
+	return 0
```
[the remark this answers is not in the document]
utku-caglayan: something similar would be great

nearcache.go
```diff
@@ -0,0 +1,689 @@
...
+	nearCacheRecordNotReserved     int64 = -1
+	nearCacheRecordReadPermitted         = -2
+)
+
+type nearCacheRecordValueConverter interface {
+	ConvertValue(value interface{}) (interface{}, error)
+}
+
+type nearCacheStorageEstimator interface {
+	GetRecordStorageMemoryCost(rec *nearCacheRecord) int64
```
utku-caglayan, line 153: Why not make key size estimation also a part of this?

nearcache_map.go
```diff
@@ -0,0 +1,230 @@
...
+		ncm.toNearCacheKey = func(key interface{}) (interface{}, error) {
+			data, err := ss.ToData(key)
+			if err != nil {
+				return nil, err
+			}
+			return data, nil
+		}
+	} else {
+		ncm.toNearCacheKey = func(key interface{}) (interface{}, error) {
+			return key, nil
```
utku-caglayan, line 58: I guess we need to return an error for unsupported key types here?

utku-caglayan (commented):
Mostly noted question, will continue to review

nearcache_map.go
```diff
@@ -0,0 +1,261 @@
...
+func newNearCacheMap(nc *nearCache, ncc *nearcache.Config, ss *serialization.Service) (nearCacheMap, error) {
+	ncm := nearCacheMap{
+		nc:  nc,
+		ncc: ncc,
+		ss:  ss,
+	}
+	// the only valid local policy on the client side is invalidate.
+	ncm.registerInvalidationListener()
+	if ncc.PreloaderConfig.Enabled {
+		if err := ncm.preload(); err != nil {
```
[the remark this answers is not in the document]
yuce: I didn't check the preloading stuff yet, but it may return an error before running the preloader goroutine.

client_it_test.go
```diff
@@ -556,15 +556,21 @@ func TestClientFailover_OSSCluster(t *testing.T) {
...
+	if it.TraceLoggingEnabled() {
+		config.Logger.Level = logger.TraceLevel
+	}
 	config.Failover.Enabled = true
-	config.Failover.TryCount = 1
-	// move the main cluster config to failover config list
-	config.Failover.SetConfigs(config.Cluster)
-	// use a non-existing cluster in the main cluster config
-	config.Cluster.Name = "non-existing-failover-test-cluster"
+	cfg1 := cls1.DefaultConfig()
```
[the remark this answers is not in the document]
yuce: It seems so, the changes in this PR should be OK now.

internal/it/map.go
```diff
@@ -21,66 +21,126 @@ import (
...
+		tcx.T.Run("Non-Smart Client", func(t *testing.T) {
+			tt := tcx
+			tt.Smart = false
+			tt.T = t
+			runner(tt)
 		})
 	}
 }
 
+func (tcx MapTestContext) OK(v bool) {
```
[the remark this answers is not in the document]
yuce: Not that I can see. `assert.X` functions don't fail immediately. Do you know an alternative in testify?

nearcache_map.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+	}
+	return value, true, nil
+}
+
+func (ncm *nearCacheMap) getFromRemote(ctx context.Context, m *Map, key interface{}) (interface{}, error) {
+	keyData, err := m.validateAndSerialize(key)
+	if err != nil {
+		return nil, err
+	}
```
srknzl, line 214: You can move this serialization code to `m.getFromRemote` as well. I missed this before.

internal/serialization/serialization.go
```diff
@@ -68,6 +68,9 @@ func (s *Service) ToData(object interface{}) (r Data, err error) {
 			err = makeError(rec)
 		}
 	}()
+	if object == nil {
```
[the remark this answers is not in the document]
srknzl: Looks like java has a check https://github.com/hazelcast/hazelcast/blob/a37a637242b225a762056be0d9ca0b4d8018290a/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/AbstractSerializationService.java#L196. We don't have this check in Node.js client though, I don't know why

nearcache.go
```diff
@@ -0,0 +1,689 @@
...
+		cfg:   cfg,
+		store: newNearCacheRecordStore(cfg, ss, rc, se),
+	}
+}
+
+func (nc *nearCache) Get(key interface{}) (interface{}, bool, error) {
+	_, ok := key.(serialization.Data)
+	if nc.cfg.SerializeKeys {
+		if !ok {
+			panic("key must be of type serialization.Data!")
```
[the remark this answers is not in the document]
yuce: No, we don't recover. That panic code should never run, and we must know if it does.

Makefile
```diff
@@ -3,7 +3,7 @@
 PORT ?= 5050
 MEMBER_COUNT ?= 3
 COVERAGE_OUT ?= coverage.out
-TEST_FLAGS ?= -v -count 1 -timeout 25m -tags=hazelcastinternal,hazelcastinternaltest
+TEST_FLAGS ?= -v -count 1 -p 1 -timeout 50m -tags=hazelcastinternal,hazelcastinternaltest
```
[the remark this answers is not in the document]
srknzl: I was going to ask this but removed my comment because this change is not included in this PR now. Is it merged in another PR?

nearcache.go
```diff
@@ -0,0 +1,689 @@
...
+	}
+	if cached != nil {
+		value = cached
+	}
+	return value, nil
+}
+
+const (
+	nearCacheRecordStoreTimeNotSet int64 = -1
+	pointerCostInBytes                   = (32 << uintptr(^uintptr(0)>>63)) >> 3
```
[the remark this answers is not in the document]
yuce: That's the size of a pointer in bytes, depending on the archtiecture.

nearcache_map.go
yuce: We would have to check every key with reflection. The user gets this panic if they try to use a non-supported key:
```
panic: runtime error: hash of unhashable type []uint8
```
Also I will mention that in the documentation PR. How about that?

internal/serialization/serialization.go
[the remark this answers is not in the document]
yuce: I removed this for now, probably it's not needed anymore but need to check again later.

internal/it/nearcache_util.go
```diff
@@ -0,0 +1,152 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/hazelcast/hazelcast-go-client/internal/serialization"
+	"github.com/hazelcast/hazelcast-go-client/nearcache"
+)
+
+type InvalidationRequestStore interface {
```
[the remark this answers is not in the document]
yuce: Removed at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/a14c1f642d5c2de5cc57b9942f6baae421864a87

internal/serialization/serialization.go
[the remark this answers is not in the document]
yuce: Removed at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/a14c1f642d5c2de5cc57b9942f6baae421864a87

nearcache.go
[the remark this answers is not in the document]
yuce: Added at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/a14c1f642d5c2de5cc57b9942f6baae421864a87

client_internals_test_utils.go
```diff
@@ -44,3 +45,16 @@ func (ci *ClientInternal) InvocationHandler() invocation.Handler {
 func (ci *ClientInternal) ClusterService() *cluster.Service {
 	return ci.client.ic.ClusterService
 }
+
+func (ci *ClientInternal) SerializationService() *serialization.Service {
+	return ci.client.ic.SerializationService
+}
+
+// MakeNearCacheAdapterFromMap returns the nearcache of the given map.
+// It returns an interface{} instead of it.NearCacheAdapter in order to not introduce an import cycle.
```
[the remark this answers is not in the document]
yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/7680785ac6d16794f25cba6159322be5ccad3b97

nearcache.go
utku-caglayan: Java client implements `getKeyStorageMemoryCost` for both `NearCacheDataRecordStore` and `NearCacheObjectRecordStore`. If key is of type "serialization.Data" there is a cost but in our case, it never is I guess. Wdyt?

nearcache.go
```diff
@@ -0,0 +1,690 @@
...
+	CreationTime         int64
+	value                internal.AtomicValue
+	UUID                 types.UUID
+	PartitionID          int32
+	lastAccessTime       int64
+	ExpirationTime       int64
+	InvalidationSequence int64
+	reservationID        int64
+	hits                 int32
+	cachedAsNil          int32
```
utku-caglayan, line 583: In the TDD, I think we can mention that this is bool for java client

nearcache.go
srknzl: I think getKeyStorageMemoryCost can directly return 0 because it is impossible to put a `serialization.Data` in the map.

nearcache.go
```diff
@@ -0,0 +1,690 @@
...
+	return nil
+}
+
+type nearCacheRecord struct {
+	CreationTime         int64
+	value                internal.AtomicValue
+	UUID                 types.UUID
+	PartitionID          int32
+	lastAccessTime       int64
+	ExpirationTime       int64
```
utku-caglayan, line 579: These are int32 on Java client. I think int64 is more convenient for implementation but it would be great to make a note about it on TDD. Because I guess it is for memory optimization.

Makefile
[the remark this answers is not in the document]
yuce: No, master branch doesn't have it.

nearcache.go
```diff
@@ -0,0 +1,690 @@
...
+
+func (nc *nearCache) TryPublishReserved(key, value interface{}, reservationID int64) (interface{}, error) {
+	cached, err := nc.store.TryPublishReserved(key, value, reservationID, true)
+	if err != nil {
+		return nil, err
+	}
+	if cached != nil {
+		value = cached
+	}
+	return value, nil
```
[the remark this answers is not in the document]
yuce: Current one returns the value itself if `cached == nil`. SImilar to: com.hazelcast.client.map.impl.nearcache.NearCachedClientMapProxy#tryPublishReserved

nearcache.go
```diff
@@ -0,0 +1,690 @@
...
+	estimator        nearCacheStorageEstimator
+}
+
+func newNearCacheRecordStore(cfg *nearcache.Config, ss *serialization.Service, rc nearCacheRecordValueConverter, se nearCacheStorageEstimator) nearCacheRecordStore {
+	stats := nearCacheStats{
+		CreationTime: time.Now(),
+	}
+	return nearCacheRecordStore{
+		recordsMu:        &sync.Mutex{},
+		records:          map[interface{}]*nearCacheRecord{},
```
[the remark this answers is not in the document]
yuce: Yes

nearcache.go
```diff
@@ -0,0 +1,690 @@
...
+}
+
+func (nc *nearCache) Get(key interface{}) (interface{}, bool, error) {
+	_, ok := key.(serialization.Data)
+	if nc.cfg.SerializeKeys {
+		if !ok {
+			panic("key must be of type serialization.Data!")
+		}
+	} else if ok {
+		panic("key cannot be of type Data!")
```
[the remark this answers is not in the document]
yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/76463bffb181783d422e7afdc7c6e69accef5f64

nearcache.go
```diff
@@ -0,0 +1,690 @@
...
+	}
+	return nc.store.Get(key)
+}
+
+func (nc *nearCache) Invalidate(key interface{}) {
+	nc.store.Invalidate(key)
```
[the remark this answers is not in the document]
yuce: Added to `Invalidate`. Java client has it for `Put`, but we don't yet have that method. https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/76463bffb181783d422e7afdc7c6e69accef5f64

Makefile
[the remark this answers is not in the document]
srknzl: Near cache feature branch has it so in the first PR we merged this change.

nearcache.go
yuce: That makes sense, thanks.

nearcache.go
```diff
@@ -0,0 +1,690 @@
...
+	cost := pointerCostInBytes + // the record is stored as a pointer in the map
+		5*int64CostInBytes + // CreationTime, lastAccessTime, ExpirationTime, InvalidationSequence, reservationID
+		3*int32CostInBytes + // PartitionID, hits, cachedAsNil
+		1*atomicValueCostInBytes + // holder of the value
+		1*uuidCostInBytes // UUID
+	value := rec.Value()
+	data, ok := value.(serialization.Data)
+	if ok {
+		cost += data.DataSize()
+	}
```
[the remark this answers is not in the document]
yuce: It seems the value of the record can be nil ATM, since this method can be called earlier than the Java implementation so the if (or if check against nil) is necessary. Let me check that...

nearcache.go
srknzl: However, since we know the memory size of serialization.Data the cost it should be at least the cost of serialization.Data because string representation will take more space. Should we get the memory size of the serialization.Data? I think it is the most accurate one.

nearcache.go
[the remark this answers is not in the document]
srknzl: value should be either nil or serialization.Data since `nearCacheDataStoreAdapter.ConvertValue`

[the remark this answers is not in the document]
yuce: Removed that check at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/5e33d6ad56495f39428a579a54d530864570ac49

nearcache.go
yuce: int32 is the correct one. I think I've changed that in part 2, let me recheck...

nearcache.go
```diff
@@ -0,0 +1,690 @@
...
+	rs.recordsMu.Lock()
+	size := len(rs.records)
+	rs.recordsMu.Unlock()
+	return size
+}
+
+func (rs nearCacheRecordStore) makeMapKey(key interface{}) interface{} {
+	data, ok := key.(serialization.Data)
+	if ok {
+		// serialization.Data is not hashable, conver it to string
```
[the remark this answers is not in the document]
yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/1fcad5f05b0b7066d202c9fed663cae58c3440ad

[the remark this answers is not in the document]
yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/1fcad5f05b0b7066d202c9fed663cae58c3440ad

nearcache.go
```diff
@@ -0,0 +1,692 @@
...
+}
+
+func (rs *nearCacheRecordStore) nextReservationID() int64 {
+	return atomic.AddInt64(&rs.reservationID, 1)
+}
+
+func (rs *nearCacheRecordStore) reserveForWriteUpdate(key interface{}, keyData serialization.Data, reservationID int64) (*nearCacheRecord, error) {
+	rs.recordsMu.Lock()
+	defer rs.recordsMu.Unlock()
+	rec, ok := rs.records[key]
```
srknzl, line 503: Is doing this better? 

```suggestion
	rs.recordsMu.Lock()
	rec, ok := rs.records[key]
	rs.recordsMu.Unlock()
```

nearcache.go
```diff
@@ -0,0 +1,692 @@
...
+	if !ok {
+		rec, err := rs.createRecord(nil)
+		if err != nil {
+			return nil, err
+		}
+		rec.SetReservationID(reservationID)
+		// initInvalidationMetaData(record, key, keyData);
```
srknzl, line 510:
```suggestion
		// initInvalidationMetaData(record, key, keyData);
		return rec, nil
```

nearcache.go
```diff
@@ -0,0 +1,692 @@
...
+		return nil, err
+	}
+	rec.SetReservationID(reservationID)
+	rs.records[key] = rec
+	// initInvalidationMetaData(record, key, keyData);
+	return rec, nil
+}
+
+func (rs *nearCacheRecordStore) createRecord(value interface{}) (*nearCacheRecord, error) {
+	// assumes recordsMu was locked elsewhere
```
srknzl, line 552: Does this method need the mutex to be locked? It does not update rs.records 🤔

nearcache.go
```diff
@@ -0,0 +1,692 @@
...
+	CreationTime time.Time
+	// LastPersistenceTime is the time of the last completed persistence task.
+	LastPersistenceTime time.Time
+	// LastPersistenceFailure is the error message of the last completed persistence task.
+	LastPersistenceFailure string
+	// LastPersistenceDuration is the duration of the last completed persistence task.
+	LastPersistenceDuration time.Duration
+}
+
+func (st *nearCacheStats) InvalidationRequests() int64 {
```
srknzl, line 690: This is not used for now, right? Will it be used in later PRs?

nearcache.go
```diff
@@ -0,0 +1,692 @@
...
+	rs.recordsMu.Unlock()
+	if canUpdateStats {
+		rs.decrementOwnedEntryCount()
+		rs.decrementOwnedEntryMemoryCost(rs.getTotalStorageMemoryCost(key, rec))
+		rs.incrementInvalidations()
+	}
+	rs.incrementInvalidationRequests()
+}
+
+func (rs *nearCacheRecordStore) TryPublishReserved(key interface{}, value interface{}, reservationID int64, deserialize bool) (interface{}, error) {
```
srknzl, line 296: nit:
```suggestion
func (rs *nearCacheRecordStore) TryPublishReserved(key, value interface{}, reservationID int64, deserialize bool) (interface{}, error) {
```

internal/it/adapters.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"context"
+
+	"github.com/hazelcast/hazelcast-go-client"
+)
+
+type DataStructureAdapter interface {
+	Get(ctx context.Context, key interface{}) (interface{}, error)
+	Put(ctx context.Context, key interface{}, value interface{}) (interface{}, error)
```
srknzl, line 27: nit:
```suggestion
	Put(ctx context.Context, key, value interface{}) (interface{}, error)
```

nearcache.go
```diff
@@ -0,0 +1,690 @@
...
+	return nc.store.Size()
+}
+
+func (nc nearCache) Stats() nearcache.Stats {
+	return nc.store.Stats()
+}
+
+func (nc nearCache) InvalidationRequests() int64 {
+	return nc.store.InvalidationRequests()
+}
```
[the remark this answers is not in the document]
yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/5e33d6ad56495f39428a579a54d530864570ac49

nearcache.go
```diff
@@ -0,0 +1,695 @@
...
+		rs.incrementInvalidations()
+	}
+	rs.incrementInvalidationRequests()
+}
+
+func (rs *nearCacheRecordStore) TryPublishReserved(key interface{}, value interface{}, reservationID int64, deserialize bool) (interface{}, error) {
+	key = rs.makeMapKey(key)
+	rs.recordsMu.Lock()
+	defer rs.recordsMu.Unlock()
+	existing, ok := rs.records[key]
```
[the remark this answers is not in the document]
yuce: I think it's better to access the record in the critical section.

nearcache.go
yuce: Like above the record is mutated in this method. I think it's better to do that in the critical section.

nearcache.go
yuce: Fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/5626f0144caeac26ef075f5c3f682dd400937ffb

nearcache.go
yuce: :thinking: probably not.

nearcache.go
srknzl: I know but what should be the memory cost of the key in that case?

nearcache.go
yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/aaa18f6bacb80c4aacf00ef00b53024843e53fbb

nearcache.go
yuce: It seems not, removed at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/aaa18f6bacb80c4aacf00ef00b53024843e53fbb

nearcache.go
srknzl: I know. I think it is not accurate since we convert Data to string before storing in the `records` map. Note that, string representation of Data will take more space than Data. On the other hand, since we convert it to a string, and think it as a  `OBJECT` the cost should be 0. So should it be 0, `int64(bits.UintSize/8 + keyData.DataSize())` or more than `int64(bits.UintSize/8 + keyData.DataSize())` ?

yuce: Why do you think the string representation takes more space that `Data` which is a byte slice?

internal/it/map.go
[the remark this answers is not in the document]
yuce: OK, changed at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/1912f650c432c7de6962a406d63408875a968d73

nearcache.go
srknzl: I thought it would print something like: 

`xff xde x12` but looks like it just stores the byte slice internally. Is this right?

yuce: That's right.

nearcache/nearcache_it_test.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+	for _, tc := range testCases {
+		testName := fmt.Sprintf("inMemoryFmt %s serializeKeys %t", inMemoryFmtToString(tc.inMemoryFmt), tc.serializeKeys)
+		t.Run(testName, func(t *testing.T) {
+			ncc := nearcache.Config{
+				Name:           "*",
+				InMemoryFormat: tc.inMemoryFmt,
+				SerializeKeys:  tc.serializeKeys,
+			}
+			ncc.SetInvalidateOnChange(false)
```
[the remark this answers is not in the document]
yuce: The test uses some on the Java side. Some of the relevant code is at:
* `com.hazelcast.client.cache.impl.nearcache.ClientCacheNearCacheBasicSlowTest`
* `com.hazelcast.internal.nearcache.impl.AbstractNearCacheBasicTest#whenGetIsUsed_thenNearCacheShouldBePopulated()`

I don't know where is the exact spot that sets invalidate on change to false, but here's how it looks in the debugger right before the test starts:
image

nearcache/nearcache_it_test.go
```diff
@@ -0,0 +1,327 @@
...
+	invalidationRunner(t, testCases)
+}
+
+func TestAfterPutNearCacheIsInvalidated(t *testing.T) {
+	// port of: com.hazelcast.client.map.impl.nearcache.ClientMapNearCacheTest#testAfterPutAsyncNearCacheIsInvalidated
+	testCases := []mapTestCase{
+		{
+			name: "Put",
+			f: func(ctx context.Context, tcx it.MapTestContext, i int64) {
+				if err := tcx.M.Delete(ctx, i); err != nil {
```
srknzl, line 122: This should be Put()

nearcache/nearcache_it_test.go
```diff
@@ -0,0 +1,327 @@
...
+				// TODO: assertNearCacheReferences
+			})
+		})
+	}
+}
+
+func invalidationRunner(t *testing.T, testCases []mapTestCase) {
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tcx := newNearCacheMapTestContext(t, nearcache.InMemoryFormatBinary, false)
```
srknzl, line 269: This should be true I guess. see [this](https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/test/java/com/hazelcast/client/map/impl/nearcache/ClientMapNearCacheTest.java#L185). How does the test pass? 🤔

nearcache_map.go
yuce: `keyData` is passed to `TryReserveForUpdate` at line 215. So `key` would be serialized twice if we move that code to `getFromRemote`.

proxy_map.go
```diff
@@ -1112,32 +1228,15 @@ func (m *Map) lock(ctx context.Context, key interface{}, ttl int64) error {
...
-		return nil, err
-	} else {
-		request := codec.EncodeMapPutRequest(m.name, keyData, valueData, lid, ttl)
-		if response, err := m.invokeOnKey(ctx, request, keyData); err != nil {
-			return nil, err
-		} else {
-			return m.convertToObject(codec.DecodeMapPutResponse(response))
-		}
-	}
+func (m *Map) putWithTTL(ctx context.Context, key, value interface{}, ttl int64) (interface{}, error) {
```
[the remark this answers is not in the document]
yuce: I've added a near cache check to `putWithTTL` to `Put` and `PutWithTTL` can call it directly, similar to `putWithMaxIdle`.

nearcache/nearcache_it_test.go
yuce: Right. But Java client test also passes with `.setInvalidateOnChange(false)`.

nearcache/nearcache_it_test.go
yuce: Fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/0340cb385720900db303b5811060b367f99c9526

nearcache/nearcache_it_test.go
yuce: Fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/0340cb385720900db303b5811060b367f99c9526

nearcache_map.go
srknzl: ok sorry.

nearcache/nearcache_it_test.go
```diff
@@ -0,0 +1,329 @@
...
+		},
+		{
+			name: "PutWithMaxIdle",
+			f: func(ctx context.Context, tcx it.MapTestContext, i int64) {
+				v, err := tcx.M.PutWithMaxIdle(ctx, i, i, 10*time.Minute)
+				if err != nil {
+					tcx.T.Fatal(err)
+				}
+				require.Equal(tcx.T, i, v)
+			},
```
srknzl, line 212:
```suggestion
			name: "TryPut",
			f: func(ctx context.Context, tcx it.MapTestContext, i int64) {
				v, err := tcx.M.TryPut(ctx, i, i)
				if err != nil {
					tcx.T.Fatal(err)
				}
				require.Equal(tcx.T, i, v)
			},
		},
		{
			name: "TryPutWithTimeout",
			f: func(ctx context.Context, tcx it.MapTestContext, i int64) {
				v, err := tcx.M.TryPutWithTimeout(ctx, i, i, 10*time.Second)
				if err != nil {
					tcx.T.Fatal(err)
				}
				require.Equal(tcx.T, i, v)
			},
```

yuce: It seems I didn't add nearcache support for TryPut, but added a wrong test for it. Thanks, added at: https://github.com/hazelcast/hazelcast-go-client/pull/806/commits/de818d94eb13ae503e6ecfa8f8f0723879d22e38

utku-caglayan (approved):
Great thanks!

srknzl (approved):

PR #807, opened 2022-06-24
NearCache Implementation Part II [API-1377]

yuce:
This PR:

* Adds `Invalidation` configuration with reconciliation settings.
* Adds support for MC near cache stats.
* Implements expiration.
* Adds listeners for invalidation messages.
* Adds the repairing task.
* Adds the stale read detector.
* Removes internal `nearcache.Stats` and uses the public `nearcache.Stats` type.*


config.go
```diff
@@ -30,12 +31,18 @@ import (
 	"github.com/hazelcast/hazelcast-go-client/types"
 )
 
+const (
+	defaultMaxToleratedMissCount            = 10
```
srknzl, line 35: we can add the location of these constants in Java client as a comment here

`com.hazelcast.internal.nearcache.impl.invalidation.RepairingTask`

config.go
```diff
@@ -264,6 +281,63 @@ func (f *FlakeIDGeneratorConfig) Validate() error {
...
+	return *pc.maxToleratedMissCount
+}
+
+func (pc *InvalidationConfig) SetReconciliationIntervalSeconds(seconds int) {
+	if err := check.NonNegativeInt32Config(seconds); err != nil {
+		pc.err = fmt.Errorf("ReconciliationIntervalSeconds: %w", err)
+		return
+	}
+	if seconds != 0 && seconds < defaultMinReconciliationIntervalSeconds {
+		pc.err = fmt.Errorf("ReconciliationIntervalSeconds: must be 0 or greater or equal to 30: %w", pubhzerrors.ErrInvalidConfiguration)
```
srknzl, line 328: I prefer a more detailed error here like in Java: 

```
Reconciliation interval can be at least %s seconds if it is not zero, but %d was configured."
                            + " Note: Configuring a value of zero seconds disables the reconciliation task.
```

nearcache.go
```diff
@@ -0,0 +1,1465 @@
...
+		rc = adapter
+		se = adapter
+	}
+	nc := &nearCache{
+		cfg:    cfg,
+		store:  newNearCacheRecordStore(cfg, ss, rc, se),
+		logger: logger,
+		doneCh: doneCh,
+	}
+	go nc.startExpirationTask(expirationInitialDelay, expirationPeriod)
```
srknzl, line 143: I think we should start this only if this condition is true: https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/DefaultNearCache.java#L103

internal/nearcache/codec.go
```diff
@@ -0,0 +1,42 @@
+/*
```
srknzl, line 1: Isn't this autogenerated as well? Why is this not in proto/codec?

nearcache_map.go
```diff
@@ -21,23 +21,39 @@ import (
...
 		ss: ss,
+		lg: lg,
 	}
 	if ncc.InvalidateOnChange() {
-		ncm.registerInvalidationListener()
+		lg.Debug(func() string {
+			return fmt.Sprintf("registering invalidation listener: name: %s, local: %t", name, local)
+		})
+		if err := ncm.registerInvalidationListener(ctx, name, p, local); err != nil {
+			return nearCacheMap{}, fmt.Errorf("hazelcast.newNearCacheMap: preloading near cache: %w", err)
```
srknzl, line 55: Is this error message correct?

nearcache_map.go
```diff
@@ -21,23 +21,39 @@ import (
...
+	inearcache "github.com/hazelcast/hazelcast-go-client/internal/nearcache"
+	"github.com/hazelcast/hazelcast-go-client/internal/proto"
+	"github.com/hazelcast/hazelcast-go-client/internal/proto/codec"
 	"github.com/hazelcast/hazelcast-go-client/internal/serialization"
 	"github.com/hazelcast/hazelcast-go-client/nearcache"
+	"github.com/hazelcast/hazelcast-go-client/types"
+)
+
+const (
+	eventTypeInvalidation = 8
```
srknzl, line 34: This is 1 << 8 in Java client. https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/core/EntryEventType.java#L133

nearcache_map.go
```diff
@@ -61,8 +77,19 @@ func newNearCacheMap(nc *nearCache, ncc *nearcache.Config, ss *serialization.Ser
...
-func (ncm *nearCacheMap) registerInvalidationListener() {
-	fmt.Println("IMPLEMENT ME: registerInvalidationListener")
+func (ncm *nearCacheMap) registerInvalidationListener(ctx context.Context, name string, p *proxy, local bool) error {
+	// port of: com.hazelcast.client.map.impl.nearcache.NearCachedClientMapProxy#registerInvalidationListener
+	sid := types.NewUUID()
+	addMsg := codec.EncodeMapAddNearCacheInvalidationListenerRequest(name, eventTypeInvalidation, local)
+	handler := func(msg *proto.ClientMessage) {
+		switch msg.Type() {
+		case inearcache.EventIMapInvalidationMessageType:
+			ncm.handleInvalidationMsg(inearcache.DecodeInvalidationMsg(msg))
```
srknzl, line 87: Why don't we support batch invalidation? Also, 

where do we run this logic: (I think this one registers RepairingHandler in ReparingTask and returns it) https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/client/map/impl/nearcache/NearCachedClientMapProxy.java#L673

internal/check/validate.go
```diff
@@ -51,3 +48,28 @@ func WithinRangeInt32(n, start, end int32) error {
...
+		return fmt.Errorf("non-negative integer expected: %w", err)
+	}
+	if n > math.MaxInt32 {
+		return fmt.Errorf("integer overflows int32: %w", err)
+	}
+	return nil
+}
+
+func ensureNonNegativeInt64(n int64, err error) error {
+	if n < 0 {
```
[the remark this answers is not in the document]
yuce: That's right.

internal/it/map.go
```diff
@@ -21,66 +21,128 @@ import (
...
 		}
-		config.Cluster.Unisocket = !smart
-		ls := "smart"
-		if !smart {
-			ls = "unisocket"
+		if tcx.ConfigCallback != nil {
+			tcx.ConfigCallback(tcx)
+		}
+		tcx.T.Logf("map name: %s", tcx.MapName)
+		tcx.T.Logf("cluster address: %s", tcx.Config.Cluster.Network.Addresses[0])
```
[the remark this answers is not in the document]
yuce: No, they are not leftovers.

client.go
```diff
@@ -67,6 +68,9 @@ type Client struct {
 	lifecycleListenerMapMu  *sync.Mutex
 	ic                      *client.Client
 	sqlService              isql.Service
+	nearCacheMgrsMu         *sync.RWMutex
+	nearCacheMgrs           map[string]*nearCacheManager
```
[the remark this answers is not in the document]
yuce: `nearcache.Manager` has the `state int32` field. If we make `nearcache.Manager` non-pointer, then we would update only the copy of that on certain operations, like after getting it from the map.

config.go
yuce: I think that belongs to the documentation. Also the user can look up their own configuration to see what's wrong.

nearcache_map.go
yuce: That's right, fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/adf231ea038a3c61a9ca428bc8a870b60458f291

nearcache.go
yuce: Fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/adf231ea038a3c61a9ca428bc8a870b60458f291

nearcache/config.go
```diff
@@ -29,6 +30,7 @@ const (
 	defaultStoreInitialDelaySeconds = 600
 	defaultStoreIntervalSeconds     = 600
 	defaultMemoryFormat             = InMemoryFormatBinary
+	// see: com.hazelcast.internal.nearcache.impl.invalidation.RepairingTask#MAX_TOLERATED_MISS_COUNT
```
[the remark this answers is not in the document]
yuce: It was a comment for another constant. It seems I've missed that removing that one. https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/adf231ea038a3c61a9ca428bc8a870b60458f291

internal/nearcache/codec.go
yuce: No, it's not autogenerated on the Java client as well.

nearcache_map.go
yuce: Thanks, fixed this.

nearcache_map.go
yuce: We support that, but not at the time of your comment :)

config.go
yuce: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/ebcf2b3b7fe69e1af93821996e0b1ea1d524c69d

nearcache.go
yuce: Really fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/ebcf2b3b7fe69e1af93821996e0b1ea1d524c69d

nearcache_map.go
yuce: It's at: nearcache_map.go:105 ( in `nearCacheMap.registerInvalidationListener` )

nearcache_map.go
yuce: Fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/ebcf2b3b7fe69e1af93821996e0b1ea1d524c69d

nearcache_map.go
```diff
@@ -62,8 +84,52 @@ func newNearCacheMap(nc *inearcache.NearCache, ncc *nearcache.Config, ss *serial
...
-	fmt.Println("IMPLEMENT ME: registerInvalidationListener")
+func (ncm *nearCacheMap) Destroy(ctx context.Context, name string) error {
+	ncm.lg.Trace(func() string {
+		return fmt.Sprintf("hazelcast.nearCacheMap.Destroy: %s", name)
+	})
+	s := ncm.invalidationListenerID.Load()
+	if s == nil {
+		return nil
+	}
+	// removeNearCacheInvalidationListener
```
srknzl, line 95: removeNearCacheInvalidationListener method in Java has these lines as well so it may be reasonable to move the comment:
```suggestion
	// removeNearCacheInvalidationListener
	s := ncm.invalidationListenerID.Load()
	if s == nil {
		return nil
	}
```

nearcache_map.go
```diff
@@ -62,8 +84,52 @@ func newNearCacheMap(nc *inearcache.NearCache, ncc *nearcache.Config, ss *serial
...
+	sid := s.(types.UUID)
+	ncm.rt.DeregisterHandler(name)
+	return ncm.lb.Remove(ctx, sid)
```
srknzl, line 98: Java client seems to destroy near cache here as well. https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/client/map/impl/nearcache/NearCachedClientMapProxy.java#L582

config.go
srknzl: Right but Java throws this exception see:
 https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/invalidation/RepairingTask.java#L114

I prefer verbose errors in general

client.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
+ * Copyright (c) 2008-2022, Hazelcast, Inc. All Rights Reserved.
```
srknzl, line 2: Some copyrights are 2021 some are 2022 (I think all new added files), is that intentional. Will you update others later?

internal/it/nearcache_util.go
```diff
@@ -139,9 +140,29 @@ func (tcx *NearCacheTestContext) AssertNearCacheContent(size int64, valueFmt str
...
-			tcx.T.FailNow()
-		}
-		// TODO: assertNearCacheRecord
+		require.Equal(tcx.T, target, value)
+		tcx.AssertNearCacheRecord(getRecordFromNearCache(nca, i), i, tcx.Config.InMemoryFormat)
 	}
 }
+
+func (tcx *NearCacheTestContext) AssertNearCacheRecord(rec *inearcache.Record, key int64, imf nearcache.InMemoryFormat) {
+	// see: com.hazelcast.internal.nearcache.impl.NearCacheTestUtils#assertNearCacheReference
```
srknzl, line 149: My guess is it should be as follows: 

```suggestion
	// see: com.hazelcast.internal.nearcache.impl.NearCacheTestUtils#assertNearCacheRecord
```

nearcache_map.go
```diff
@@ -62,8 +84,52 @@ func newNearCacheMap(nc *inearcache.NearCache, ncc *nearcache.Config, ss *serial
...
+}
+
+func (ncm *nearCacheMap) registerInvalidationListener(ctx context.Context, name string, local bool) error {
+	// port of: com.hazelcast.client.map.impl.nearcache.NearCachedClientMapProxy#registerInvalidationListener
+	sid := types.NewUUID()
```
srknzl, line 103: I believe it will be easier to read if you move sid variable to a place closer to where it is used. (above `removeMsg := codec.EncodeMapRemoveEntryListenerRequest(name, sid)`)

internal/nearcache/manager.go
```diff
@@ -47,9 +74,23 @@ func (m *Manager) GetOrCreateNearCache(name string, cfg nearcache.Config) *NearC
...
 	}
 	m.nearCachesMu.Unlock()
 	return nc
 }
+
+func (m *Manager) RepairingTask() *ReparingTask {
+	return m.rt
+}
+
+func (m *Manager) GetNearCacheStats() []proto.Pair {
```
srknzl, line 88: Why not return the exact type but []proto.Pair here?
Something like this:

`func (m *Manager) GetNearCacheStats()  []NearCacheStat`
```
struct NearCacheStat {
    name string
    stats nearcache.Stats
}
```

client.go
```diff
@@ -92,16 +95,25 @@ func newClient(config Config) (*Client, error) {
...
 		cfg:                     &config,
 	}
+	if c.ic.StatsService != nil {
+		c.ic.StatsService.SetNCStatsGetter(func(service string) stats.NearCacheStatsGetter {
+			ncmgr, ok := c.nearCacheMgrs[service]
+			if !ok {
+				return nil
+			}
+			return ncmgr
+		})
```
srknzl, line 115: We can also consider setting this in createComponents, in `if config.StatsEnabled {` block.

client.go
```diff
@@ -139,7 +151,9 @@ func (c *Client) GetMap(ctx context.Context, name string) (*Map, error) {
 		// there is a near cache config for this map
 		ncmgr := c.getNearCacheManager(ServiceNameMap)
 		nc := ncmgr.GetOrCreateNearCache(name, ncc)
-		m.ncm, err = newNearCacheMap(nc, &ncc, c.ic.SerializationService)
+		ss := c.ic.SerializationService
+		rt := ncmgr.RepairingTask()
```
srknzl, line 155: As far as I see in Java only one repairing task is created per data structure type. Can we do the same?  see https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/invalidation/RepairingTask.java#L45 and https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/client/impl/spi/ClientContext.java#L117

proxy_manager.go
```diff
@@ -165,13 +165,19 @@ func (m *proxyManager) removeDistributedObjectEventListener(ctx context.Context,
...
-	if _, ok := m.proxies[name]; !ok {
+	p, ok := m.proxies[name]
+	if !ok {
 		return false
 	}
+	// run the local destroy method of Map
+	if serviceName == ServiceNameMap {
+		mp := p.(*Map)
+		mp.destroyLocally(ctx)
+	}
```
srknzl, line 180: Instead of this servicename check we can also do a conditional cast `mp, ok := p.(*Map)` then `if ok { destroyLocally()`

nearcache/types.go
```diff
@@ -64,34 +64,36 @@ func (p EvictionPolicy) String() string {
 
 // Stats contains statistics for a Near Cache instance.
 type Stats struct {
```
srknzl, line 66: I see some failures here about field alignment

internal/stats/stats_service.go
```diff
@@ -424,6 +439,43 @@ func (g gaugeOS) updateDescr(bt *binTextStats) {
 	bt.stats = append(bt.stats, makeTextStat(&g.openDecrCount, nfd))
 }
 
+type gaugeNearCache struct {
+	serviceName string
+	f           func() func(service string) NearCacheStatsGetter
```
srknzl, line 444: Why don't we directly store `func(service string) NearCacheStatsGetter` here?

internal/stats/stats_service.go
```diff
@@ -460,6 +512,46 @@ func makePercentMD(prefix, metric string) metricDescriptor {
...
+func makeNearCacheCountMD(ncName, metric string) metricDescriptor {
+	return metricDescriptor{
+		Prefix:             nearCacheDescriptorPrefix,
+		Metric:             metric,
+		Discriminator:      nearCacheDescriptorDiscriminator,
+		DiscriminatorValue: ncName,
+		HasUnit:            true,
+		Unit:               metricUnitCount,
+	}
+}
```
srknzl, line 535: Alternative: send `Unit` as a parameter and remove one of the two methods.

internal/stats/stats_service.go
```diff
@@ -460,6 +512,46 @@ func makePercentMD(prefix, metric string) metricDescriptor {
...
+	"hits":                        func(s nearcache.Stats) int64 { return s.Hits },
+	"lastPersistenceKeyCount":     func(s nearcache.Stats) int64 { return s.LastPersistenceKeyCount },
+	"lastPersistenceWrittenBytes": func(s nearcache.Stats) int64 { return s.LastPersistenceWrittenBytes },
+	"misses":                      func(s nearcache.Stats) int64 { return s.Misses },
+	"ownedEntryCount":             func(s nearcache.Stats) int64 { return s.OwnedEntryCount },
+	"expirations":                 func(s nearcache.Stats) int64 { return s.Expirations },
+	"invalidations":               func(s nearcache.Stats) int64 { return s.Invalidations },
+	"invalidationRequests":        func(s nearcache.Stats) int64 { return s.InvalidationRequests },
+	"ownedEntryMemoryCost":        func(s nearcache.Stats) int64 { return s.OwnedEntryMemoryCost },
+}
```
srknzl, line 557: lastPersistenceFailure is missing as far as I see. Will you add it later?

internal/stats/stats_service.go
```diff
@@ -424,6 +439,43 @@ func (g gaugeOS) updateDescr(bt *binTextStats) {
 	bt.stats = append(bt.stats, makeTextStat(&g.openDecrCount, nfd))
 }
 
+type gaugeNearCache struct {
+	serviceName string
+	f           func() func(service string) NearCacheStatsGetter
+}
+
+func newGaugeNearCache(svc string, f func() func(service string) NearCacheStatsGetter) gaugeNearCache {
```
srknzl, line 447: Don't we need to send near cache stats as basic stats as well? Java sends in both formats as far as I see.

internal/nearcache/record.go
```diff
@@ -24,26 +24,25 @@ import (
 )
 
 type Record struct {
-	CreationTime         int64
-	lastAccessTime       int64
-	ExpirationTime       int64
-	InvalidationSequence int64
+	invalidationSequence int64
 	reservationID        int64
 	value                internal.AtomicValue
```
srknzl, line 29: Can we use atomic.Value instead of this? Why did we need a new type?

internal/nearcache/record.go
```diff
@@ -74,27 +73,50 @@ func (r *Record) SetReservationID(rid int64) {
...
 }
 
-func (r *Record) LastAccessTimeMS() int64 {
-	return atomic.LoadInt64(&r.lastAccessTime)
+func (r *Record) CreationTime() int64 {
+	t := atomic.LoadInt32(&r.creationTime)
+	return RecomputeWithBaseTime(t)
+}
+
+func (r *Record) SetCreationTime(ms int64) {
```
srknzl, line 81: Some fields including creation time are not updated atomically in Java. Why did you update them atomically?

internal/nearcache/stale_read_detector.go
```diff
@@ -0,0 +1,49 @@
...
+		rh: rh,
+		ps: ps,
+	}
+}
+
+func (sr StaleReadDetector) IsStaleRead(rec *Record) bool {
+	// port of: com.hazelcast.internal.nearcache.impl.invalidation.StaleReadDetectorImpl#isStaleRead
+	// key param in the original implementation is not used.
+	md := sr.rh.GetMetaDataContainer(rec.PartitionID())
+	return rec.UUID() != md.UUID() || rec.InvalidationSequence() < md.StaleSequence()
```
srknzl, line 40: uuid fields are atomic.Value right now in the go client. if there are no values stored in both uuid fields of `rec` and `md`,  `rec.UUID() != md.UUID()` will be false in go client, however `!rec.hasSameUuid(md)` in Java client will return true because it checks for null: https://github.com/hazelcast/hazelcast/blob/a0604ffa968a5d301b0b7c5adf1d1c8b887ac65b/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/record/AbstractNearCacheRecord.java#L169

So I guess we should fix this behaviour difference.

internal/nearcache/record_store.go
```diff
@@ -17,45 +17,80 @@
...
+	atomic.StoreInt64(&rs.stats.OwnedEntryCount, 0)
+	atomic.StoreInt64(&rs.stats.OwnedEntryMemoryCost, 0)
+	atomic.AddInt64(&rs.stats.Invalidations, int64(size))
+	rs.incrementInvalidationRequests()
+}
+
 func (rs *RecordStore) Get(key interface{}) (value interface{}, found bool, err error) {
 	// checkAvailable() does not apply since rs.records is always created
+	rs.recordsMu.RLock()
+	defer rs.recordsMu.RUnlock()
```
srknzl, line 93: I think we can shrink the locked range by moving these lines but I don't know if performance will be noticably better

internal/nearcache/record_store.go
```diff
@@ -116,6 +163,79 @@ func (rs *RecordStore) Invalidate(key interface{}) {
...
+	// assumes rs.recordsMu is locked elsewhere
+	if _, exists := rs.records[key]; !exists {
+		return false
+	}
+	delete(rs.records, key)
+	return true
+}
+
+func (rs *RecordStore) sample(count int) []evictionCandidate {
+	// port of: com.hazelcast.internal.nearcache.impl.store.HeapNearCacheRecordMap#sample
```
srknzl, line 220: Is this from https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/internal/util/SampleableConcurrentHashMap.java#L300

because from where you link I reached this point. Though, I did not see a similarity between your code and the code I give a link to

internal/nearcache/record_store.go
```diff
@@ -394,3 +566,75 @@ func (rs *RecordStore) updateRecordValue(rec *Record, value interface{}) error {
...
+}
+
+func evaluateForEviction(cmp nearcache.EvictionPolicyComparator, candies []evictionCandidate) evictionCandidate {
+	now := time.Now().UnixMilli()
+	var selected evictionCandidate
+	var hasSelected bool
+	for _, current := range candies {
+		// initialize selected by setting it to current candidate.
+		if !hasSelected {
+			selected = current
```
srknzl, line 626: Should you set hasSelected here as well? Also the name candies looks strange :D It means "şekerler"

internal/nearcache/repairing_task.go
```diff
@@ -0,0 +1,568 @@
...
+		ps:                          ps,
+		lg:                          lg,
+		handlers:                    &sync.Map{},
+		partitionCount:              ps.PartitionCount(),
+	}
+	return nc
+}
+
+func (rt *ReparingTask) start() {
+	// see: com.hazelcast.internal.nearcache.impl.invalidation.RepairingTask#scheduleNextRun
```
srknzl, line 75: I think `scheduleNextRun` schedules only one shot run not infinite runs like you did.

internal/nearcache/repairing_task.go
```diff
@@ -0,0 +1,568 @@
...
+}
+
+func (rt *ReparingTask) RegisterAndGetHandler(ctx context.Context, name string, nc *NearCache) (RepairingHandler, error) {
+	handler := NewRepairingHandler(name, nc, rt.partitionCount, rt.ss, rt.ps, rt.lg)
+	// ignoring the "loaded" return value
+	h, loaded := rt.handlers.LoadOrStore(name, handler)
+	if !loaded {
+		if err := rt.invalidationMetaDataFetcher.Init(ctx, handler); err != nil {
+			return RepairingHandler{}, nil
+		}
```
srknzl, line 99: This is done after setting staleReadDetector in Java. Just to be sure that there is no behaviour change can you do the same? Also, won't you implement the fallback initializer `initRepairingHandlerAsync`

internal/nearcache/repairing_task.go
```diff
@@ -0,0 +1,568 @@
...
+	if err := df.is.SendRequest(ctx, inv); err != nil {
+		return nil, err
+	}
+	return inv, nil
+}
+
+func (df *InvalidationMetaDataFetcher) processMemberMetadata(ctx context.Context, inv invocation.Invocation, handlers map[string]RepairingHandler) error {
+	// port of: com.hazelcast.internal.nearcache.impl.invalidation.InvalidationMetaDataFetcher#processMemberMetadata
+	// extractMemberMetadata
+	res, err := inv.GetWithContext(ctx)
```
srknzl, line 517: Java waits for 1 minute for this async result [here](https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/client/map/impl/nearcache/invalidation/ClientMapInvalidationMetaDataFetcher.java#L68). What will go client do?

internal/nearcache/repairing_task.go
```diff
@@ -0,0 +1,568 @@
...
+	for i := 0; i < sz; i++ {
+		if err := h.Handle(keys[i], sources[i], partitions[i], seqs[i]); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (h *RepairingHandler) CheckOrRepairUUID(partition int32, new types.UUID) {
+	// this method may be called concurrently: anti-entropy, event service
```
srknzl, line 298:
```suggestion
	// this method may be called concurrently by anti-entropy and event service
```

internal/nearcache/repairing_task.go
```diff
@@ -0,0 +1,568 @@
...
+		panic("CheckOrRepairUUID: new UUID should not be the default UUID")
+	}
+	md := h.GetMetaDataContainer(partition)
+	for {
+		prev := md.UUID()
+		if prev == new {
+			break
+		}
+		if md.CASUUID(prev, new) {
+			md.ResetStaleSequence()
```
srknzl, line 309:
```suggestion
			md.ResetSequence()
```

internal/nearcache/repairing_task.go
```diff
@@ -0,0 +1,568 @@
...
+	npsPairs, psPairs := codec.DecodeMapFetchNearCacheInvalidationMetadataResponse(res)
+	// repairUuids
+	// see: com.hazelcast.internal.nearcache.impl.invalidation.InvalidationMetaDataFetcher#repairUuids
+	for _, p := range psPairs {
+		k := p.Key.(int32)
+		v := p.Value.(types.UUID)
+		for _, handler := range handlers {
+			handler.CheckOrRepairUUID(k, v)
+		}
+	}
```
srknzl, line 530: We can make these separate methods as well.

internal/nearcache/repairing_task.go
```diff
@@ -0,0 +1,568 @@
...
+				}
+				di--
+			}
+			mems[i], mems[di] = mems[di], mems[i]
+		}
+	}
+	// all deleted items are at the end.
+	// shrink the slice to get rid of them.
+	return mems[:di]
+}
```
srknzl, line 568: I don't know if this logic will work or not. There won't be millions of members, why not create a new slice and do a basic for loop and if check?

internal/nearcache/repairing_task.go
```diff
@@ -0,0 +1,568 @@
...
+			})
+			break
+		}
+	}
+}
+
+func (h *RepairingHandler) CheckOrRepairSequence(partition int32, nextSeq int64, viaAntiEntropy bool) {
+	if nextSeq <= 0 {
+		panic("CheckOrRepairSequence <= 0")
+	}
```
srknzl, line 323: Same comment about assertion. Will java client throw in runtime?

internal/nearcache/manager.go
yuce: `stats.Service` should not import `inearcache`, since that creates an import cycle. I would have to put `NearCacheStat` somewhere else, unrelated. I think current approach is OK for now. It is used in a single place and not in the hot path.

internal/stats/stats_service.go
yuce: `f` is called before `hazelcast.Client` is initialized and has a chance to set the correct `NearCacheStatsGetter`.

internal/stats/stats_service.go
yuce: That's not sent with the binary statistics.

internal/stats/stats_service.go
yuce: Yes, but that's required only for 4.x. Since that's out of support, sending only the binary stats.

internal/nearcache/record.go
yuce: Java client uses `volatile` for those, which is not available for Go.

internal/nearcache/record_store.go
```diff
@@ -67,23 +102,21 @@ func (rs *RecordStore) Get(key interface{}) (value interface{}, found bool, err
...
-	/*
-	   if (staleReadDetector.isStaleRead(key, record)) {
-	       invalidate(key);
-	       nearCacheStats.incrementMisses();
-	       return null;
-	   }
-	*/
+	// instead of ALWAYS_FRESH staleReadDetector, the nil value used
+	if rs.staleReadDetector != nil && rs.staleReadDetector.IsStaleRead(rec) {
+		rs.invalidate(key)
```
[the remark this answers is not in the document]
yuce: That would cause a deadlock in the current code. But using an RLock here is already wrong. Let me think about this a bit.

internal/nearcache/record_store.go
yuce: No, the code above is port of com.hazelcast.internal.eviction.impl.evaluator.EvictionPolicyEvaluator#evaluate I'll add a comment

internal/nearcache/repairing_task.go
yuce: Thanks. It doesn't have a timeout, better to add one.

client.go
yuce: That's not intentional, but I don't think it's very important. We can update the copyright headers altogether after we merge the feature branches.

internal/it/nearcache_util.go
yuce: Fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/0d49f98e13b8fc0d4843db1366d353fb6a96a174

nearcache_map.go
yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/0d49f98e13b8fc0d4843db1366d353fb6a96a174

internal/stats/stats_service.go
yuce: Added another helper at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/0d49f98e13b8fc0d4843db1366d353fb6a96a174

internal/nearcache/record_store.go
yuce: Adde an explanation at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/0d49f98e13b8fc0d4843db1366d353fb6a96a174

internal/nearcache/record_store.go
yuce: Added at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/0d49f98e13b8fc0d4843db1366d353fb6a96a174

internal/nearcache/repairing_task.go
```diff
@@ -0,0 +1,568 @@
...
+		if err := h.Handle(keys[i], sources[i], partitions[i], seqs[i]); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (h *RepairingHandler) CheckOrRepairUUID(partition int32, new types.UUID) {
+	// this method may be called concurrently: anti-entropy, event service
+	if new.Default() {
```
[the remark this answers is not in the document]
yuce: Replaced at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/0d49f98e13b8fc0d4843db1366d353fb6a96a174

internal/nearcache/repairing_task.go
yuce: Fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/0d49f98e13b8fc0d4843db1366d353fb6a96a174

internal/nearcache/repairing_task.go
yuce: OK: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/0d49f98e13b8fc0d4843db1366d353fb6a96a174

internal/nearcache/repairing_task.go
yuce: Replaced at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/0d49f98e13b8fc0d4843db1366d353fb6a96a174

proxy_map.go
```diff
@@ -1160,6 +1160,17 @@ func (m *Map) LocalMapStats() LocalMapStats {
 	return LocalMapStats{}
 }
 
+func (m *Map) destroyLocally(ctx context.Context) {
+	m.logger.Trace(func() string {
+		return fmt.Sprintf("hazelcast.Map.destroyLocally: %s", m.name)
+	})
+	if m.hasNearCache {
```
utku-caglayan, line 1167: I think the current state where we access a proxy, checking whether it is a map, checking whether it has near cache is too specific. Can we make these an interface or a field in the proxy struct?

What I want to say is that it would be great to make these part of the general flow. Wdyt?

internal/nearcache/repairing_task.go
```diff
@@ -0,0 +1,576 @@
...
+			h.name, partition, md.StaleSequence(), md.Sequence())
+	})
+}
+
+// Handle handles a single invalidation.
+func (h *RepairingHandler) Handle(key serialization.Data, source, partition types.UUID, seq int64) error {
+	// port of: com.hazelcast.internal.nearcache.impl.invalidation.RepairingHandler#handle(com.hazelcast.internal.serialization.Data, java.util.UUID, java.util.UUID, long)
+	// Apply invalidation if it's not originated by local member/client.
+	// Since local Near Caches are invalidated immediately there is no need to invalidate them twice.
+	if source != partition {
```
srknzl, line 264: In java this if is comparing localUUid (client uuid) with source uuid

internal/nearcache/repairing_task.go
```diff
@@ -0,0 +1,576 @@
...
+	df := InvalidationMetaDataFetcher{
+		cs:         cs,
+		is:         is,
+		invFactory: invFactory,
+		lg:         lg,
+	}
+	return df
+}
+
+func (df InvalidationMetaDataFetcher) Init(ctx context.Context, handler RepairingHandler) error {
```
srknzl, line 460: Are you sure this function correctly ports ? com.hazelcast.internal.nearcache.impl.invalidation.InvalidationMetaDataFetcher#init

internal/nearcache/comparators_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (s simpleEntry) CreationTime() int64 {
+	return s.creationTime
+}
+
+func (s simpleEntry) LastAccessTime() int64 {
+	return s.lastAccessTime
+}
+
+func TestLRUComparatorDoesNotPrematurelySelectNewlyCreatedEntries(t *testing.T) {
```
srknzl, line 55: Can you write where this is ported from as a comment?

internal/nearcache/comparators_test.go
```diff
@@ -0,0 +1,145 @@
...
+	// 1. Create expected list of ordered elements by
+	// sorting entries based on their idle-times. Longest
+	// idle time must be the first element of the list.
+	// 2. Then sort given entries by using LRU eviction comparator.
```
srknzl, line 59: I think we can move these comments into evictionPolicyHelper as it fits better. Also you can add the step 3 as well.

utku-caglayan (commented):
Will continue

internal/nearcache/record_store.go
```diff
@@ -228,7 +352,9 @@ func (rs *RecordStore) getRecordStorageMemoryCost(rec *Record) int64 {
 }
```
srknzl, line 352: can you remove the todo in  getRecordStorageMemoryCost if it is not necessary ?

nearcache/nearcache_it_test.go
```diff
@@ -34,8 +36,120 @@ import (
...
+			ncc := nearcache.Config{Name: tcx.MapName}
+			ncc.SetInvalidateOnChange(true)
+			tcx.Config.AddNearCache(ncc)
+		},
+	}
+	tcx.Tester(func(tcx it.MapTestContext) {
+		m := tcx.M
+		t := tcx.T
+		ctx := context.Background()
+		const mapSize = 1
```
srknzl, line 58: Using 1 in mapSize and maxCacheSize looks ok because we call MapSetOnServer, however such a low number may miss some errors. What about something bigger like 3, 5?

nearcache/nearcache_it_test.go
```diff
@@ -325,5 +448,42 @@ func newNearCacheMapTestContext(t *testing.T, fmt nearcache.InMemoryFormat, inva
...
+		}
+		if stats.Expirations != int64(size) {
+			return false
+		}
+		if stats.Evictions != 0 {
+			return false
+		}
+		if !assert.Equal(t, int64(0), stats.Evictions) {
+			return false
+		}
```
srknzl, line 486: These two if looks duplicates to me.

internal/nearcache/record_store.go
```diff
@@ -51,11 +67,29 @@ func NewRecordStore(cfg *nearcache.Config, ss *serialization.Service, rc nearCac
...
+	atomic.StoreInt64(&rs.stats.OwnedEntryCount, 0)
+	atomic.StoreInt64(&rs.stats.OwnedEntryMemoryCost, 0)
+	atomic.AddInt64(&rs.stats.Invalidations, int64(size))
+	rs.incrementInvalidationRequests()
+}
+
 func (rs *RecordStore) Get(key interface{}) (value interface{}, found bool, err error) {
 	// checkAvailable() does not apply since rs.records is always created
+	rs.recordsMu.Lock()
+	defer rs.recordsMu.Unlock()
```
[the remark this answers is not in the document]
utku-caglayan: Old comment nvm

client.go
utku-caglayan: As far as I can trace, go client can create multiple repair tasks for map data type. Am I wrong 🤔

utku-caglayan: Maybe we should pass the service name to `GetOrCreateNearCache` to replicate the same behaviour in `nearcache/manager.go`

internal/nearcache/nearcache.go
```diff
@@ -102,3 +137,23 @@ func (nc *NearCache) checkKeyFormat(key interface{}) {
 		panic("key cannot be of type serialization.Data!")
 	}
 }
+
+func (nc *NearCache) startExpirationTask(delay, timeout time.Duration) {
+	time.Sleep(delay)
+	timer := time.NewTicker(timeout)
```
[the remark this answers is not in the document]
utku-caglayan: I traced the Java client executor implementations for this and found out that
1) DelegatingTaskScheduler::scheduleWithRepetition
2)ScheduledThreadPoolExecutor::scheduleAtFixedRate 

First one skips the tick and does not execute even when the time comes, but timer keeps on ticking. Let's say task takes 3 seconds, and the period(timeout) is 2
[task] [task]     [ task]
1 2  3  4 5 6  7  8 9 10

I'm not sure about the behaviour of `ScheduledThreadPoolExecutor`, but this implementation does not satisfy the first one IMO

client.go
srknzl: > It's the same with the Go client. Currently only one repairing task is created when the near cache manager is initialized.

You are right. 

> As far as I can trace, go client can create multiple repair tasks for map data type

Can you share the scenario? I thought a new reparing task is being created in this line but it is not. 

so, resolving the comment

internal/stats/stats_service.go
srknzl: So if you move `c.ic.StatsService.SetNCStatsGetter` call into the if block (`if config.StatsEnabled {` in internal/client/client.go) we can simplify here as well. We can directly pass the function to the gaugeNearCache.

internal/stats/stats_service.go
srknzl: ClientMetricsConfig class has the annotation `@since 4.0`. So is it added in 4.0? Why do we send basic stats in our clients anyway? 🤔 Good point to discuss.

internal/nearcache/record_store.go
srknzl: Explanation: So, you are using hasSelected instead of the `null` value. So when selected is assigned to a nonnull value in Java code, you should set hasSelected to `true`.  I am assuming `current` variable in `for (C current : candidates) {` is not `null` though. As far as I understand, `sample()` method does not return `null`.

internal/nearcache/repairing_task.go
srknzl: > Java client re-schedules after the handler is run, almost the same thing. See: com.hazelcast.spi.impl.executionservice.TaskScheduler#schedule(java.lang.Runnable, long, java.util.concurrent.TimeUnit) which schedules run after one second which in turn calls scheduleNextRun again.

yeah I noticed that you schedule a bit different than Java. I am worried if there is a behavior difference. I did not notice but there may be something I missed. 

> Also we do not have this protection, with saying that I have no clue what it serves https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/invalidation/RepairingTask.java#L130

Seems like running always true after it is set to true, so the check is always true.

internal/nearcache/repairing_task.go
srknzl: Even there are thousands of members I don't see a noticable benefit in saving memory (since your algorithm is in-place). As far as the time complexity both are the same. The benefit of my suggestion is, obviously, readability.

internal/nearcache/repairing_task.go
yuce: I changed this bit at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/c67bb647844dec1c173a226819e5f398deac8fe2

internal/nearcache/repairing_task.go
yuce: Added at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/4626aca1993000c79bfaaaf4acecca604b3d16e2

internal/nearcache/repairing_task.go
```diff
@@ -0,0 +1,576 @@
...
+	}
+	h.CheckOrRepairUUID(pid, partition)
+	h.CheckOrRepairSequence(pid, seq, false)
+	return nil
+}
+
+// HandleBatch handles a batch of validations.
+func (h *RepairingHandler) HandleBatch(keys []serialization.Data, sources []types.UUID, partitions []types.UUID, seqs []int64) error {
+	// port of: com.hazelcast.internal.nearcache.impl.invalidation.RepairingHandler#handle(java.util.Collection<com.hazelcast.internal.serialization.Data>, java.util.Collection<java.util.UUID>, java.util.Collection<java.util.UUID>, java.util.Collection<java.lang.Long>)
+	// assumes len(keys) == len(source) -- len(partitions) == len(seqs)
```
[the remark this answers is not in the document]
yuce: There is a set of sequences per partition, and each key has a partition. Source is the member UUID that sends the invalidation for that key. I don't see how handling this invalidation event would work if the assumption above doesn't hold. Here is the relevant TDD: https://hazelcast.atlassian.net/wiki/spaces/HZC/pages/1364131869/Non-Java+Clients+Eventually+Nearcache+Design+Document (especially the Invalidation section).

Note that the `Collection` interface doesn't have an index method, so it is not possible to get the keys etc. by index, unless it is converted to an array.

internal/nearcache/repairing_task.go
yuce: I've refactored this one at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/4626aca1993000c79bfaaaf4acecca604b3d16e2

config.go
```diff
@@ -269,6 +281,63 @@ func (f *FlakeIDGeneratorConfig) Validate() error {
...
+	}
+	return *pc.maxToleratedMissCount
+}
+
+func (pc *InvalidationConfig) SetReconciliationIntervalSeconds(seconds int) {
+	if err := check.NonNegativeInt32Config(seconds); err != nil {
+		pc.err = fmt.Errorf("ReconciliationIntervalSeconds: %w", err)
+		return
+	}
+	if seconds != 0 && seconds < defaultMinReconciliationIntervalSeconds {
```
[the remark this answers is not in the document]
yuce: The user cannot set `seconds` other than by using `SetReconciliationIntervalSeconds` and errors in that method are checked at `Validate`. Am I missing anything?

internal/it/nearcache_util.go
```diff
@@ -139,9 +140,29 @@ func (tcx *NearCacheTestContext) AssertNearCacheContent(size int64, valueFmt str
...
+	t := tcx.T
+	require.NotNil(t, rec)
+	require.Equal(t, inearcache.RecordReadPermitted, rec.ReservationID())
+	value := rec.Value()
+	switch imf {
+	case nearcache.InMemoryFormatBinary:
+		require.Equal(t, reflect.TypeOf(serialization.Data{}), reflect.TypeOf(value))
```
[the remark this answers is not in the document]
yuce: Good idea, but there doesn't seem to be the `require.IsNotType` method which is required in the second case statement. So I have to use something else for the second one. Let's keep it if you don't have a strong opinion against it.

internal/stats/stats_service.go
```diff
@@ -38,9 +38,21 @@ import (
 	"github.com/hazelcast/hazelcast-go-client/internal/event"
 	"github.com/hazelcast/hazelcast-go-client/internal/invocation"
 	"github.com/hazelcast/hazelcast-go-client/internal/logger"
+	"github.com/hazelcast/hazelcast-go-client/internal/proto"
 	"github.com/hazelcast/hazelcast-go-client/internal/proto/codec"
+	"github.com/hazelcast/hazelcast-go-client/nearcache"
 )
 
+const (
+	serviceNameMap                   = "hz:impl:mapService"
```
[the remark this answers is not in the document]
yuce: Yes, it introduces a cyclic import.

proxy_map.go
yuce: We can consider that if we implement the support for Replicated Map. There are no other data structures that support the near cache.

nearcache/nearcache_it_test.go
yuce: The original value for that is 1000, changed that and forgot to set it back: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/22d0f4598f4b66079464377bdbe5e4aee0db7a09

internal/nearcache/record_store.go
yuce: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/22d0f4598f4b66079464377bdbe5e4aee0db7a09

nearcache/types.go
```diff
@@ -64,34 +64,36 @@ func (p EvictionPolicy) String() string {
...
 	// LastPersistenceWrittenBytes is number of bytes written in the last persistence task when the pre-load feature is enabled.
 	LastPersistenceWrittenBytes int64
 	// PersistenceCount is the number of completed persistence tasks when the pre-load feature is enabled.
 	PersistenceCount int64
-	// Hits is the number of times an entry was found in the Near Cache.
-	Hits int64
+	// CreationTime is the time the Near Cache was initialized.
+	CreationTime time.Time
+	// LastPersistenceTime is the time of the last completed persistence task when the pre-load feature is enabled.
+	LastPersistenceTime time.Time
```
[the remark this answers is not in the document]
yuce: The TDD for this is already approved. Not sure having MS has any advantage over `time.Time`.

nearcache/config.go
```diff
@@ -28,7 +29,6 @@ const (
 	defaultEvictionPolicy           = EvictionPolicyLRU
 	defaultStoreInitialDelaySeconds = 600
 	defaultStoreIntervalSeconds     = 600
-	defaultMemoryFormat             = InMemoryFormatBinary
```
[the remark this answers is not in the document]
yuce: Probably it wasn't useful.

internal/nearcache/store_adapter_test.go
```diff
@@ -0,0 +1,125 @@
...
+			name:      "data estimator: nil record",
+			estimator: dataEstimator,
+			makeRec: func() *Record {
+				return nil
+			},
+			cost:   0,
+			cost32: 0,
+		},
+		{
+			name:      "data estimator: record with nil value",
```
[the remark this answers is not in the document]
yuce: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/22d0f4598f4b66079464377bdbe5e4aee0db7a09

internal/nearcache/nearcache.go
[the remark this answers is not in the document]
yuce: Here's an attempt to simulate that: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/22d0f4598f4b66079464377bdbe5e4aee0db7a09 Could you check the test to see whether it matches what you expect?

internal/nearcache/codec.go
```diff
@@ -0,0 +1,52 @@
...
+
+const (
+	EventIMapInvalidationMessageType              = 81666
+	EventIMapBatchInvalidationMessageType         = 81667
+	eventIMapInvalidationSourceUUIDFieldOffset    = proto.PartitionIDOffset + proto.IntSizeInBytes
+	eventIMapInvalidationPartitionUUIDFieldOffset = eventIMapInvalidationSourceUUIDFieldOffset + proto.UUIDSizeInBytes
+	eventIMapInvalidationSequenceFieldOffset      = eventIMapInvalidationPartitionUUIDFieldOffset + proto.UUIDSizeInBytes
+)
+
+func DecodeInvalidationMsg(msg *proto.ClientMessage) (key serialization.Data, source types.UUID, partition types.UUID, seq int64) {
```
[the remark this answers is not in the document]
yuce: Renamed them at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/22d0f4598f4b66079464377bdbe5e4aee0db7a09

internal/nearcache/nearcache_internal_test.go
```diff
@@ -0,0 +1,197 @@
...
+				{UUID: uuids[5], LiteMember: false},
+				{UUID: uuids[6], LiteMember: false},
+			},
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			r := filterDataMembers(tc.mems)
+			// sort by UUID
+			sort.SliceStable(r, func(i, j int) bool {
```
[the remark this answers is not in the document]
yuce: Good idea: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/22d0f4598f4b66079464377bdbe5e4aee0db7a09

nearcache_map.go
```diff
@@ -62,8 +84,52 @@ func newNearCacheMap(nc *inearcache.NearCache, ncc *nearcache.Config, ss *serial
...
+		return nil
+	}
+	// removeNearCacheInvalidationListener
+	sid := s.(types.UUID)
+	ncm.rt.DeregisterHandler(name)
+	return ncm.lb.Remove(ctx, sid)
+}
+
+func (ncm *nearCacheMap) registerInvalidationListener(ctx context.Context, name string, local bool) error {
+	// port of: com.hazelcast.client.map.impl.nearcache.NearCachedClientMapProxy#registerInvalidationListener
```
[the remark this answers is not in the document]
yuce: I've changed the code to not return an error on registration, but log it: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/d6d75b4566197995c50ed6a08eab6176e2479712

nearcache/nearcache_it_test.go
yuce: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/d6d75b4566197995c50ed6a08eab6176e2479712

internal/nearcache/repairing_task.go
```diff
@@ -0,0 +1,568 @@
...
+		if err := rt.invalidationMetaDataFetcher.Init(ctx, handler); err != nil {
+			return RepairingHandler{}, nil
+		}
+		sr := NewStaleReadDetector(handler, rt.ps)
+		nc.store.staleReadDetector = &sr
+	}
+	if atomic.CompareAndSwapInt32(&rt.running, 0, 1) {
+		// this is the first added handler
+		go rt.start()
+		atomic.StoreInt64(&rt.lastAntiEntropyRunNanos, time.Now().UnixNano())
```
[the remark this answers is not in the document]
yuce: I think it's good practice to use atomics for a variable everywhere if it was used somewhere. Probably that code is transformed to the equivalent of `rt.lastAntiEntropyRunNanos = time.Now().UnixNano()` for most platforms.

internal/nearcache/repairing_task.go
yuce: I will keep the original one if you don't have a strong opinion about it.

internal/nearcache/repairing_task.go
yuce: This algorithm never creates a copy of the member list and the algorithm you suggest always creates a copy. Although the work done is similar, copying the member list means more work for the GC. That's probably trivial, but as long as the code is correct I would prefer the first one.

internal/nearcache/repairing_task.go
yuce: Fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/22d0f4598f4b66079464377bdbe5e4aee0db7a09

internal/nearcache/comparators_test.go
yuce: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/22d0f4598f4b66079464377bdbe5e4aee0db7a09

internal/nearcache/comparators_test.go
yuce: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/22d0f4598f4b66079464377bdbe5e4aee0db7a09

internal/nearcache/record_store.go
```diff
@@ -116,6 +163,79 @@ func (rs *RecordStore) Invalidate(key interface{}) {
...
+}
+
+func (rs *RecordStore) sample(count int) []evictionCandidate {
+	// port of: com.hazelcast.internal.nearcache.impl.store.HeapNearCacheRecordMap#sample
+	// note that count is fixed to 15 in the reference implementation, it is always positive
+	// assumes recordsMu is locked
+	samples := make([]evictionCandidate, count)
+	var idx int
+	for k, v := range rs.records {
+		// access to maps is random
```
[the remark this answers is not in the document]
yuce: It's definitely not cryptographically random (neither the Java implementation) but I think it's OK for this purpose. Here is a discussion about that at: https://stackoverflow.com/questions/41019703/is-map-iteration-sufficiently-random-for-randomly-selecting-keys

Slightly modified example on that page: https://go.dev/play/p/gsmgJxH76Gk

The alternative is keeping a separate list of keys and shuffle it which requires more memory or trying to reduce the distribution imbalance by oversampling and choosing the keys using another random method.

internal/nearcache/record_store.go
yuce: You guys are right, here's the update: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/24624ef616401d25ca45b672f2c963f2ff050f55

internal/nearcache/record_store.go
```diff
@@ -67,23 +102,21 @@ func (rs *RecordStore) Get(key interface{}) (value interface{}, found bool, err
...
+		rs.invalidate(key)
+		rs.incrementMisses()
+		return nil, false, nil
+	}
 	nowMS := time.Now().UnixMilli()
 	if rs.recordExpired(rec, nowMS) {
 		rs.Invalidate(key)
-		// onExpire
+		rs.onExpire()
 		rs.incrementExpirations()
```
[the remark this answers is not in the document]
yuce: I've fixed this in one of the commits but I can't find it. The up-to-date code is correct: https://github.com/hazelcast/hazelcast-go-client/pull/807/files#diff-d98395b374977a39a074c89372c5decc6008263673166c80286d42e9f2dd5bcaR114

config.go
```diff
@@ -269,6 +281,63 @@ func (f *FlakeIDGeneratorConfig) Validate() error {
...
+
+func (pc InvalidationConfig) Clone() InvalidationConfig {
+	return InvalidationConfig{
+		maxToleratedMissCount:         pc.maxToleratedMissCount,
+		reconciliationIntervalSeconds: pc.reconciliationIntervalSeconds,
+		err:                           pc.err,
+	}
+}
+
+func (pc *InvalidationConfig) Validate() error {
```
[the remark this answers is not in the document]
yuce: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/24624ef616401d25ca45b672f2c963f2ff050f55

yuce:
@utku-caglayan @srknzl Field aligmnent should be fixed: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/391fb150163f2c7235f6910baf221b54b85875a4

internal/nearcache/record_store.go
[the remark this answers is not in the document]
utku-caglayan: Thanks a lot for the resources! I think you are right, it gets the job done for now.
The only concern right now is "randomness" of the maps depends on the go version. I think we can either introduce randomness explicitly or just create a test that ensures iterations are somewhat random.

Even if we do not explicitly introduce randomness, I think we need a mechanism to alert us if the behaviour of the maps change, otherwise it would be quite hard to find that bug. Wdyt?

internal/nearcache/repairing_task.go
srknzl: @utku-caglayan to resolve since he asked for ticker reset, not me.

internal/nearcache/repairing_task.go
srknzl: Well, of course I have no strong opinion, but at first read I could not understand what it meant until I read Java side. Hence I have commented

internal/nearcache/repairing_task.go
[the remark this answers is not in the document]
srknzl: Ok they have the same size in server side code. https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/invalidation/AbstractBaseNearCacheInvalidationListener.java#L79 

Resolving

nearcache/nearcache_it_test.go
srknzl: You forgot to update `maxCacheSize`

config.go
yuce: We return this error when the user attempts `StartNewClientWithConfig`. If there's an error in the config, the client doesn't start and returns an error, so the user immediately knows where the problem is and how to fix that. That's in contrast with the Java client, where this exception is thrown only if the user tries to get a map with the near cache, so it may not be apparent what the fix is.

The documentation is missing for some of the configs here, and I agree that "Configuring a value of zero seconds disables the reconciliation task." is an important information. I will add that in the documentation PR.

nearcache_map.go
yuce: https://github.com/hazelcast/hazelcast-go-client/pull/807/commits/be6741a1d8290e83a043566cce08ef14ea4e0ed3

internal/nearcache/repairing_task.go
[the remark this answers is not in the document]
yuce: What I mean is, if atomics is used on a variable, IMO it is better to use it even when not strictly necessary, in order not to forget to use it where necessary.

internal/nearcache/record_store.go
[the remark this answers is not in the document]
utku-caglayan: To be addressed later

internal/nearcache/repairing_task.go
```diff
@@ -115,14 +115,58 @@ func (rt *ReparingTask) DeregisterHandler(name string) {
...
+		case <-rt.doneCh:
+			return
+		case <-timer.C:
+			rt.initRepairingHandlerAsyncRun(handler, 1)
+		}
+	}()
+}
+func (rt *ReparingTask) initRepairingHandlerAsyncRun(handler RepairingHandler, roundNumber int32) {
+	// reference implementation calls initRepairingHandler, but I think that's wrong
+	if rt.invalidationMetaDataFetcher.Init(context.Background(), handler) {
```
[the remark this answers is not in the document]
yuce: `Init` returns `true` if the initialization is successful. In that case it should reschedule this `initRepairingHandlerAsyncRun` task.

utku-caglayan (approved):
Huge work, thanks!

utku-caglayan (approved):

srknzl (approved):

PR #808, opened 2022-06-27
Fix uptime gauge [API-1382]

utku-caglayan:


internal/stats/stats_service.go
```diff
@@ -257,7 +264,11 @@ func (g runtimeGauges) updateNumCPU(bt *binTextStats) {
 }
 
 func (g runtimeGauges) updateUptime(bt *binTextStats) {
-	if uptime, err := host.Uptime(); err != nil {
+	if g.proc == nil {
+		// gopsutil.process is not registered, skip gauge
+		return
+	}
+	if uptime, err := g.proc.CreateTime(); err != nil {
```
yuce, line 271: The uptime would be calculated as: `time.Now() - g.proc.CreateTime()`.

yuce (changes requested):
Just uptime calculation is not correct.

internal/stats/stats_service.go
utku-caglayan: Thanks for catching that. Copied time.Time to ms utility from Nearcache PR. This way we won't make duplication when they are merged.

https://github.com/hazelcast/hazelcast-go-client/pull/808/commits/94f37a2947d456aedaab0b20093381cb89c23057

srknzl (dismissed):

internal/stats/stats_service.go
yuce: OK, but I've removed that stuff from NC, since we will upgrade to Go 1.17 that one is not necessary anymore.

yuce (approved):

PR #815, opened 2022-07-02
Remove MakeSubscriptionID method

yuce:
This PR removes `event.MakeSubscriptionID` method from the few places it was used. `event.MakeSubscriptionID` returned an automatic subscription ID based on the function pointer. That kind of usage is problematic with the release of Go 1.17, see: https://go.dev/doc/go1.17#compiler


yuce:
@utku-caglayan Discovered bugs in the group lost event handling in the Invocation Service. One of the test tasks events fail because of that. Could you check out https://github.com/hazelcast/hazelcast-go-client/pull/816 before this one?

srknzl (approved):

PR #816, opened 2022-07-03
Fix Invocation Service group lost handling

yuce:
This PR fixes group lost event handling in the Invocation Service:

* The subscriber to the group lost event in the invocation service should return if the invocation service is stopped.
* `handleGroupLost` relied on an atomic value for protection, which was wrong. Replaced that with a mutex.


utku-caglayan (approved):

utku-caglayan (commented):
I think this works and understand the reasoning behind adding the doneChannel control. But I do not get replacing the atomic variable with a mutex-protected one. Can you share your reasoning on that one?

nevzatseferoglu (approved):
LGTM! Just a quick remark,`s.executor.stop()` use a `waitgroup` inside it. In case of any unintended infinite spinning on wait creates a deadlock on the invocation service state mutex.

PR #817, opened 2022-07-03
Cloud APIs

alisengul53:
New auto-generetaed remote_controller.go is added.
Cloud APIs are added to util.go


internal/it/util.go
```diff
@@ -458,3 +458,37 @@ func WaitEventuallyWithTimeout(t *testing.T, wg *sync.WaitGroup, timeout time.Du
 		t.FailNow()
 	}
 }
+
+// Cloud APIs
+
+func LoginToHazelcastCloudUsingEnvironment() error {
+	return rc.LoginToHazelcastCloudUsingEnvironment(context.Background())
```
yuce, line 465: Could you add a `ctx context.Context` parameter to these APIs? Like, this one would become:
```
func LoginToHazelcastCloudUsingEnvironment(ctx context.Context) error {
	return rc.LoginToHazelcastCloudUsingEnvironment(ctx)
}
```

yuce (changes requested):
Just the context param comment...

internal/it/util.go
alisengul53: Updated

internal/it/util.go
```diff
@@ -458,3 +458,37 @@ func WaitEventuallyWithTimeout(t *testing.T, wg *sync.WaitGroup, timeout time.Du
...
+
+func LoginToHazelcastCloud(ctx context.Context, uri string, apiKey string, apiSecret string) error {
+	return rc.LoginToHazelcastCloud(ctx, uri, apiKey, apiSecret)
+}
+
+func CreateHazelcastCloudStandardCluster(ctx context.Context, hzVersion string, isTlsEnabled bool) (*CloudCluster, error) {
+	return rc.CreateHazelcastCloudStandardCluster(ctx, hzVersion, isTlsEnabled)
+}
+
+func GetHazelcastCloudCluster(ctx context.Context, clusterId string) (*CloudCluster, error) {
```
yuce, line 476: Could you rename `clusterId` to `clusterID` due to Go naming conventions. Same comment for others.

yuce (changes requested):
Sorry, one more small thing...

internal/it/util.go
alisengul53: Updated

PR #826, opened 2022-07-06
Nearcache Implementation Part 4

yuce:
TODO


nearcache_map.go
```diff
@@ -182,6 +229,44 @@ func (ncm *nearCacheMap) Get(ctx context.Context, m *Map, key interface{}) (inte
...
+	defer ncm.releaseRemainingReservedKeys(resMap)
+	pairs, err := m.getAllFromRemote(ctx, len(missKeys), partitionToKeys)
+	if err != nil {
+		return nil, fmt.Errorf("nearCacheMap.GetAll: getting keys from remote: %w", err)
+	}
+	keyCount, err := ncm.populateResultFromRemote(pairs, entries, resMap)
+	if err != nil {
+		return nil, err
+	}
+	return entries[:keyCount], nil
```
utku-caglayan, line 267: I don't understand the reason for slicing the `entries`? Can you please explain?

nearcache_map.go
```diff
@@ -319,3 +440,88 @@ func (ncm *nearCacheMap) handleBatchInvalidationMsg(rth *inearcache.RepairingHan
...
+	for _, k := range keys {
+		nk, err := ncm.toNearCacheKey(k)
+		if err != nil {
+			return nil, err
+		}
+		cached, ok, err := ncm.getCachedValue(nk, true)
+		if err != nil {
+			return nil, err
+		}
+		if ok && cached != nil {
```
utku-caglayan, line 457: Can `cached` be nil? Java docs say you cannot cache a nil value. I may be wrong

nearcache_map.go
```diff
@@ -319,3 +440,88 @@ func (ncm *nearCacheMap) handleBatchInvalidationMsg(rth *inearcache.RepairingHan
...
+	// assumes entries has max capacity.
+	serialize := ncm.serializeKeys
+	i := len(entries) - len(pairs)
+	for _, p := range pairs {
+		kd := p.Key.(serialization.Data)
+		k, err := ncm.ss.ToObject(kd)
+		if err != nil {
+			return 0, err
+		}
+		e := types.Entry{Key: k}
```
utku-caglayan, line 507: Java client assigns the keys according to the `serializeKeys` flag. This may be missing that switch

nearcache/nearcache_it_test.go
```diff
@@ -73,6 +77,80 @@ func TestSmokeNearCachePopulation(t *testing.T) {
...
+				t.Fatal(err)
+			}
+			require.Equal(t, i, v)
+		}
+		// GetAll() generates the Near Cache hits
+		vs, err := m.GetAll(ctx, keys...)
+		if err != nil {
+			t.Fatal(err)
+		}
+		sort.Slice(vs, func(i, j int) bool {
```
utku-caglayan, line 110: I think we can use `require.ElementsMatch` instead of sorting

nearcache/nearcache_it_test.go
```diff
@@ -73,6 +77,80 @@ func TestSmokeNearCachePopulation(t *testing.T) {
...
+			return k1 < k2
+		})
+		require.Equal(t, target, vs)
+		stats := m.LocalMapStats().NearCacheStats
+		require.Equal(t, int64(size), stats.OwnedEntryCount)
+		require.Equal(t, int64(size), stats.Hits)
+	})
+}
+
+func TestGetAllPopulatesNearCache(t *testing.T) {
```
utku-caglayan, line 122: As far as I can see, this test does nothing new compared to the test above. I wonder why we have this.

nearcache/nearcache_it_test.go
```diff
@@ -484,3 +712,28 @@ func assertNearCacheExpiration(tcx it.MapTestContext, size int32) {
 		return true
 	})
 }
+
+// this is the same entry processor from the hazelcast_test package.
+// TODO: move this to it
```
utku-caglayan, line 717: Is this forgotten?

nearcache/nearcache_it_test.go
```diff
@@ -275,6 +417,92 @@ func TestAfterSetNearCacheIsInvalidated(t *testing.T) {
...
+		keySet := make(map[int32]struct{})
+		for len(keySet) < int(keyCount) {
+			key := int32(rand.Intn(int(size)))
+			keySet[key] = struct{}{}
+		}
+		keys := make([]interface{}, 0, keyCount)
+		for k := range keySet {
+			keys = append(keys, k)
+		}
+
```
utku-caglayan, line 496: redundant line

utku-caglayan:
Idea: Would running the map integration tests with near cache configuration be straightforward and beneficial? Maybe we can add an extra step like we run many tests for both smart and unisocket connection

utku-caglayan (commented):
Looks great! Here few comments

nearcache_map.go
```diff
@@ -182,6 +229,44 @@ func (ncm *nearCacheMap) Get(ctx context.Context, m *Map, key interface{}) (inte
 	return value, nil
 }
 
+func (ncm *nearCacheMap) GetAll(ctx context.Context, m *Map, keys []interface{}) ([]types.Entry, error) {
+	// see: com.hazelcast.client.map.impl.nearcache.NearCachedClientMapProxy#getAllInternal
+	entries := make([]types.Entry, len(keys))
+	missKeys, err := ncm.populateResultFromNearCache(keys, entries)
```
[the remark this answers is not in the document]
yuce: `populateResultFromNearCache` serializes the keys if necessary using `toNearCacheKey`: https://github.com/yuce/hazelcast-go-client/blob/3662f4a646ff96c754b2b608939ccaef4f2a5f1c/nearcache_map.go#L449

nearcache_map.go
```diff
@@ -182,6 +229,44 @@ func (ncm *nearCacheMap) Get(ctx context.Context, m *Map, key interface{}) (inte
...
+	partitionToKeys, err := m.partitionToKeys(keys, false)
+	if err != nil {
+		return nil, err
+	}
+	defer ncm.releaseRemainingReservedKeys(resMap)
+	pairs, err := m.getAllFromRemote(ctx, len(missKeys), partitionToKeys)
+	if err != nil {
+		return nil, fmt.Errorf("nearCacheMap.GetAll: getting keys from remote: %w", err)
+	}
+	keyCount, err := ncm.populateResultFromRemote(pairs, entries, resMap)
```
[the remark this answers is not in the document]
yuce: `entries` is returned to the user and it should always contain deserialized keys. I had to make `GetAll` code a bit different from the reference implementation. But the reference implementation seems to do something similar. Unless I am missing something.

nearcache_map.go
yuce: Yes, `cached` can be `nil`, but the user cannot cache a nil value. The code corresponds to this reference implementation code at com/hazelcast/client/map/impl/nearcache/NearCachedClientMapProxy.java:431
```
if (cached != null && cached != NOT_CACHED) {
```

nearcache/nearcache_it_test.go
yuce: Probably they were added at different times to the reference implementation. Added it for test parity.

nearcache/nearcache_it_test.go
yuce: https://github.com/hazelcast/hazelcast-go-client/pull/826/commits/288838714331c22b574ef533bb32054d9c682ddd

nearcache/nearcache_it_test.go
yuce: Makes sense: https://github.com/hazelcast/hazelcast-go-client/pull/826/commits/288838714331c22b574ef533bb32054d9c682ddd

yuce:
> Idea: Would running the map integration tests with near cache configuration be straightforward and beneficial? Maybe we can add an extra step like we run many tests for both smart and unisocket connection.
> I am asking this because it would be great to leverage all the tests we have, and also test the near cache with more happy paths

We can consider that, but all of the near cache tests check for the returned value, so not sure about the benefit of that.

nearcache_map.go
yuce: That's returned to the user, and should have the deserialized key.

utku-caglayan (approved):
lgtm

alisengul53 (approved):

PR #828, opened 2022-07-07
[master] Fixes map PutIfAbsentXXX methods

yuce:
Map's `PutIfAbsentXXX` methods erroneously returned `serialization.Data` instead of actual values. This PR fixes that and updates the tests to catch that problem.


utku-caglayan (approved):
Great catch, thanks!

alisengul53 (approved):

PR #838, opened 2022-07-09
Fixed races in the codebase

yuce:
Fixed the following races in the codebase, which caused tests fail when the race detector is enabled.

* ConnectionManager reset
* BaseCredentials endpoint
* `TestReplicatedMap_AddEntryListenerToKeyWithPredicate`

Also changed the target for the race enabled tests to `test-race`.

Added a workflow for running the tests with race detector.


utku-caglayan (approved):
Looks good

alisengul53 (approved):

PR #845, opened 2022-07-14
Near Cache docs and code samples [API-1418]

yuce:


config.go
```diff
@@ -185,6 +185,7 @@ func (c *Config) Validate() error {
 	return nil
 }
 
+// MarshalJSON marshals the configuratio to JSON.
```
utku-caglayan, line 188:
```suggestion
// MarshalJSON marshals the configuration to JSON.
```

examples/map/nearcache/json_config/config.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+const (
+	defaultKeyCount       = 1_000
+	defaultRepeat         = 10_000
+	defaultGoroutineCount = 1
+)
+
+type Config struct {
+	MapName        string
```
utku-caglayan, line 35: alignment err above this line

examples/map/nearcache/json_config/config.go
```diff
@@ -0,0 +1,66 @@
...
+	GoroutineCount int
+	Warmup         bool
+	Client         hazelcast.Config
+}
+
+func (c *Config) LoadFromPath(path string) error {
+	if _, err := os.Stat(path); err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("path not found: %s: %w", path, err)
+		}
```
utku-caglayan, line 47: I think we need to handle the errors that are not `os.ErrNotExist` as well

utku-caglayan: Also how about skipping the `os.Stat` part and directly trying to read from the path?

examples/map/nearcache/json_config/main.go
```diff
@@ -0,0 +1,93 @@
...
+	var config Config
+	if err := config.LoadFromPath(os.Args[1]); err != nil {
+		bye(err.Error())
+	}
+	ctx := context.Background()
+	cl, err := hazelcast.StartNewClientWithConfig(ctx, config.Client)
+	if err != nil {
+		panic(err)
+	}
+	m, err := cl.GetMap(context.Background(), config.MapName)
```
utku-caglayan, line 60:
```suggestion
	m, err := cl.GetMap(ctx, config.MapName)
```

examples/map/nearcache/json_config/runners.go
```diff
@@ -0,0 +1,72 @@
...
+	for i := 0; i < gc; i += 1 {
+		go func(start, count int64, wg *sync.WaitGroup) {
+			for v := start; v < start+count; v++ {
+				if _, err := m.Get(context.Background(), v); err != nil {
+					panic(err)
+				}
+			}
+			wg.Done()
+			atomic.AddInt64(&opCnt, count)
+		}(int64(i*keyPerGoroutine), int64(keyPerGoroutine), wg)
```
utku-caglayan, line 68: I think we can also pass `opCnt` as a parameter to be similar with other params

examples/map/nearcache/programmatic_config/main.go
```diff
@@ -0,0 +1,80 @@
...
+		value := fmt.Sprintf("value-%d", i)
+		if err := m.Set(ctx, i, value); err != nil {
+			panic(err)
+		}
+	}
+	printStats(m)
+	// start hitting the near cache
+	tic := time.Now()
+	for i := 1; i <= repeat; i++ {
+		if i%1000 == 0 {
```
utku-caglayan, line 67: This prints only once. Is it intended?

proxy_map.go
```diff
@@ -98,6 +98,83 @@ Once the lock context is created, it can be used to lock/unlock entries and used
...
+
+Map entries in Hazelcast are partitioned across the cluster members.
+Hazelcast clients do not have local data at all.
+Suppose you read the key k a number of times from a Hazelcast client or k is owned by another member in your cluster.
+Then each map.Get(k) will be a remote operation, which creates a lot of network trips.
+If you have a data structure that is mostly read, then you should consider creating a local Near Cache, so that reads are sped up and less network traffic is created.
+
+These benefits do not come for free. See the following trade-offs:
+
+    * Members with a Near Cache has to hold the extra cached data, which increases memory consumption.
```
utku-caglayan, line 111:
```suggestion
    * Members and clients with a Near Cache has to hold the extra cached data, which increases memory consumption.
```

proxy_map.go
```diff
@@ -98,6 +98,83 @@ Once the lock context is created, it can be used to lock/unlock entries and used
...
+Note that Near Cache configuration is specific to the server or client itself.
+A data structure on a server may not have Near Cache configured while the same data structure on a client may have Near Cache configured.
+They also can have different Near Cache configurations.
+
+If you are using the Near Cache, you should take into account that your hits to the keys in the Near Cache are not reflected as hits to the original keys on the primary members.
+This has for example an impact on Map's maximum idle seconds or time-to-live seconds expiration.
+Therefore, even though there is a hit on a key in Near Cache, your original key on the primary member may expire.
+
+Note: Near Cache works only when you access data via map.Get(k).
+Data returned using a predicate is not stored in the Near Cache.
```
utku-caglayan, line 127:
```suggestion
Data returned using a predicate or SQL is not stored in the Near Cache.
```
I think we should add SQL too

utku-caglayan (commented):
Thanks, left a few suggestions!

examples/map/nearcache/programmatic_config/main.go
```diff
@@ -0,0 +1,80 @@
...
+	const mapName = "mymap-go"
+	const keyCount = 1_000
+	repeat := 1_000
+	ctx := context.Background()
+	// configuration
+	config := hazelcast.Config{}
+	config.Stats.Enabled = true
+	ec := nearcache.EvictionConfig{}
+	ec.SetPolicy(nearcache.EvictionPolicyLFU)
+	ec.SetSize(keyCount)
```
utku-caglayan, line 43: Getters are not present for these fields. We can try to include them here or another example to cover the whole API

utku-caglayan:
What would you think about adding an example that registers a custom `EvictionPolicyComparator`? ATM we do not cover related API

examples/map/nearcache/json_config/config.go
yuce: I think checking field alignment for examples is not necessary, will ignore them in `check.sh`

examples/map/nearcache/json_config/config.go
yuce: > I think we need to handle the errors that are not os.ErrNotExist as well
If the file exist, we attempt to read it with `b, err := os.ReadFile(path)`, in case there are other problems (like no permission, ...) then we would get an error there too.

> Also how about skipping the os.Stat part and directly trying to read from the path?
I think it's nice to know when the file does not exist.

examples/map/nearcache/json_config/runners.go
yuce: Do you mean passing `opCnt` to the goroutine? `opCnt` is supposed to be calculated in the goroutine.

examples/map/nearcache/programmatic_config/main.go
yuce: The number of prints depends on the `repeat` constant. The user can set it to a multiple of 1000 to see more output.

proxy_map.go
yuce: I've directly copied it from the reference implementation, but I agree using clients instead of members makes more sense for us.

examples/map/nearcache/programmatic_config/main.go
yuce: Although getters have to be exported in this case, they are supposed to be "internal". It's not very beneficial to be used by the user.

examples/map/nearcache/json_config/main.go
yuce: https://github.com/hazelcast/hazelcast-go-client/pull/845/commits/262c5cf8f465d87b6c0fa06c1fb59ed656c7745e

proxy_map.go
yuce: https://github.com/hazelcast/hazelcast-go-client/pull/845/commits/262c5cf8f465d87b6c0fa06c1fb59ed656c7745e

proxy_map.go
yuce: https://github.com/hazelcast/hazelcast-go-client/pull/845/commits/262c5cf8f465d87b6c0fa06c1fb59ed656c7745e

yuce:
> What would you think about adding an example that registers a custom EvictionPolicyComparator? ATM we do not cover related API

That's a good idea.

yuce:
> What would you think about adding an example that registers a custom EvictionPolicyComparator? ATM we do not cover related API

Added at: https://github.com/hazelcast/hazelcast-go-client/pull/845/commits/6a47bcc05cce290a963b46f35cde2d5c39907c02

examples/map/nearcache/json_config/runners.go
utku-caglayan: We can pass `&opCnt` as a runner function parameter instead of using it from the outer scope. I'm perfectly fine with using everything from the outer scope, it's just other params are handled that way.

utku-caglayan (approved):

proxy_map.go
```diff
@@ -98,6 +98,89 @@ Once the lock context is created, it can be used to lock/unlock entries and used
 As mentioned before, lock context is a regular context.Context which carry a special lock ID.
 You can pass any context.Context to any Map function, but in that case lock ownership between operations using the same hazelcast.Client instance is not possible.
 
+Using the Near Cache
+
+Map entries in Hazelcast are partitioned across the cluster members.
+Hazelcast clients do not have local data at all.
+Suppose you read the key k a number of times from a Hazelcast client or k is owned by another member in your cluster.
+Then each map.Get(k) will be a remote operation, which creates a lot of network trips.
+If you have a data structure that is mostly read, then you should consider creating a local Near Cache, so that reads are sped up and less network traffic is created.
```
emreyigit, line 107: Typo
```suggestion
If you have a data structure that is mostly read, then you should consider creating a local Near Cache, so that reads are speed up and less network traffic is created.
```

yuce: I took most of the text straight from the Hazelcast docs. I think `sped up` is correct: https://forum.wordreference.com/threads/speed-sped-speeded.858459/

emreyigit: Didn't know that, thanks.

PR #847, opened 2022-07-17
Unskip skipped near cache tests

yuce:
This PR:
* Makes `TestSmokeNearCachePopulation` as close as possible to the reference implementation test and unskips it. Previously it started the client before manipulating the map, which resulted in test failure due to ocassionally receiving invalidation messages after map manipulation.
* Makes `TestMemberLoadAllInvalidatesClientNearCache` use a single member cluster, similar to the reference implemenation and unskipts it.
* Skips `TestAfterPutNearCacheIsInvalidated` for Hz 4.2 series. This is required for compatibility suite tests.

### DIFF
--- a/nearcache/nearcache_it_test.go
+++ b/nearcache/nearcache_it_test.go
@@ -48,43 +48,48 @@ const (
 
 func TestSmokeNearCachePopulation(t *testing.T) {
 	// ported from: com.hazelcast.client.map.impl.nearcache.ClientMapNearCacheTest#smoke_near_cache_population
+	const memberCount = 3
 	tcx := it.MapTestContext{
 		T: t,
-		ConfigCallback: func(tcx it.MapTestContext) {
-			ncc := nearcache.Config{Name: tcx.MapName}
-			ncc.SetInvalidateOnChange(true)
-			tcx.Config.AddNearCache(ncc)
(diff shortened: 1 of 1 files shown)


nearcache/nearcache_it_test.go
```diff
@@ -592,6 +585,7 @@ func TestAfterDeleteNearCacheIsInvalidated(t *testing.T) {
 
 func TestAfterPutNearCacheIsInvalidated(t *testing.T) {
 	// port of: com.hazelcast.client.map.impl.nearcache.ClientMapNearCacheTest#testAfterPutAsyncNearCacheIsInvalidated
+	it.SkipIf(t, "hz > 4.1.9, hz < 4.3")
```
utku-caglayan, line 588: I think skip.If PR will be merged soon, let's leverage that when it does here

utku-caglayan (commented):
Looks great, just one comment

emreyigit (approved):

nearcache/nearcache_it_test.go
yuce: Let's merge this one first, if it's OK for you. Since this one is crucial.

utku-caglayan (approved):

PR #851, opened 2022-07-22
Remove unncessary MarkRacy annotations

yuce:
Currently the codebase doesn't contain any known race condition. Some of the tests were marked as `MarkRacy`, which is wrong.

Also bumped the client version.


utku-caglayan (approved):
Thanks

mehmettokgoz (approved):
LGTM.

PR #855, opened 2022-08-01
Prevent enterprise test to run on forked repo [API-1462]

nevzatseferoglu:
Ensure nightly jobs are only running on original repository.
Remove unnecessary obsolete explanation and fix typos.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -5,10 +5,10 @@ Number of the clients:
 Cluster size, i.e. the number of Hazelcast cluster members:
 OS version : Windows/Linux/OSX
 
-#### Expected behaviour
+#### Expected behavior
 
 
-#### Actual behaviour
+#### Actual behavior
 
 
-#### Steps to reproduce the behaviour
+#### Steps to reproduce the behavior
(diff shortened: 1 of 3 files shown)


utku-caglayan (approved):
Looks good!

.github/ISSUE_TEMPLATE.md
```diff
@@ -5,10 +5,10 @@ Number of the clients:
...
+#### Expected behavior
 
 
-#### Actual behaviour
+#### Actual behavior
 
 
-#### Steps to reproduce the behaviour
\ No newline at end of file
+#### Steps to reproduce the behavior
```
yuce: `behaviour` is not wrong, it's written like that in British English.

nevzatseferoglu: My bad, but Google itself uses as `behavior`,
https://support.google.com/analytics/answer/2785577?hl=en#zippy=%2Cin-this-article
I think we should choose the same phrase with them as being a Go client.
Also, overall usage of `behavior` is bigger.
Screen Shot 2022-08-01 at 10 07 02
Wdyt ?

yuce: Google is a US company, so that makes sense for them :) I don't agree with your reasoning, but `behavior` is OK.

yuce (approved):

PR #858, opened 2022-08-04
Fix nightly bug

nevzatseferoglu:
Add Go 1.19 to nightly tests.
There was a bug due to location of shutdown handlers execution, the design has changed.
`TestProxyManagerShutdown` is using SSL enabled member side XML config, client config has changed according to it.

### DIFF
--- a/client_internals_it_test.go
+++ b/client_internals_it_test.go
@@ -387,9 +387,9 @@ func TestClientInternal_EncodeData(t *testing.T) {
 func TestProxyManagerShutdown(t *testing.T) {
 	clientTester(t, func(t *testing.T, smart bool) {
 		ctx := context.Background()
-		tc := it.StartNewClusterWithOptions("proxy-manager-graceful-shutdown", 5701, 1)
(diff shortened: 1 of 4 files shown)


client_internals_it_test.go
```diff
@@ -387,9 +387,9 @@ func TestClientInternal_EncodeData(t *testing.T) {
 func TestProxyManagerShutdown(t *testing.T) {
 	clientTester(t, func(t *testing.T, smart bool) {
 		ctx := context.Background()
-		tc := it.StartNewClusterWithOptions("proxy-manager-graceful-shutdown", 5701, 1)
+		tc := it.StartNewClusterWithOptions("proxy-manager-graceful-shutdown", 35701, 1)
```
yuce, line 390: Could you use `t.Name()` as the cluster name instead of `proxy-manager-graceful-shutdown` ?

yuce (changes requested):
Looks good, except one small change.

yuce (approved):

PR #859, opened 2022-08-08
Integration with OpenSSL, introduces sample ssl integration tests

nevzatseferoglu:
The PR scope introduces OpenSSL self-signed certificate approach for testing both unit tests and integration tests of current SSL support.
By doing that,
- Client will not need to rely on member-side cryptographic certificates or artifacts anymore and creates its own for both side.
- Existing certificates, truststores and keystores for both client and server will need to be renewed annually, but it is alterable.
- Client can test different cipher suits with underlying implementation by becoming isolated from the server-side. It will make unknown bugs more visible.

Currently,
- Mutual Authentication test fails because we don't have a support for decrypting `pkcs12` signed certificates. It looks like bug 🐛 .
- Since we create our own certificate authority for ourselves, underlying host machine must trust that CA. Different OS does do that in a different ways. Right now,  the PR status is draft, after solving the bug, we can add a instruction section for different operating system.


client_internals_it_test.go
```diff
@@ -542,3 +543,176 @@ func ensureClient(config hz.Config) *hz.Client {
 	}
 	panic("the client could not connect to the cluster in 60 seconds.")
 }
+
+type MutualAuthenticationStatus string
+
+const (
+	OPTIONAL MutualAuthenticationStatus = "OPTIONAL"
+	REQUIRED MutualAuthenticationStatus = "REQUIRED"
```
yuce, line 551: Could you rename these constants to:
```
MutualAuthenticationStatusOptional
MutualAuthenticationStatusRequired
```

client_internals_it_test.go
```diff
@@ -542,3 +543,176 @@ func ensureClient(config hz.Config) *hz.Client {
...
+)
+
+func TestMutualAuthentication(t *testing.T) {
```
yuce, line 554: Could you rename the test to `mutualAuthenticationTest` and add it to `TestClientInternal`?

client_internals_it_test.go
```diff
@@ -542,3 +543,176 @@ func ensureClient(config hz.Config) *hz.Client {
...
+	testCases := []struct {
+		name          string
+		muStatus      MutualAuthenticationStatus
+		ca, cert, key string
+		wantErr       bool
+	}{
+		{
+			name:     "testMA_RequiredClientAndServerAuthenticated",
```
yuce, line 562: No need to use the `testMA_` prefix.

client_internals_it_test.go
```diff
@@ -542,3 +543,176 @@ func ensureClient(config hz.Config) *hz.Client {
...
+			muStatus: OPTIONAL,
+			ca:       "testdata/OpenSSL/rootCA.crt",
+			cert:     "",
+			key:      "",
+			wantErr:  false,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			skip.IfNot(t, "oss")
```
yuce, line 644: Why do are the tests skipped for enterprise? SSL is supposed to work only for enterprise.

client_internals_it_test.go
yuce: Btw, why is this test in `client_internals_it_test.go`? As far as I see it doesn't require anything from `ClientInternals`.

client_internals_it_test.go
```diff
@@ -542,3 +543,176 @@ func ensureClient(config hz.Config) *hz.Client {
...
+			if tc.wantErr {
+				require.Error(t, sslTestUtil(t, tc.muStatus, tc.ca, tc.cert, tc.key))
+			} else {
+				require.NoError(t, sslTestUtil(t, tc.muStatus, tc.ca, tc.cert, tc.key))
+			}
+		})
+	}
+}
+
+func sslTestUtil(t *testing.T, mu MutualAuthenticationStatus, ca, cert, key string) error {
```
yuce, line 654: How about `status` instead of `mu`?

yuce: I think no need to pass `t` since it is being used only for `t,Name()`. You can pass the cluster name instead.

client_internals_it_test.go
```diff
@@ -542,3 +543,176 @@ func ensureClient(config hz.Config) *hz.Client {
...
+           http://www.hazelcast.com/schema/config/hazelcast-config-4.0.xsd">
+			<cluster-name>%s</cluster-name>
+			<network>
+				<port>%d</port>
+				<ssl enabled="true">
+					<factory-class-name>
+						com.hazelcast.nio.ssl.ClasspathSSLContextFactory
+					</factory-class-name>
+					<properties>
+						<property name="keyStore">testdata/OpenSSL/ca.keystore</property>
```
yuce, line 694: This assumes RC is being run in the project root:
```
<property name="keyStore">testdata/OpenSSL/ca.keystore</property>
```
I think that should be more flexible, and use an absolute path instead of the relative path. The absolute path can be found out using something like:
```
wd, err := os.Getwd()
path := filepath.Join(wd, "testdata/OpenSSL/ca.keystore")
```

client_internals_it_test.go
```diff
@@ -542,3 +543,176 @@ func ensureClient(config hz.Config) *hz.Client {
...
+				<map-store enabled="true">
+					<class-name>com.hazelcast.client.test.SampleMapStore</class-name>
+				</map-store>
+			</map>
+			<serialization>
+				<data-serializable-factories>
+					<data-serializable-factory factory-id="66">com.hazelcast.client.test.IdentifiedFactory</data-serializable-factory>
+					<data-serializable-factory factory-id="666">com.hazelcast.client.test.IdentifiedDataSerializableFactory</data-serializable-factory>
+				</data-serializable-factories>
+			</serialization>
```
yuce, line 715: These shouldb't be required for the SSL tests.

cluster/ssl_config_test.go
```diff
@@ -58,11 +58,22 @@ func TestSSLConfig_AddClientCertAndKeyPath(t *testing.T) {
...
-	err = sslConfig.AddClientCertAndKeyPath("testdata/client1-cert.pem", "testdata/client1-key.pem")
+	err = sslConfig.AddClientCertAndKeyPath("../testdata/OpenSSL/client/client.crt", "../testdata/OpenSSL/client/client.key")
 	require.NoError(t, err)
 	require.Equal(t, len(sslConfig.TLSConfig().Certificates), 1)
-	// another valid client certificate, key pair
-	err = sslConfig.AddClientCertAndKeyPath("testdata/client2-cert.pem", "testdata/client2-key.pem")
+}
+
+func TestSSLConfig_AddClientCertAndEncryptedKeyPath(t *testing.T) {
+	it.MarkFlaky(t, "PKCS12 fails while decrypting")
```
yuce, line 67: This is not a flaky test. Could you instead use something like:
```
t.Skipf("PKCS12 encrypted certificates are not yet supported")
```

testdata/OpenSSL/build.sh
```diff
@@ -0,0 +1,104 @@
...
+CN = test.hazelcast.com
+
+[ req_ext ]
+subjectAltName = @alt_names
+
+[ alt_names ]
+DNS.1 = test.hazelcast.com
+DNS.2 = www.test.hazelcast.com
+IP.1 = 192.168.1.5
+IP.2 = 192.168.1.6
```
yuce, line 37: How about saving this to `csr.conf` instead and load it?

testdata/OpenSSL/build.sh
```diff
@@ -0,0 +1,104 @@
...
+
+authorityKeyIdentifier=keyid,issuer
+basicConstraints=CA:FALSE
+keyUsage = digitalSignature, nonRepudiation, keyEncipherment, dataEncipherment
+subjectAltName = @alt_names
+
+[alt_names]
+DNS.1 = test.hazelcast.com
+
+
```
yuce, line 51: Same comment for `csr.conf`.

yuce (changes requested):
I'll continue to review

client_internals_it_test.go
yuce: Could you use the same name as the Java counterpart, like: `whenRequired_andServerNotAuthenticated` and add a comment about where is the test coming from?

testdata/OpenSSL/build.sh
```diff
@@ -0,0 +1,104 @@
...
+EOF
+
+mkdir server
```
yuce, line 54: Could you change this to `mkdir -p server` so it can be run without removing the `server` directory. The same for the `client` directory.

nevzatseferoglu: Fixed, [c6481ec](https://github.com/hazelcast/hazelcast-go-client/pull/859/commits/c6481ec04328cf044ab3bec8e31bd712bd663bd2).

testdata/OpenSSL/build.sh
nevzatseferoglu: The command already exports the configuration as file to the current directory, I think there is no need for applying extra file operation, if you are okay with that.

testdata/OpenSSL/build.sh
nevzatseferoglu: The command already exports the configuration as file to the current directory, I think there is no need for applying extra file operation, if you are okay with that.

cluster/ssl_config_test.go
nevzatseferoglu: Changed, [c6481ec](https://github.com/hazelcast/hazelcast-go-client/pull/859/commits/c6481ec04328cf044ab3bec8e31bd712bd663bd2).

client_internals_it_test.go
nevzatseferoglu: Removed, [c6481ec](https://github.com/hazelcast/hazelcast-go-client/pull/859/commits/c6481ec04328cf044ab3bec8e31bd712bd663bd2).

client_internals_it_test.go
nevzatseferoglu: Absolute path could not be resolved for the certificate artifacts, the only working way is putting resources to the root directory, then using relative path. I could not figure out the reason.

client_internals_it_test.go
nevzatseferoglu: Fixed, [c6481ec](https://github.com/hazelcast/hazelcast-go-client/pull/859/commits/c6481ec04328cf044ab3bec8e31bd712bd663bd2).

client_internals_it_test.go
nevzatseferoglu: It was intentional but fixed [c6481ec](https://github.com/hazelcast/hazelcast-go-client/pull/859/commits/c6481ec04328cf044ab3bec8e31bd712bd663bd2).

client_internals_it_test.go
nevzatseferoglu: Added, [c6481ec](https://github.com/hazelcast/hazelcast-go-client/pull/859/commits/c6481ec04328cf044ab3bec8e31bd712bd663bd2).

client_internals_it_test.go
nevzatseferoglu: Moved to another source, [c6481ec](https://github.com/hazelcast/hazelcast-go-client/pull/859/commits/c6481ec04328cf044ab3bec8e31bd712bd663bd2).

client_internals_it_test.go
nevzatseferoglu: Renamed, [c6481ec](https://github.com/hazelcast/hazelcast-go-client/pull/859/commits/c6481ec04328cf044ab3bec8e31bd712bd663bd2).

ssl_it_test.go
```diff
@@ -0,0 +1,246 @@
...
+		//{
+		//	// ported from, com.hazelcast.client.nio.ssl.ClientAuthenticationTest#whenOptional_andWrongClientAuthenticated
+		//	name:     "whenOptional_andWrongClientAuthenticated",
+		//	muStatus: MutualAuthenticationStatusOptional,
+		//	ca:       "testdata/openssl/server/server.crt",
+		//	cert:     "testdata/unknown-client-cert.pem",
+		//	key:      "testdata/unknown-client-key.pem",
+		//	wantErr:  true,
+		//},
+		{
```
yuce, line 123: Why was this test commented out?

nevzatseferoglu: As I mentioned in PR comment;

`The test whenOptional_andWrongClientAuthenticated supposed to fail but it passes. For now, I commented down that test case in codebase, waiting for the to be reviewed.`

yuce: Could you uncomment it?

ssl_it_test.go
```diff
@@ -0,0 +1,246 @@
...
+			key:      "",
+			wantErr:  false,
+		},
+	}
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			if tc.wantErr {
+				require.Error(t, sslTestUtil(t.Name()+"_cluster", tc.muStatus, tc.ca, tc.cert, tc.key))
```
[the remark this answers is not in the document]
nevzatseferoglu: It is not. Removed. [e48b360](https://github.com/hazelcast/hazelcast-go-client/pull/859/commits/e48b360e7d97f6fcca8d8d8e204a4524c842e17b)

ssl_it_test.go
nevzatseferoglu: Uncommented, [e48b360](https://github.com/hazelcast/hazelcast-go-client/pull/859/commits/e48b360e7d97f6fcca8d8d8e204a4524c842e17b).

testdata/OpenSSL/build.sh
yuce: What I meant was moving that content to the separate `csr.conf` file and load it from the script. Is there an advantage of keeping it in the script?

PR #862, opened 2022-08-10
Add cluster shutdown test case for SQL

utku-caglayan:
Tests whether SQL driver respects context cancelation on cluster reconnect

### DIFF
--- a/sql/driver/driver_it_test.go
+++ b/sql/driver/driver_it_test.go
@@ -31,6 +31,7 @@ import (
 	hz "github.com/hazelcast/hazelcast-go-client"
 	"github.com/hazelcast/hazelcast-go-client/cluster"
 	"github.com/hazelcast/hazelcast-go-client/internal/it"
+	"github.com/hazelcast/hazelcast-go-client/internal/it/skip"
 	"github.com/hazelcast/hazelcast-go-client/logger"
 	"github.com/hazelcast/hazelcast-go-client/serialization"
 	"github.com/hazelcast/hazelcast-go-client/types"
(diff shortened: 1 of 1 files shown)


sql/driver/driver_it_test.go
```diff
@@ -527,6 +528,46 @@ func TestContextCancelAfterFirstPage(t *testing.T) {
...
+			go func() {
+				// check if it respects context cancel
+				<-time.After(2 * time.Second)
+				cancel()
+			}()
+		}
+		t.Log(v)
+	}
+	err = rows.Err()
+	require.Contains(t, err.Error(), "context canceled")
```
yuce, line 568: Would it be better to check using `errors.Is`?:
```
if !errors.Is(err, context.Canceled) {
   t.Fatal("expected context canceled error")
}

**Review by yuce — COMMENTED** [review]
just an idea...

File: sql/driver/driver_it_test.go (hunk 1 above)
Comment (utku-caglayan) [reply sql/driver/driver_it_test.go]: https://github.com/hazelcast/hazelcast-go-client/pull/862/commits/46d23408a379bab46104a824d03c0e1f8ab793e9

**Review by yuce — APPROVED** [review]

File: sql/driver/driver_it_test.go [hunk 2]
```diff
@@ -527,6 +528,48 @@ func TestContextCancelAfterFirstPage(t *testing.T) {
 	})
 }
 
+func TestClusterShutdownDuringQuery(t *testing.T) {
+	it.SkipIf(t, "hz < 5.0")
+	ctx := context.Background()
+	port := it.NextPort()
+	cls := it.StartNewClusterWithConfig(it.MemberCount(), it.SQLXMLConfig(t.Name(), "localhost", port), port)
+	defer cls.Shutdown()
+	config := cls.DefaultConfig()
```
yuce, line 537: I think this should be `cls.DefaultConfigWithNoSSL()`. Could you try with `ENABLE_SSL=1`?

yuce (changes requested):

sql/driver/driver_it_test.go
utku-caglayan: You are right, thanks https://github.com/hazelcast/hazelcast-go-client/pull/862/commits/680f341c9f032567ffe5fa090f962b9ee2ec064b

yuce (approved):

PR #864, opened 2022-08-11
Stringer interface for EntryNotified type.

mehmettokgoz:
PR closed by Hazelcast automation as no activity (>3 months). Please reopen with comments, if necessary. Thank you for using Hazelcast and your valuable contributions

### DIFF
--- a/events.go
+++ b/events.go
@@ -17,6 +17,7 @@
 package hazelcast
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/hazelcast/hazelcast-go-client/cluster"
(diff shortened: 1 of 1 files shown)


srknzl (approved):

events.go
```diff
@@ -79,6 +107,11 @@ func (e *EntryNotified) EventName() string {
 	return eventEntryNotified
 }
 
+func (e *EntryNotified) String() string {
```
utku-caglayan, line 110: Can you add a test for this, please?

utku-caglayan:
Also let's cc @yuce since this changes the public API

events.go
mehmettokgoz: Where can I write test about that? There is `events_tests.go` file but I couldn't be sure.

utku-caglayan: `events_tests.go` makes sense 👍

PR #873, opened 2022-08-22
Map.RemoveAll doesn't clear the near cache [API-1519]

mehmettokgoz:
Fix: https://github.com/hazelcast/hazelcast-go-client/issues/872

Near Cache clear() operation was missing when the removeAll() method is called on a map. 

**Changes**

- Added check for clearing Near Cache
- New test for map.removeAll() operation

### DIFF
--- a/nearcache/nearcache_it_test.go
+++ b/nearcache/nearcache_it_test.go
@@ -325,9 +325,9 @@ func TestMapRemoveAll_WithNearCache(t *testing.T) {
 		const size = int32(100)
 		populateMap(tcx, size)
 		populateNearCache(tcx, size)
-		err := m.RemoveAll(context.Background(), predicate.True())
+		err := m.RemoveAll(context.Background(), predicate.Less("__key", int32(50)))
 		require.NoError(t, err)
-		require.Equal(t, int64(0), m.LocalMapStats().NearCacheStats.OwnedEntryCount)
+		require.Equal(t, int64(50), m.LocalMapStats().NearCacheStats.OwnedEntryCount)
 	})
 }
 
--- a/nearcache/nearcache_it_test.go
+++ b/nearcache/nearcache_it_test.go
@@ -22,6 +22,7 @@ package nearcache_test
 import (
 	"context"
 	"fmt"
+	"github.com/hazelcast/hazelcast-go-client/predicate"
 	"math/rand"
 	"os"
 	"strconv"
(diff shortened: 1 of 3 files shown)


mehmettokgoz:
There is an issue in `return m.ncm.Clear()` line, do not review it until fix.

nearcache/nearcache_it_test.go
```diff
@@ -316,6 +317,20 @@ func TestMapRemove_WithNearCache(t *testing.T) {
...
+	tcx := newNearCacheMapTestContext(t, nearcache.InMemoryFormatBinary, true)
+	tcx.Tester(func(tcx it.MapTestContext) {
+		t := tcx.T
+		m := tcx.M
+		const size = int32(100)
+		populateMap(tcx, size)
+		populateNearCache(tcx, size)
+		err := m.RemoveAll(context.Background(), predicate.Less("__key", int32(50)))
+		require.NoError(t, err)
+		require.Equal(t, int64(50), m.LocalMapStats().NearCacheStats.OwnedEntryCount)
```
yuce, line 330: Owned entry count should be 0 after the near cache is deleted. I think near cache was not setup for this map.

nearcache_map.go
```diff
@@ -360,6 +361,20 @@ func (ncm *nearCacheMap) Remove(ctx context.Context, m *Map, key interface{}) (i
...
+	keys, err := m.GetKeySetWithPredicate(ctx, predicate)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		for _, key := range keys {
+			key, err = ncm.toNearCacheKey(key)
+			ncm.nc.Invalidate(key)
+		}
+	}()
```
yuce, line 374: This code doesn't look correct, all needed should be something like this:
```
defer func() {
    ncm.Clear(ctx, m)
}()
```

proxy_map.go
```diff
@@ -748,6 +748,16 @@ func (m *Map) removeFromRemote(ctx context.Context, key interface{}) (interface{
 	return m.convertToObject(codec.DecodeMapRemoveResponse(response))
 }
 
+func (m *Map) removeAllFromRemote(ctx context.Context, predicate predicate.Predicate) error {
+	if predicateData, err := m.validateAndSerialize(predicate); err != nil {
+		return err
+	} else {
```
yuce, line 754: Could you reorganize this code so `else` is not needed:
```
predicateData, err := m.validateAndSerialize(predicate)
if err != nil {
    return err
}
```
```

yuce (changes requested):

nearcache_map.go
mehmettokgoz: Fixed at https://github.com/hazelcast/hazelcast-go-client/pull/873/commits/59d4abf8629e69964b3905af346d81a7dbfb6fd3.

nearcache/nearcache_it_test.go
mehmettokgoz: Fixed at https://github.com/hazelcast/hazelcast-go-client/pull/873/commits/59d4abf8629e69964b3905af346d81a7dbfb6fd3.

yuce (approved):
Looks good!

PR #885, opened 2022-09-23
[DRAFT] do not try to connect to all the seed addresses

utku-caglayan:


client_it_test.go
```diff
@@ -1043,6 +1044,62 @@ func clientStartShutdownWithNilContextTest(t *testing.T) {
 	it.Must(client.Shutdown(nil))
 }
 
+func clusterConnectionToMultipleAddrs(t *testing.T) {
```
yuce, line 1047: Could you add a `Test` suffix to this function's name?

client_it_test.go
```diff
@@ -1043,6 +1044,62 @@ func clientStartShutdownWithNilContextTest(t *testing.T) {
...
+			addresses: []string{
+				fmt.Sprintf("localhost:%d", tc.Port),
+				fmt.Sprintf("localhost:%d", mem.Port),
+			},
+		},
+		{
+			name: "first addr valid, second is invalid",
+			addresses: []string{
+				fmt.Sprintf("localhost:%d", tc.Port),
+				"non-existent.com:1234",
```
yuce, line 1074: It seems `non-existent.com` exists, could you use a domain guaranteed to be not used, maybe: `non-existent.non-existent`.

internal/cluster/cluster_service.go
```diff
@@ -154,28 +154,31 @@ func (s *Service) sendMemberListViewRequest(ctx context.Context, conn *Connectio
...
 
 type AddrSet struct {
-	addrs map[string]pubcluster.Address
+	addrs       map[string]struct{}
+	orderedAddr []pubcluster.Address
 }
 
-func NewAddrSet() AddrSet {
-	return AddrSet{addrs: map[string]pubcluster.Address{}}
+func NewAddrSet() *AddrSet {
```
yuce, line 161: Is there a reason to return a pointer?

internal/cluster/cluster_service.go
```diff
@@ -154,28 +154,31 @@ func (s *Service) sendMemberListViewRequest(ctx context.Context, conn *Connectio
...
 }
 
-func (a AddrSet) Addrs() []pubcluster.Address {
-	addrs := make([]pubcluster.Address, 0, len(a.addrs))
-	for _, addr := range a.addrs {
-		addrs = append(addrs, addr)
-	}
+func (a *AddrSet) Addrs() []pubcluster.Address {
+	addrs := make([]pubcluster.Address, len(a.orderedAddr))
+	copy(addrs, a.orderedAddr)
```
yuce, line 181: I think it's OK to return `orderedAddr` as is.

yuce (changes requested):
Thanks, here are a few ideas...

internal/cluster/cluster_service.go
utku-caglayan: `AddrSet` contains a slice. It is updated with `AddAddr` method. If we do not use a pointer receiver, updates to the slice are not reflected to the instance.

internal/cluster/cluster_service.go
utku-caglayan: Sounds good https://github.com/hazelcast/hazelcast-go-client/pull/885/commits/b6c392dce32ad6f9628881f5492531354940969d

client_it_test.go
utku-caglayan: Makes sense https://github.com/hazelcast/hazelcast-go-client/pull/885/commits/b6c392dce32ad6f9628881f5492531354940969d

client_it_test.go
utku-caglayan: Sure thanks https://github.com/hazelcast/hazelcast-go-client/pull/885/commits/b6c392dce32ad6f9628881f5492531354940969d

internal/cluster/cluster_service.go
yuce: We should use a pointer receiver with `AddAddr` but we can still return non-pointer `AddrSet` from `NewAddrSet`, they are not related.

utku-caglayan: I find it intuitive for constructors to return pointers to the original struct if that struct implements its method with a pointer receiver. It is harder to misuse on caller's side. Is there a reason not to return a pointer?

yuce: In this case, the code is not in the hot path, and Go can probably optimize the `AddrSet` instance to be allocated on the stack but it's generally preferred to create non-pointer values if possible in order to reduce GC pressure. I am not sure how the user can misuse `AddrSet` when it's not a pointer, but converting it a pointer makes it easier to make mistakes:
```
func NewAddrSet() AddrSet {
    // ...
}
a1 := NewAddrSet()
for i := 0; i < 10; i++ {
	a1.AddAddr(fmt.Sprintf("%d.com", i))
}
b1 := a1
for i := 10; i < 100; i++ {
	b1.AddAddr(fmt.Sprintf("%d.com", i))
}
fmt.Println(len(a1.Addrs()))
```
Prints 10 as expected, but:
```
func NewAddrSet() *AddrSet {
    // ...
}
a1 := NewAddrSet()
for i := 0; i < 10; i++ {
	a1.AddAddr(fmt.Sprintf("%d.com", i))
}
b1 := a1
for i := 11; i < 100; i++ {
	b1.AddAddr(fmt.Sprintf("%d.com", i))
}
fmt.Println(len(a1.Addrs()))
```
prints 100

utku-caglayan: I think implicitly, value receiver conveys the meaning "it is safe to copy me". For a type such AddrSet, I expect it to be statefult, meaning that it should recall the things I inserted even if I copy it.
I'm not so sure about re-assignment. I don't think it is a common usecase. On the other hand, passing custom types to  functions as parameters is. In that case, pointer receiver automatically behaves as expected.
I agree that performance shouldn't be part of the discussion for this case.
Please correct me if I'm wrong but I think there is no strong argument for it to be value receiver.

utku-caglayan: After some discussion, we decided to refactor AddrSet type into a function to avoid any misuse in the future https://github.com/hazelcast/hazelcast-go-client/pull/885/commits/d951ef7360b3eff55bf4fcb3c3e0cb7b87942f93

yuce (approved):
Thanks, looks good!

yuce (approved):

PR #896, opened 2022-10-03
Fixed flakeIDGeneratorServiceNameTest

yuce:
`GetDistributedObjectsInfo` is sent to a random member, but there may be a window where the created FlakeIDGenerator was not synced to all members.

This PR makes sure that the test doesn't give up until the FlakeIDGenrator is synced to all members.

The following command always reproduces the test failure:
```
MEMBER_COUNT=15 go test -count 1000 -timeout 60m -run TestFlakeIDGenerator
```


utku-caglayan (approved):
Great catch, thanks!

PR #905, opened 2022-12-02
Remove discovery token from debug log [API-1739]

mehmettokgoz:
Previously, the client was printing the `url.Error()` error directly, so the token was also leaking to logs. Instead, it should print an `InvalidConfigurationError`.

Switch statement is added since we might get an invalid discovery token error here, and its print log is different. 

Old error:
```
2022/12/02 11:28:03 DEBUG: cluster.ConnectionManager: error connecting to cluster, attempt 1: failed to refresh seed addresses: Get "https://ap.viridian.hazelcast.com/cluster/discovery?token=<TOKEN_LEAK>": dial tcp: lookup ap.viridian.hazelcast.com: no such host
```

New error:
```
2022/12/02 12:13:20 DEBUG: cluster.ConnectionManager: error connecting to cluster, attempt 1: failed to refresh seed addresses: failed to reach Hazelcast Cloud/Viridian service: https://foo.com: invalid configuration error
```

### DIFF
--- a/internal/cloud/cloud_discovery.go
+++ b/internal/cloud/cloud_discovery.go
@@ -19,12 +19,13 @@ package cloud
 import (
 	"context"
 	"fmt"
-	"github.com/hazelcast/hazelcast-go-client/cluster"
-	"github.com/hazelcast/hazelcast-go-client/internal/logger"
-	"github.com/hazelcast/hazelcast-go-client/internal/rest"
 	"os"
 	"strconv"
 	"strings"
+
+	"github.com/hazelcast/hazelcast-go-client/cluster"
+	"github.com/hazelcast/hazelcast-go-client/internal/logger"
+	"github.com/hazelcast/hazelcast-go-client/internal/rest"
 )
 
 const envCoordinatorBaseURL = "HZ_CLOUD_COORDINATOR_BASE_URL"
--- a/internal/cloud/cloud_discovery.go
+++ b/internal/cloud/cloud_discovery.go
@@ -19,6 +19,7 @@ package cloud
 import (
 	"context"
 	"fmt"
+	"github.com/hazelcast/hazelcast-go-client/internal/hzerrors"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


internal/cloud/cloud_discovery.go
```diff
@@ -19,6 +19,8 @@ package cloud
 import (
 	"context"
 	"fmt"
+	"github.com/hazelcast/hazelcast-go-client/internal/hzerrors"
+	url2 "net/url"
```
yuce, line 23: GoLand automatically creates import aliases like these, let's avoid using them.

internal/cloud/cloud_discovery.go
```diff
@@ -47,7 +49,12 @@ func NewDiscoveryClient(config *cluster.CloudConfig, logger logger.LogAdaptor) *
 func (c *DiscoveryClient) DiscoverNodes(ctx context.Context) ([]Address, error) {
 	url := makeCoordinatorURL(c.token)
 	if j, err := c.httpClient.GetJSONArray(ctx, url); err != nil {
-		return nil, err
+		switch err.(type) {
+		case *url2.Error:
+			return nil, hzerrors.NewInvalidConfigurationError(fmt.Sprintf("failed to reach Hazelcast Cloud/Viridian service: %s", baseURL()), nil)
+		default:
+			return nil, err
+		}
```
yuce, line 57: Instead of checking the error here, how about doing that in `HTTPClient.Get` around line 68:
```
if resp, err := c.httpClient.Do(req); err != nil {
	var e *url.Error
	if errors.As(err, &e) {
		e.URL = ""
	}
	return nil, err
// ...
```
In order to use `url.Error`, you will have to change the `url` parameter to something like `uri`.

yuce (changes requested):

internal/cloud/cloud_discovery.go
mehmettokgoz: IMO, removing the URL from error is not convenient since HTTPClient is more general purpose. But I searched and couldn't find any other usage of it so moved the check to there. See https://github.com/hazelcast/hazelcast-go-client/pull/905/commits/89df180ed0646b11b9cd31e1e675ffa78d8c5794.

yuce (approved):

PR #910, opened 2022-12-28
Compact Schema Distribution [API-1746] [API-1885]

yuce:
* Added the schema service.
* Added schema fetch on read.
* Added sending all registered schemas to the cluster on connection.
* Moved invoker logic to the internal client.
* Updated Client Internal to use the invoker.
* Refactored compact tester.
* More compact tests.


internal/proto/codec/builtin.go
```diff
@@ -123,6 +123,8 @@ func (codecUtil) NextFrameIsDataStructureEndFrame(frameIterator *proto.ForwardFr
 	return frameIterator.PeekNext().IsEndFrame()
 }
 
+// NextFrameIsNullFrame
+// Deprecated:
 func (codecUtil) NextFrameIsNullFrame(frameIterator *proto.ForwardFrameIterator) bool {
```
mdumandag, line 128: Is the first line in the comment the name of the alternative or this function? I mean, we say that it is deprecated and I would like to make sure that we are linking the alternative.

internal/proto/codec/client_fetch_schema_codec.go
```diff
@@ -0,0 +1,53 @@
+/*
```
mdumandag, line 1: Can you link the protocol PR associated with these codecs?

internal/serialization/compact_serializer.go
```diff
@@ -53,11 +56,12 @@ func (CompactStreamSerializer) ID() int32 {
 }
 
 func (c CompactStreamSerializer) Read(input pubserialization.DataInput) interface{} {
-	schema := c.getOrReadSchema(input)
-	typeName := schema.TypeName()
-	serializer, ok := c.typeNameToSerializer[typeName]
+	// TODO: move context to the method signature
+	ctx := context.Background()
+	schema := c.getOrReadSchema(ctx, input)
```
mdumandag, line 61: can schema be null?

internal/serialization/compact_test.go
```diff
@@ -0,0 +1,130 @@
...
+		map1 := it.MustValue(tcx.Client1.GetMap(ctx, tcx.MapName)).(*hz.Map)
+		it.Must(map1.Set(ctx, 1, v))
+		map2 := it.MustValue(tcx.Client2.GetMap(ctx, tcx.MapName)).(*hz.Map)
+		it.AssertEquals(t, it.MustValue(map2.Get(ctx, 1)), v)
+	})
+}
+
+func basicQueryTest(t *testing.T) {
+	// ported from: com.hazelcast.internal.serialization.impl.compact.integration.CompactFormatIntegrationTest#testBasicQuery
+	skip.If(t, "hz < 5.2")
```
mdumandag, line 72: don't we already skip it at the top?

internal/serialization/compact_test.go
```diff
@@ -0,0 +1,130 @@
...
+		for i := int32(0); i < 100; i++ {
+			it.MustValue(map1.Put(ctx, i, EmployeeDTO{age: i, id: 102310312}))
+		}
+		map2 := it.MustValue(tcx.Client2.GetMap(ctx, tcx.MapName)).(*hz.Map)
+		ks := it.MustValue(map2.GetKeySetWithPredicate(ctx, predicate.SQL("age > 19"))).([]interface{})
+		assert.Equal(t, 80, len(ks))
+	})
+}
+
+func joinedMemberQueryTest(t *testing.T) {
```
mdumandag, line 89: This is not really a client test. we are expecting the joined member to receive the schemas from the master in this test

internal/serialization/schema_service.go
```diff
@@ -19,21 +16,67 @@
...
+	s.mu.RLock()
+	schema, ok = s.schemaMap[schemaId]
+	s.mu.RUnlock()
+	if !ok {
+		rch := make(chan *Schema)
+		s.ch <- SchemaMsg{ID: schemaId, ResponseCh: rch}
+		select {
+		case schema = <-rch:
+			ok = schema != nil
+			s.putLocal(schema)
```
mdumandag, line 66: if schema is null, wouldn't this call throw?

serialization/api.go
```diff
@@ -414,53 +414,53 @@ type PortableReader interface {
...
+	FieldKindNullableInt16          FieldKind = 37
+	FieldKindArrayOfNullableInt16   FieldKind = 38
+	FieldKindNullableInt32          FieldKind = 39
+	FieldKindArrayOfNullableInt32   FieldKind = 40
+	FieldKindNullableInt64          FieldKind = 41
+	FieldKindArrayOfNullableInt64   FieldKind = 42
+	FieldKindNullableFloat32        FieldKind = 43
+	FieldKindArrayOfNullableFloat32 FieldKind = 44
+	FieldKindNullableFloat64        FieldKind = 45
+	FieldKindArrayOfNullableFloat64 FieldKind = 46
```
mdumandag, line 463: why these look like 
```
-
   -
 -
```
?

Can you fix the formatting?

internal/cluster/connection_manager.go
```diff
@@ -192,6 +192,9 @@ func (m *ConnectionManager) start(ctx context.Context) error {
 	} else if addr, err = m.startUnisocket(ctx); err != nil {
 		return err
 	}
+	if err = m.sendStateToCluster(ctx); err != nil {
```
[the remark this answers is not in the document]
yuce: Yes, `ConnectionManager.start` is executed whenever the client reconnects.

internal/serialization/compact_serializer.go
yuce: No, `getOrReadSchema` panics if ithe schema is not found.

internal/serialization/compact_test.go
yuce: It also checks if the schema was sent from the client, so I think it's useful.

serialization/api.go
yuce: That's automatically formatted by gofmt, we have no control over it.

cluster/failover_config.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
+ * Copyright (c) 2008-2022, Hazelcast, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
yuce: The years are not very important. Specifying a previous year doesn't remove the copyright.

internal/proto/codec/builtin.go
yuce: The doc comment must start with the name of the identifier. I've improved the deprecation notices at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/05d7f6629a2d0b144925608c3f3e9e07b5f9a570

internal/serialization/compact_test.go
yuce: Removed at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/05d7f6629a2d0b144925608c3f3e9e07b5f9a570

internal/serialization/schema_service.go
yuce: It would, fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/05d7f6629a2d0b144925608c3f3e9e07b5f9a570

internal/proto/codec/client_fetch_schema_codec.go
yuce: There's no protocol PR, since https://github.com/hazelcast/hazelcast-client-protocol/pull/420 is WIP.

cluster/failover_config.go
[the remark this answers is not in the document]
mdumandag: I don't know much about the legal aspect of these copyright years but just wanted to remind you, as in the Hazelcast repo, the copyright years are updated just after the new year so I thought it was needed.

internal/cluster/connection_manager.go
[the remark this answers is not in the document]
mdumandag: Also, where do we have the logic of not allowing invocations until the state is sent?

internal/serialization/compact_test.go
mdumandag: You test that in the `basicQueryTest` as well, this doesn't add anything

internal/serialization/compact_test.go
```diff
@@ -0,0 +1,129 @@
...
+		v := EmployeeDTO{age: 30, id: 102310312}
+		k := int64(1)
+		it.MustValue(tcx.M.Put(ctx, k, v))
+		// changeCluster
+		tcx.Cluster.Shutdown()
+		it.MustValue(tcx.Cluster.StartMember(ctx))
+		it.MustValue(tcx.M.Put(ctx, k, v))
+		v2 := it.MustValue(tcx.M.Get(ctx, k))
+		assert.Equal(t, v, v2)
+	})
```
mdumandag, line 128: Since you cache the schemas generated from the serializers, I believe we won't fetch the schemas from the cluster in any of these tests. Can you confirm this? If this is the case, you might add a test where there are two versions of the schema for the same class and make sure that you can read data written by the other client

internal/serialization/field_description.go
```diff
@@ -16,20 +16,20 @@
 
 package serialization
 
-import pserialization "github.com/hazelcast/hazelcast-go-client/serialization"
+import pubserialization "github.com/hazelcast/hazelcast-go-client/serialization"
 
 type FieldDescriptor struct {
-	fieldName string
-	fieldKind pserialization.FieldKind
+	FieldName string
```
mdumandag, line 22: maybe we can rename this to `name` in a similar fashion to `fieldKind` -> `kind` change

internal/serialization/schema_service.go
```diff
@@ -19,21 +16,69 @@
...
+			}
+		}
+	}
 	return schema, ok
 }
 
-func (s *SchemaService) PutLocal(schema Schema) {
-	s.schemaMap[schema.ID()] = schema
+func (s *SchemaService) putLocal(schema *Schema) {
+	s.mu.Lock()
```
mdumandag, line 79: In Java I believe there is a check to make sure that if `ok` is True, the schema you have in your local is the same as `schema` as a sanity check. maybe we can add it here as well

mdumandag (commented):
@yuce is there a way to see the code coverage for this change?

internal/proto/codec/client_fetch_schema_codec.go
yuce: They are auto-generated, but some manual changes are necessary, like the imports.

client.go
```diff
@@ -118,9 +121,46 @@ func newClient(config Config) (*Client, error) {
...
+				req := codec.EncodeClientFetchSchemaRequest(msg.ID)
+				resp, err := c.ic.Invoker.InvokeOnRandomTarget(ctx, req, nil)
+				if err != nil {
+					c.ic.Logger.Errorf("invoking ClientFetchSchema with schema ID: %d: %s", msg.ID, err)
+					msg.ResponseCh <- nil
+					continue
+				}
+				schema := codec.DecodeClientFetchSchemaResponse(resp)
+				msg.ResponseCh <- schema
+			}
```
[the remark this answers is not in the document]
yuce: The channel here is never closed, so `ok` cannot be false. Removed the `if` check altogether at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/e05971dea6e4a15abc845f2fba679be0cd760144

internal/proto/codec/client_fetch_schema_codec.go
mdumandag: OK, then since this is not something specific to Compact, I am resolving this

yuce:
> What happens when this fails? In Java, we retry sending the state again. Also, in Java, there are other parts of the state that might apply to the Go client, like local proxies. I believe we should also send them (maybe can be done in a separate PR, but I believe we should do it)

If sending state to the cluster fails during `StartNewClient`, ConnectionManager start fails too, and then `StartNewClient` fails. This may happen if the client disconnects from the cluster right before sending the state.

I've updated the code to retry at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/45ba429f57eb87c4c2a80dd10d6aa393ece3a2a2

I found a bug during testing this, and created this ticket: https://github.com/hazelcast/hazelcast-go-client/issues/934

internal/serialization/schema_service.go
yuce: In the TDD it states that https://hazelcast.atlassian.net/wiki/spaces/IMDG/pages/3976593439/Compact+Schema+Distribution+TDD#Assumptions%2FLimitations
> In case of a collision, we will simply reject the second schema with the same schema id but different content.

I think Go implementation follows the TDD closer, it rejects the new schema in case there's a collision but Java implementation raises an exception.

internal/cluster/connection_manager.go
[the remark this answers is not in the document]
yuce: > What happens when this fails? In Java, we retry sending the state again. Also, in Java, there are other parts of the state that might apply to the Go client, like local proxies. I believe we should also send them (maybe can be done in a separate PR, but I believe we should do it)

If sending state to the cluster fails during `StartNewClient`, ConnectionManager start fails too, and then `StartNewClient` fails. This may happen if the client disconnects from the cluster right before sending the state.

I've updated the code to retry at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/45ba429f57eb87c4c2a80dd10d6aa393ece3a2a2

I found a bug during testing this, and created this ticket: https://github.com/hazelcast/hazelcast-go-client/issues/934

internal/it/compact.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+func MakeCompactClusterConfig(clusterName, mapName, inMemoryFormat string, port int) string {
+	return fmt.Sprintf(`
+        <hazelcast xmlns="http://www.hazelcast.com/schema/config"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="http://www.hazelcast.com/schema/config
+            http://www.hazelcast.com/schema/config/hazelcast-config-4.0.xsd">
+            <cluster-name>%s</cluster-name>
+            <map name="%s">
```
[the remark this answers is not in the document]
yuce: Reformatted at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/bee476ee105340c4f7c18018cdc7c0159b991f80

internal/serialization/compact_test.go
yuce: It seems I've missed something while refactoring the test helper. Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/a3d1fad7e21a66cc158a7bc9aba4b6257dcfc9af

internal/serialization/field_description.go
yuce: That makes sense. Renamed at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/8506b81cd1fc06f50be586facc47cc0e57fd396b

internal/cluster/connection_manager.go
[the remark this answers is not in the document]
yuce: > where do we have the logic of not allowing invocations until the state is sent?

I've added some logic to pause sending non-urgent invocations at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/7471a22a203060a34d4436fa8a174cd8ade08e13

https://github.com/yuce/hazelcast-go-client/blob/ad4231ffbac8ec70c57a8c235e732b44b415cb0c/internal/invocation/invocation_service.go#L121

internal/invocation/invocation_service.go
```diff
@@ -78,6 +83,20 @@ func NewService(handler Handler, ed *event.DispatchService, lg logger.LogAdaptor
 			}
 		}()
 	})
+	s.eventDispatcher.Subscribe(EventInvocationStateChanged, stateSubID, func(event event.Event) {
+		e := event.(*InvocationStateChanged)
+		if !e.Enabled {
+			atomic.CompareAndSwapInt32(&s.paused, 0, 1)
```
mdumandag, line 89: Also, how can we be sure that no invocation is taking place, while we are executing this event callback?

internal/it/compact.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+		cfg1 := tcx.Cluster.DefaultConfig()
+		cfg1.Serialization.Compact.SetSerializers(tcx.Serializers1...)
+		tcx.T.Logf("map name: %s", tcx.MapName)
+		tcx.T.Logf("cluster address: %s", cfg1.Cluster.Network.Addresses[0])
+		if tcx.Client1 == nil {
+			tcx.Client1 = getDefaultClient(&cfg1)
+		}
+		cfg2 := tcx.Cluster.DefaultConfig()
+		cfg2.Serialization.Compact.SetSerializers(tcx.Serializers1...)
```
mdumandag, line 78: should this be serializers2?

internal/proto/codec/builtin.go
```diff
@@ -1119,3 +1135,61 @@ func decodeLocalTime(buffer []byte, offset int32) (h, m, s, nanos int) {
 	nanos = int(FixSizedTypesCodec.DecodeInt(buffer, offset))
 	return h, m, s, nanos
 }
+
+func EncodeInt(buffer []byte, offset, value int32) {
+	binary.LittleEndian.PutUint32(buffer[offset:], uint32(value))
```
mdumandag, line 1140: TIL go does not have signed versions of these methods 😄

internal/serialization/compact_serializer.go
```diff
@@ -88,11 +86,27 @@ func (c CompactStreamSerializer) IsRegisteredAsCompact(t reflect.Type) bool {
...
+	for _, ser := range cfg.Serializers() {
+		s := makeSchemaFromSerializer(ser)
+		r[s.ID()] = s
+	}
+	return r
+}
+
+func makeSchemaFromSerializer(ser pubserialization.CompactSerializer) *Schema {
+	sw := NewSchemaWriter(ser.TypeName())
+	// create the zero value for the type in the serializer and
```
mdumandag, line 109: minor: you might forget to finish this sentence

internal/invocation/invocation_service.go
```diff
@@ -78,6 +83,20 @@ func NewService(handler Handler, ed *event.DispatchService, lg logger.LogAdaptor
 			}
 		}()
 	})
+	s.eventDispatcher.Subscribe(EventInvocationStateChanged, stateSubID, func(event event.Event) {
+		e := event.(*InvocationStateChanged)
```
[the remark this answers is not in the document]
yuce: Those events won't get lost. The "normal" invocations are not accepted to the invocation channel if the invocation service is paused: https://github.com/yuce/hazelcast-go-client/blob/ad4231ffbac8ec70c57a8c235e732b44b415cb0c/internal/invocation/invocation_service.go#L120

internal/invocation/invocation_service.go
yuce: The "normal" invocations are not accepted to the invocation channel if the invocation service is paused: https://github.com/yuce/hazelcast-go-client/blob/ad4231ffbac8ec70c57a8c235e732b44b415cb0c/internal/invocation/invocation_service.go#L120

internal/it/compact.go
yuce: Right, fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/e6f38dc4f62fadf3654d60034195dc914ac20dc4

internal/proto/codec/builtin.go
yuce: :)

internal/serialization/compact_serializer.go
yuce: Completed at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/024d0660e11937c621ffe9c764c9fd6e002d9231

internal/cluster/connection_listener_binder.go
```diff
@@ -225,7 +218,14 @@ func (b *ConnectionListenerBinder) sendRemoveListenerRequest(ctx context.Context
 		return fmt.Sprintf("%d: removing listener", conn.connectionID)
 	})
 	inv := b.invocationFactory.NewConnectionBoundInvocation(request, conn, nil, start)
-	if err := b.invocationService.SendRequest(ctx, inv); err != nil {
+	req := inv.Request()
+	_, err := b.connectionManager.invoker.CB().TryContext(ctx, func(ctx context.Context, attempt int) (interface{}, error) {
```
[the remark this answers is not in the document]
yuce: Listener registration / deregistration is currently not urgent in the Go client because of that.

internal/invocation/invocation_service.go
```diff
@@ -99,9 +118,16 @@ func (s *Service) SetHandler(handler Handler) {
 }
 
 func (s *Service) SendRequest(ctx context.Context, inv Invocation) error {
+	if atomic.LoadInt32(&s.paused) == 1 {
+		err := fmt.Errorf("non-urgent invocations are paused: %w", hzerrors.ErrRetryableIO)
```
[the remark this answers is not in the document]
yuce: No, there's not.

client.go
```diff
@@ -118,9 +121,44 @@ func newClient(config Config) (*Client, error) {
...
+	c.ic.Logger.Debug(func() string {
+		return "Started the schema invoker"
+	})
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	sid := event.NextSubscriptionID()
+	c.ic.EventDispatcher.Subscribe(eventLifecycleEventStateChanged, sid, func(ev event.Event) {
+		e := ev.(*lifecycle.StateChangedEvent)
+		if e.State == lifecycle.StateShuttingDown {
+			cancel()
```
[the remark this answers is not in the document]
yuce: There's a small chance that the goroutine leaks if it is suspended before subscribing to the lifecycle event. I've added a ticker that checks whether the client is alive at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/5111285a1b84ae8dd77782fac8a5330ce2cdecc7

internal/cluster/connection_listener_binder.go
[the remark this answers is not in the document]
mdumandag: We made the listener invocations urgent in Java because they are connection-bound requests. And, we would throw IOError in case of trying to send a non-urgent invocation while the client is not fully connected. We do not retry invocations that receive IOException, if they are connection bound in Java. Is this not the case in Go? What do we do differently here that allows us to get away with non-urgent listener registrations?

internal/invocation/invocation_service.go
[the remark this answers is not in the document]
mdumandag: OK, not losing them is good. But, the second part of my question remains open. In the time that passed between we call the publish for this event and it starts executing, we allow invocations, right? If so, I believe that is a problem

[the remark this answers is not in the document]
yuce: Invocation service starts paused, and only enabled after the client connects to the client, so during that time there's no chance of sending non-urgent invocations.

Invocation service is paused when the client is disconnected from the cluster (all connections are removed). If an invocation is accepted after all connections are removed and before `InvocationStateChanged` event is received, the invocation wouldn't be send to any connection, since there wouldn't be any live connections to the cluster at that stage.

internal/cluster/connection_listener_binder.go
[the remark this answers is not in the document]
yuce: `b.connectionManager.invoker.CB().TryContext` retries sending the invocation, if the invocation is rejected due to invocation service is paused.

client.go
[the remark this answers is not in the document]
yuce: I've refactored the exit code at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/fe69a65a62527954bf89010036a3627de27f1d23

internal/invocation/invocation_service.go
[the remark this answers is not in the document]
yuce: Event order is guaranteed: https://github.com/hazelcast/hazelcast-go-client/blob/180f5ebe9c878e0ec1aed02a2edf63339e929668/internal/event/dispatch_service_test.go#L99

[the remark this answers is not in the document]
mdumandag: OK, that's good. But, we still have a problem with the unknown scheduling of the event callbacks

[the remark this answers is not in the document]
yuce: You are right. Refactored that at: https://github.com/hazelcast/hazelcast-go-client/pull/910/commits/154aa2354f6296bc05f16d6c7df8fa1d4c725002

internal/cluster/connection_listener_binder.go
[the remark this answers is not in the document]
mdumandag: I see that, but normally, IOException should be terminal for the connection bound invocations. Is this not the case in the Go client?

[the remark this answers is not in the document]
yuce: `SendRequest` just puts the invocation to a channel to be processed. If the invocation service is paused it directly returns an error. That's the only difference with `SendUrgentRequest`, which wouldn't return an error in this case.  `b.connectionManager.invoker.CB().TryContext` retries only the `SendRequest` bit.

IOException is supposed to be not retried in the Go client too.

mdumandag (approved):

Issue #907, opened 2022-12-09
UUID serializer read and write methods conflicts with each other [API-1784]


mehmettokgoz:
In the UUID serializer, the serializer first reads the most significant bit and then least where the write method writes the least significant bit first. 

See: https://github.com/hazelcast/hazelcast-go-client/blob/master/internal/serialization/default_serializer.go#L384

PR #911, opened 2022-12-29
UUID serializer read and write methods use different byte order [API-1784]

mehmettokgoz:
The write and read orders for the UUID fields in default serializer were wrong. Write was writing the LSB first while `Read()` reads MSB first. 

Added a test for default UUID serialization.


yuce (approved):
Looks good!

mehmettokgoz:
Closes https://github.com/hazelcast/hazelcast-go-client/issues/907.

PR #912, opened 2022-12-29
Partition ID and serializer ID byte orders depend on the configuration [API-1782]

mehmettokgoz:
If users configures byte order as LE, the partition and serializers ids are written as LE too in `service.ToData()` method. These IDs needs to encoded in BE regardless of the serialization configuration. 

Added a new `WriteInt32_()` function that takes the byte order as parameter and used it while writing IDs. 
Added a test for LE enabled configuration.

### DIFF
--- a/internal/serialization/object_data.go
+++ b/internal/serialization/object_data.go
@@ -134,6 +134,12 @@ func (o *ObjectDataOutput) WriteInt32(v int32) {
 	o.position += Int32SizeInBytes
 }
 
+func (o *ObjectDataOutput) WriteInt32_(v int32, bo binary.ByteOrder) {
+	o.EnsureAvailable(Int32SizeInBytes)
+	WriteInt32(o.buffer, o.position, v, bo)
+	o.position += Int32SizeInBytes
+}
+
 func (o *ObjectDataOutput) WriteInt64(v int64) {
 	o.EnsureAvailable(Int64SizeInBytes)
 	WriteInt64(o.buffer, o.position, v, o.bo)
--- a/internal/serialization/object_data.go
+++ b/internal/serialization/object_data.go
@@ -134,9 +134,9 @@ func (o *ObjectDataOutput) WriteInt32(v int32) {
 	o.position += Int32SizeInBytes
 }
 
-func (o *ObjectDataOutput) WriteInt32_(v int32, bo binary.ByteOrder) {
+func (o *ObjectDataOutput) WriteInt32BigEndian(v int32) {
 	o.EnsureAvailable(Int32SizeInBytes)
-	WriteInt32(o.buffer, o.position, v, bo)
+	WriteInt32(o.buffer, o.position, v, binary.BigEndian)
 	o.position += Int32SizeInBytes
 }
 
--- a/internal/serialization/serialization_test.go
+++ b/internal/serialization/serialization_test.go
@@ -19,6 +19,7 @@ package serialization_test
 import (
 	"bytes"
 	"encoding/gob"
+	"github.com/hazelcast/hazelcast-go-client/internal/it"
 	"reflect"
 	"testing"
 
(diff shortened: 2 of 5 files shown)


internal/serialization/serialization_test.go
```diff
@@ -68,6 +68,28 @@ func TestSerializationService_ToData(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	temp, err := service.ToObject(data)
+	if err != nil {
+		t.Fatal(err)
+	}
+	ret := temp.(int32)
+	if expected != ret {
+		t.Error("ToData() returns ", ret, " expected ", expected)
+	}
```
yuce, line 90: How about
```
require.Equal(expected, obj)
```

internal/serialization/serialization_test.go
```diff
@@ -68,6 +68,28 @@ func TestSerializationService_ToData(t *testing.T) {
...
+	c := &serialization.Config{}
+	c.LittleEndian = true
+	service, err := iserialization.NewService(c)
+	if err != nil {
+		t.Fatal(err)
+	}
+	data, err := service.ToData(expected)
+	if err != nil {
+		t.Fatal(err)
+	}
```
yuce, line 82: How about:
```
data := it.MustValue(service.ToData(expected))
```

internal/serialization/serialization_test.go
```diff
@@ -68,6 +68,28 @@ func TestSerializationService_ToData(t *testing.T) {
...
+	temp, err := service.ToObject(data)
+	if err != nil {
+		t.Fatal(err)
+	}
```
yuce, line 86: How about:
```
obj := it.MustValue(service.ToObject(data))
```

internal/serialization/object_data.go
```diff
@@ -134,6 +134,12 @@ func (o *ObjectDataOutput) WriteInt32(v int32) {
 	o.position += Int32SizeInBytes
 }
 
+func (o *ObjectDataOutput) WriteInt32_(v int32, bo binary.ByteOrder) {
```
yuce, line 137: How about renaming this method to `WriteInt32BigEndian` and removing the `bo` parameter?

yuce (changes requested):
Looks good, a few ideas...

internal/serialization/object_data.go
mehmettokgoz: Makes more sense since we already have LE one, done at https://github.com/hazelcast/hazelcast-go-client/pull/912/commits/b9c72d900a0da806a1b7b419c59514656f2c5b84.

internal/serialization/serialization_test.go
mehmettokgoz: Applied this comment to all tests at https://github.com/hazelcast/hazelcast-go-client/pull/912/commits/b9c72d900a0da806a1b7b419c59514656f2c5b84.

internal/serialization/serialization_test.go
mehmettokgoz: Applied this comment to all serialization tests at https://github.com/hazelcast/hazelcast-go-client/pull/912/commits/b9c72d900a0da806a1b7b419c59514656f2c5b84.

internal/serialization/serialization_test.go
mehmettokgoz: Fixed at https://github.com/hazelcast/hazelcast-go-client/pull/912/commits/b9c72d900a0da806a1b7b419c59514656f2c5b84.

mehmettokgoz:
I have removed some duplicate codes at serialization tests, I think now they are more readable. See https://github.com/hazelcast/hazelcast-go-client/pull/912/commits/b9c72d900a0da806a1b7b419c59514656f2c5b84 @yuce.

yuce (approved):
Looks good!

PR #918, opened 2023-01-18
Synchronized access to class definition

### LINKED ISSUE #916: Race Condition in ClassDefinitionContext [API-1830]
The access to ClassDefinitionContext.classDefs is not synchronized, so calling its Register method concurrently causes a data race.

ClassDefinitionContext.Register may be called concurrently if the serialization service cannot find the class definition and read it from the object.

Here's a test that ouputs race warnings when run with the race detector:

```go
func TestPortableContext_RegisterClassDefinition_Race(t *testing.T) {
	pc := serialization.NewPortableContext(nil, 0)
	const c = 1000
	wg := &sync.WaitGroup{}
	wg.Add(c)
	for i := 0; i < c; i++ {
		go func() {
			cd := pubserialization.NewClassDefinition(1, 1, 0)
			it.Must(pc.RegisterClassDefinition(cd))
			wg.Done()
		}()
	}
	wg.Wait()
}
```

yuce:
Fixes #916

### DIFF
--- a/internal/serialization/class_definition_context.go
+++ b/internal/serialization/class_definition_context.go
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
+ * Copyright (c) 2008-2022, Hazelcast, Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License")
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"sync"
 
 	ihzerrors "github.com/hazelcast/hazelcast-go-client/internal/hzerrors"
 
@@ -28,6 +29,7 @@ import (
 
 type ClassDefinitionContext struct {
 	classDefs map[string]*serialization.ClassDefinition
+	mu        *sync.RWMutex
 	factoryID int32
 }
 
@@ -35,14 +37,18 @@ func NewClassDefinitionContext(factoryID int32) *ClassDefinitionContext {
 	return &ClassDefinitionContext{
 		factoryID: factoryID,
 		classDefs: make(map[string]*serialization.ClassDefinition),
+		mu:        &sync.RWMutex{},
 	}
 }
 
 func (c *ClassDefinitionContext) LookUp(classID int32, version int32) *serialization.ClassDefinition {
-	return c.classDefs[encodeVersionedClassID(classID, version)]
+	c.mu.RLock()
+	cd := c.classDefs[encodeVersionedClassID(classID, version)]
+	c.mu.RUnlock()
+	return cd
 }
 
-func (c *ClassDefinitionContext) Register(classDefinition *serialization.ClassDefinition) error {
+func (c *ClassDefinitionContext) register(classDefinition *serialization.ClassDefinition) error {
 	if classDefinition == nil {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


internal/serialization/portable_context.go
```diff
@@ -110,19 +114,30 @@ func (c *PortableContext) LookUpOrRegisterClassDefiniton(portable serialization.
...
 
-func (c *PortableContext) RegisterClassDefinition(classDefinition *serialization.ClassDefinition) error {
-	factoryID := classDefinition.FactoryID
-	if c.classDefContext[factoryID] == nil {
-		c.classDefContext[factoryID] = NewClassDefinitionContext(factoryID)
+func (c *PortableContext) RegisterClassDefinition(cd *serialization.ClassDefinition) error {
+	c.mu.RLock()
+	factoryID := cd.FactoryID
+	f := c.classDefContext[factoryID]
+	c.mu.RUnlock()
```
mehmettokgoz, line 130: What about `defer c.mu.RUnlock()` after line 127 instead of locking again if `f==nil`?

internal/serialization/class_definition_context.go
```diff
@@ -28,32 +29,39 @@ import (
...
 		classDefs: make(map[string]*serialization.ClassDefinition),
+		mu:        &sync.RWMutex{},
 	}
 }
 
 func (c *ClassDefinitionContext) LookUp(classID int32, version int32) *serialization.ClassDefinition {
-	return c.classDefs[encodeVersionedClassID(classID, version)]
+	c.mu.RLock()
+	cd := c.classDefs[encodeVersionedClassID(classID, version)]
+	c.mu.RUnlock()
```
mehmettokgoz, line 47: Is there any difference between calling `c.mu.RUnlock()` directly or in `defer` statement in this function?

internal/serialization/class_definition_context.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
+ * Copyright (c) 2008-2022, Hazelcast, Inc. All Rights Reserved.
```
mehmettokgoz, line 2: Just a little comment, the year should be 2023.

internal/serialization/class_definition_context.go
mehmettokgoz: Same comment applies PortableContext.LookUp too.

internal/serialization/class_definition_context.go
```diff
@@ -28,32 +29,39 @@ import (
...
+	k := encodeVersionedClassID(classDefinition.ClassID, classDefinition.Version)
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	current := c.classDefs[k]
 	if current == nil {
-		c.classDefs[classDefKey] = classDefinition
+		c.classDefs[k] = classDefinition
 		return nil
 	}
 	if !reflect.DeepEqual(current, classDefinition) {
```
mehmettokgoz, line 67: I'm not sure if this comment is right but do we expect this `DeepEqual` function to take short time? Because the mutex is unlocked when function returns and it might block other requests to continue. Maybe we can unlock the mutex after setting the `classDefs`?

mehmettokgoz (commented):
Thanks, looks good. Just added some questions and comments.

internal/serialization/portable_context.go
yuce: We cannot lock the mutex for writing while it it is locked for reading. We could use a `sync.Mutex` instead of `sync.RWMutex` at the start of the method and defer unlocking it. But in our use case, we would mostly reading the map and seldomly updating it. Since the mutex can be read locked concurrently but can be write locked only one goroutine, using `sync.Mutex, would create a bottleneck for all  goroutines. So, it is better to use `sync.RWMutex` to check the existence of the value we're looking for.

internal/serialization/class_definition_context.go
yuce: The time `DeepEqual` takes isn't significant in this context, also `register` method is not in the hot path.

internal/serialization/class_definition_context.go
yuce: The copyright year is not very important.

internal/serialization/class_definition_context.go
mehmettokgoz: Thanks.

internal/serialization/portable_context.go
mehmettokgoz: Thanks for explanation.

mehmettokgoz (approved):
LGTM!

PR #921, opened 2023-01-26
Near Cache Fix [API-1854]

yuce:
Avoids creating the backing near cache map and adding an invalidation handler with each `Client.GetMap` call.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008-2022, Hazelcast, Inc. All Rights Reserved.
+ * Copyright (c) 2008-2023, Hazelcast, Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License")
  * you may not use this file except in compliance with the License.
@@ -149,6 +149,9 @@ func (c *Client) GetMap(ctx context.Context, name string) (*Map, error) {
 	m, err := c.proxyManager.getMap(ctx, name)
 	if err != nil {
 		return nil, err
+	}
+	if m.hasNearCache {
+		return m, nil
 	}
 	ncc, ok, err := c.cfg.GetNearCache(name)
 	if err != nil {
--- a/client.go
+++ b/client.go
@@ -146,19 +146,16 @@ func (c *Client) GetMap(ctx context.Context, name string) (*Map, error) {
 	if c.ic.State() != client.Ready {
 		return nil, hzerrors.ErrClientNotActive
 	}
-	m, err := c.proxyManager.getMap(ctx, name)
-	if err != nil {
-		return nil, err
-	}
-	if m.hasNearCache {
-		return m, nil
-	}
-	ncc, ok, err := c.cfg.GetNearCache(name)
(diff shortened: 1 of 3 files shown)


yuce:
A test failed because RC snapshot jar cannot be downloaded, restarting the test.

proxy_manager.go
```diff
@@ -64,10 +64,8 @@ func (m *proxyManager) Proxies() map[string]interface{} {
 	return cp
 }
 
-func (m *proxyManager) getMap(ctx context.Context, name string) (*Map, error) {
-	p, err := m.proxyFor(ctx, ServiceNameMap, name, func(p *proxy) (interface{}, error) {
-		return newMap(p), nil
-	})
+func (m *proxyManager) getMap(ctx context.Context, name string, fn func(p *proxy) (interface{}, error)) (*Map, error) {
+	p, err := m.proxyFor(ctx, ServiceNameMap, name, fn)
```
srknzl, line 68: What we are doing is effectively the same thing but different from other getter methods.(getQueue, getTopic etc.) Are you aware of that?

client.go
```diff
@@ -146,16 +146,16 @@ func (c *Client) GetMap(ctx context.Context, name string) (*Map, error) {
...
+	return c.proxyManager.getMap(ctx, name, func(p *proxy) (interface{}, error) {
+		m := newMap(p)
+		ncc, ok, err := c.cfg.GetNearCache(name)
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			// there is no near cache config for this map
+			return m, nil
+		}
```
srknzl, line 158: I did not really get what is different and fixing the issue here.

nearcache/nearcache_it_test.go
```diff
@@ -1097,6 +1097,23 @@ func TestRepairingTaskRun(t *testing.T) {
 	time.Sleep(2 * time.Second)
 }
 
+func TestSameNearCacheBackingStoreForEachGetMap(t *testing.T) {
```
srknzl, line 1100: I ran this test on current master it passes. What do we fix in this PR?

client.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008-2022, Hazelcast, Inc. All Rights Reserved.
+ * Copyright (c) 2008-2023, Hazelcast, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
yuce: These are changed by the IDE. The copyright year is not too important.

client.go
yuce: A new backing map for the near cache is created every time `GetMap` is called, this PR fixes that.

nearcache/nearcache_it_test.go
yuce: I've changed the test to check for a memory leak at: https://github.com/hazelcast/hazelcast-go-client/pull/921/commits/5fd475e1bc74e079661db9365f3554c6f181862f

srknzl (approved):

PR #922, opened 2023-01-30
Use a sync.Map instead of map + mutex for proxy local cache [API-1855]

yuce:
This PR changes map + sync.RWMutex to `sync.Map` for proxy local cache in order to have better write performance under high concurrency.

### DIFF
--- a/proxy_manager.go
+++ b/proxy_manager.go
@@ -28,7 +28,6 @@ import (
 )
 
 type proxyManager struct {
-	mu              *sync.RWMutex
 	proxies         *sync.Map
 	invocationProxy *proxy
 	serviceBundle   creationBundle
@@ -39,7 +38,6 @@ type proxyManager struct {
 func newProxyManager(bundle creationBundle) *proxyManager {
 	bundle.Check()
 	pm := &proxyManager{
-		mu:             &sync.RWMutex{},
 		proxies:        &sync.Map{},
 		serviceBundle:  bundle,
 		refIDGenerator: iproxy.NewReferenceIDGenerator(1),
@@ -188,8 +186,6 @@ func (m *proxyManager) removeDistributedObjectEventListener(ctx context.Context,
 
 func (m *proxyManager) remove(ctx context.Context, serviceName string, objectName string) bool {
 	name := makeProxyName(serviceName, objectName)
-	m.mu.Lock()
-	defer m.mu.Unlock()
 	p, ok := m.proxies.Load(name)
 	if !ok {
 		return false
--- a/proxy_manager.go
+++ b/proxy_manager.go
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008-2022, Hazelcast, Inc. All Rights Reserved.
+ * Copyright (c) 2008-2023, Hazelcast, Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License")
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


yuce:
I've restarted the tests, since they failed while downloading the RC jar.

proxy_manager.go
```diff
@@ -29,7 +29,7 @@ import (
 
 type proxyManager struct {
 	mu              *sync.RWMutex
-	proxies         map[string]interface{}
+	proxies         *sync.Map
```
srknzl, line 32: According to https://pkg.go.dev/sync#Map:


```
The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

```

Is our case one of these two cases? If not, maybe we should not change map + RWMutex usage to sync.Map. Can you explain the reasoning of the change more in detail ? I see some information in the jira ticket. It may make sense to copy it to the PR desc. 


Looking at the results in the jira ticket, I don't see any improvement. It seems like latencies stayed the same.

I did not get the `operations with t >= 0.5ms` benchmark as well. What is an operation here?

proxy_manager.go
```diff
@@ -55,12 +55,11 @@ func newProxyManager(bundle creationBundle) *proxyManager {
...
 
 func (m *proxyManager) Proxies() map[string]interface{} {
-	cp := make(map[string]interface{}, len(m.proxies))
-	m.mu.RLock()
-	defer m.mu.RUnlock()
-	for k, p := range m.proxies {
-		cp[k] = p
-	}
+	cp := map[string]interface{}{}
+	m.proxies.Range(func(key, value interface{}) bool {
```
srknzl, line 59: There is no way to get the length of `sync.Map`, right?

proxy_manager.go
```diff
@@ -29,7 +29,7 @@ import (
 
 type proxyManager struct {
 	mu              *sync.RWMutex
```
srknzl, line 31: Can't we remove this mutex now? 🤔

srknzl:
The codecov failed somehow(check publish step of https://github.com/hazelcast/hazelcast-go-client/actions/runs/4042281379/jobs/6950796150), can you check the workflow? I am not sure why the workflow did not fail though, I think we should fix it as well.

proxy_manager.go
yuce: I've updated the Jira ticket.

Readers block when a writer holds the RWMutex, that increases latencies in this particular scenario where `GetMap` is called with thousands of names concurrently. `sync.Map` has better latencies with that.

Since the map used for local proxy cache grows but the items there are not updated, `sync.Map` performs better.

proxy_manager.go
yuce: There's no `Len` method, but you can range over it to get the length, which is not useful in this case.

yuce:
@srknzl RC script couldn't download the RC jar: https://github.com/hazelcast/hazelcast-go-client/actions/runs/4042281379/jobs/6949804064 I've restarted test tasks because of that: https://github.com/hazelcast/hazelcast-go-client/pull/922#issuecomment-1408330504

proxy_manager.go
yuce: Yes, good catch, removed at: https://github.com/hazelcast/hazelcast-go-client/pull/922/commits/75741c9ffe1fb3a94937b48de519458410fdf894

proxy_manager.go
srknzl: I asked something about numbers in Jira ticket, after that I am ok with this PR.

srknzl (approved):

srknzl (approved):

Issue #909, opened 2022-12-20
Client Cannot Connect To Cluster When All Members Go Down and Up  [API-1825]


mtyazici:
Go version: 1.19
Hazelcast Go Client version: v1.3.1
Hazelcast server version: 5.2.1
Number of the clients: 1
Cluster size, i.e. the number of Hazelcast cluster members: 3
OS version : Linux

#### Expected behavior

When I delete all member pods in Hazelcast StatefulSet at the same time and recreate them client cannot reconnect to the restarted members and `OrderedMembers` returns the old UUIDs. 

#### Actual behavior

Client should be able to reconnect to the cluster

#### Steps to reproduce the behavior

- Create Default Hazelcast cluster using the Hazelcast Platform Operator.
- Run a Go client in a pod, see the client can access all members
- Delete the StatefulSet, thus deleting all member pods
- Check the client logs and see it is now unable to connect to restarted members.

PR #927, opened 2023-02-03
Try to connect to the cluster infinitely if reconnect mode is ON [API-1825]

yuce:
The client reboots the connection manager if it is disconnected from the cluster (all connections are closed). But it tries rebooting only once, which causes a client shutdown on Kubernetes.

Fixes: https://github.com/hazelcast/hazelcast-go-client/issues/909

### DIFF
--- a/client_internals_it_test.go
+++ b/client_internals_it_test.go
@@ -58,7 +58,7 @@ func TestClientInternal(t *testing.T) {
 		{name: "ClusterID_2", f: clientInternalClusterID_2Test},
 		{name: "ConnectedToMember", f: clientInternalConnectedToMemberTest},
 		{name: "EncodeData", f: clientInternalEncodeDataTest},
-		{name: "InfiniteRestart", f: infiniteRestartTest},
+		{name: "InfiniteRestart", f: infiniteReconnectTest},
 		{name: "InternalInvokeOnKey", f: clientInternalInvokeOnKeyTest},
 		{name: "InternalListenersAfterClientDisconnected", f: clientInternalListenersAfterClientDisconnectedTest},
 		{name: "InvokeOnMember", f: clientInternalInvokeOnMemberTest},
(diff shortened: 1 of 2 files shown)


internal/client/client.go
```diff
@@ -298,15 +298,20 @@ func (c *Client) handleClusterEvent(event event.Event) {
...
-		go c.Shutdown(ctx)
+	for {
+		if c.State() != Ready {
+			break
+		}
+		// try to reboot cluster connection
+		c.ConnectionManager.Stop()
+		c.ClusterService.Reset()
+		c.PartitionService.Reset()
+		if err := c.ConnectionManager.Start(ctx); err != nil {
```
srknzl, line 309: Does Start() call being successful ensure we are connected to the cluster?

yuce: That's right.

client_internals_it_test.go
```diff
@@ -488,6 +490,71 @@ func proxyManagerShutdownTest(t *testing.T) {
 	require.True(t, ok)
 }
 
+func infiniteRestartTest(t *testing.T) {
+	skip.If(t, "enterprise")
```
srknzl, line 494: why skip if enterprise?

client_internals_it_test.go
```diff
@@ -488,6 +490,71 @@ func proxyManagerShutdownTest(t *testing.T) {
 	require.True(t, ok)
 }
 
+func infiniteRestartTest(t *testing.T) {
```
srknzl, line 493: infinite reconnect?

client_internals_it_test.go
yuce: When entprise flag is on, we create clusters with SSL enabled. This test creates a cluster with SSL disabled, so the client wouldn't be able to connect to the cluster if enterprise flag is on.

client_internals_it_test.go
yuce: OK, changed the test name at: https://github.com/hazelcast/hazelcast-go-client/pull/927/commits/1a676da1b5abfe84fad771dbe3bf54f2b9d7d56f

srknzl (approved):

PR #937, opened 2023-02-27
Discovery [API-1954]

yuce:
This PR adds public discovery types.


internal/cluster/discovery.go
```diff
@@ -0,0 +1,179 @@
...
+		mu:          &sync.RWMutex{},
+		lg:          lg,
+		opts: discovery.StrategyOptions{
+			Logger:      lg,
+			UsePublicIP: cfg.UsePublicIP,
+		},
+	}
+}
+
+func (d *DiscoveryStrategyAdapter) Addresses(ctx context.Context) ([]pubcluster.Address, error) {
```
srknzl, line 62: As far as I see, loadAddresses method in java returns private addresses regardless of the usePublicIp config. See https://github.com/hazelcast/hazelcast/blob/7c3894b1474dba56648cf10ad8441615486fcdea/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/discovery/RemoteAddressProvider.java#L42

internal/cluster/discovery.go
```diff
@@ -0,0 +1,179 @@
...
+		if err := d.start(ctx); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *DiscoveryStrategyAdapter) start(ctx context.Context) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
```
srknzl, line 129: Don't we have this lock for the 

```go
// run strategy's Start method if it implements that.
	if starter, ok := d.strategy.(discovery.StrategyStarter); ok {
		d.lg.Debug(func() string {
			return "Starting discovery strategy"
		})
		if err := starter.Start(ctx, d.opts); err != nil {
			return err
		}
	}
```

block? Does moving lock and unlock to the borders of this block make sense? 🤔

internal/cluster/discovery.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (d *DiscoveryStrategyAdapter) updateTranslation(nodes []discovery.Node) {
+	translator := make(map[pubcluster.Address]pubcluster.Address, len(nodes))
+	for _, node := range nodes {
+		// map private addresses to public, if usePublicIP
+		// map public IPs to public
+		if node.PrivateAddr == "" && node.PublicAddr == "" {
+			continue
+		}
```
srknzl, line 161: Do we have these checks in Java side? I think these are confusing. How these can be empty? Here is the code to get privateToPublicAddress map in Java (there is no conditionals): https://github.com/hazelcast/hazelcast/blob/7c3894b1474dba56648cf10ad8441615486fcdea/hazelcast/src/main/java/com/hazelcast/client/impl/clientside/ClusterDiscoveryServiceBuilder.java#L161

internal/cluster/discovery_it_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/hazelcast/hazelcast-go-client/cluster/discovery"
+	"github.com/hazelcast/hazelcast-go-client/internal/it"
+)
+
+func TestDiscoveryStrategy(t *testing.T) {
+	st := &staticDiscoveryStrategy{}
+	tcx := it.MapTestContext{
+		T: t,
+		ConfigCallback: func(tcx it.MapTestContext) {
+			st.addrs = tcx.Cluster.DefaultConfig().Cluster.Network.Addresses
```
srknzl, line 36: Can we set Network.Addresses to nil to ensure discovery works?

internal/cluster/discovery_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+type dummyDiscoverStrategy struct {
+	nodes []discovery.Node
+}
+
+func (d dummyDiscoverStrategy) DiscoverNodes(ctx context.Context) ([]discovery.Node, error) {
+	return d.nodes, nil
+}
+
+func TestDiscoveryStrategyAdapter_Addresses(t *testing.T) {
```
srknzl, line 39: Tests look ok but won't we port any java test?

internal/cluster/discovery_test.go
```diff
@@ -0,0 +1,139 @@
...
+				{PrivateAddr: "10.10.10.10:5000"},
+				{PrivateAddr: "10.11.11.11:5000"},
+			},
+			target: []pubcluster.Address{
+				"10.10.10.10:5000",
+				"10.11.11.11:5000",
+			},
+		},
+		{
+			name: "no public ip with only public adresses",
```
srknzl, line 76: As I said in another comment, empty private address is illegal for a DiscoveryNode

internal/cluster/discovery_test.go
```diff
@@ -0,0 +1,139 @@
...
+				{PrivateAddr: "10.10.10.10:5000", PublicAddr: "20.10.10.10:5000"},
+				{PrivateAddr: "10.11.11.11:5000", PublicAddr: "20.11.11.11:5000"},
+			},
+			target: []pubcluster.Address{
+				"20.10.10.10:5000",
+				"20.11.11.11:5000",
+			},
+		},
+		{
+			name:     "public ip with only private adresses",
```
srknzl, line 99: This case throws in java, if you request public IP but it cannot be translated the client throws. See my other comments, I mentioned about this.

srknzl (commented):
Left some comments

internal/cluster/discovery.go
yuce: Are you sure, there's this line: https://github.com/hazelcast/hazelcast/blob/7c3894b1474dba56648cf10ad8441615486fcdea/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/discovery/RemoteAddressProvider.java#L44
```
privateToPublic = getAddresses.call();
```

internal/cluster/discovery.go
yuce: I didn't understand the question. whole `(d *DiscoveryStrategyAdapter) start` function is protected with the lock.

internal/cluster/discovery.go
yuce: The discovery strategy can send empty public/private addresses. Instead of panicking we are ignoring them.

internal/cluster/discovery_it_test.go
yuce: Sure, at: https://github.com/hazelcast/hazelcast-go-client/pull/937/commits/1f2ad596fc07b107142873a9280f22f94071f0f5

internal/cluster/discovery_test.go
yuce: I allowed it, so we can also refactor cloud discovery as a `DiscoveryStrategy`.

internal/cluster/discovery_test.go
yuce: The idea to allow that is, if the discovery strategy doesn't care about `usePublicIP` setting, it would just return the addresses it has, and the Go client would remedy that by copying them to public addresses if  `usePublicIP == true`.

internal/cluster/discovery_test.go
yuce: The tests in `discovery_test.go` and `discovery_it_test.go` seem to cover the important ones. Discovery strategy in the Go client is more limiited, like the metadata doesn't exist.

internal/cluster/discovery.go
srknzl: That `.call()` calls this method:
 https://github.com/hazelcast/hazelcast/blob/7c3894b1474dba56648cf10ad8441615486fcdea/hazelcast/src/main/java/com/hazelcast/client/impl/clientside/ClusterDiscoveryServiceBuilder.java#L161

but it does not check the usePublicIp config as well.

internal/cluster/discovery.go
srknzl: I am questioning if we need to protect the whole method, won't it be enough to protect only the critical area? Also will doing it improve performance?

internal/cluster/discovery.go
srknzl: It is an option as well but different from Java. Are you sure we want a different behaviour?

internal/cluster/discovery_test.go
srknzl: Can cloud API return empty private addresses?  I think I did not quite get why we are allowing empty/null private addresses. In Java an error is thrown.

internal/cluster/discovery_test.go
srknzl: If usePublicIP == true, why do we use private IP by copying it to public IP? I think go client should respect the usePublicIp config. In java, we throw as far as I see. I did not get why we do differently in go.

internal/cluster/discovery.go
yuce: The only parts in that function except the block is getting/setting the value of `d.started` which should be done under the lock, that doesn't have any affect on the performance.

internal/cluster/discovery_test.go
yuce: Address provider / translator in the Go client probably works differently than the Java client.
Go client's cloud address provider only returns public addresses.

internal/cluster/discovery_test.go
yuce: Go client respects the `usePublicIP` config if both publicIP and privateIP exists. If one of them is missing, Go client tries to remedy that instead of panicking.

internal/cluster/discovery.go
yuce: As far as I see in the Java client there's that comment about private address cannot be null, but there's no check for that. So a discovery strategy may return a null private address. The difference in the Go client is, it checks this and does not update the translation table with the null address.

srknzl (approved):

PR #938, opened 2023-02-28
Removed k8s example

yuce:
`go.mod` and `go.sum` files were causing auto-complete problems in the IDE, so removed the k8s example.


srknzl (approved):
as I said when we huddled, I think we should put the example somewhere else if we remove it but that can be done later on

PR #941, opened 2023-03-08
Merge Features/compact serialization

yuce:
Merges the features/compact branch into `main`.


internal/serialization/field_operations.go
```diff
@@ -0,0 +1,373 @@
...
+type ArrayOfNullableFloat64FieldKindOperation struct{}
+
+func (ArrayOfNullableFloat64FieldKindOperation) KindSizeInBytes() int32 {
+	return variableKindSize
+}
+
+func FieldOperations(fieldKind pserialization.FieldKind) FieldKindOperation {
+	switch fieldKind {
+	case pserialization.FieldKindBoolean:
+		return BooleanFieldKindOperation{}
```
mdumandag, line 287: minor: maybe we can put these into an array for much faster lookups

serialization/serialization_config.go
```diff
@@ -26,6 +26,7 @@ import (
 type Config struct {
 	globalSerializer                    Serializer
 	customSerializers                   map[reflect.Type]Serializer
+	Compact                             CompactConfig
```
mdumandag, line 29: why is this public, while the others are private?

mdumandag (approved):

internal/serialization/field_operations.go
yuce: That makes sense.

serialization/serialization_config.go
yuce: We've decided to have a separate `Compact` section to make it similar to all clients. `CompactConfig` has the private `serializers map[string]CompactSerializer` field and public `SetSerializers` method, similar to global/custom serializers.

PR #942, opened 2023-03-08
Compact Implementation Improvements

yuce:


internal/serialization/compact_it_test.go
```diff
@@ -0,0 +1,131 @@
...
+			if (size != 80) {
+				throw new Error("size is " + size);
+			}			
+		`, tcx.MapName))
+		assert.NoError(t, err)
+	})
+}
+
+func clusterRestartTest(t *testing.T) {
+	// ported from: com.hazelcast.internal.serialization.impl.compact.integration.CompactFormatIntegrationTest#testJoinedMemberQuery
```
mdumandag, line 114: this is the same as above, probably miscopy of the original test from above

internal/serialization/compact_serializer.go
```diff
@@ -95,18 +101,26 @@ func (c CompactStreamSerializer) getOrReadSchema(ctx context.Context, input pubs
...
+	defer func() {
+		if rec := recover(); rec != nil {
+			err = makeError(rec)
+		}
+	}()
 	sw := NewSchemaWriter(ser.TypeName())
 	// create the zero value for the type in the serializer and build the schema
 	ser.Write(sw, reflect.Zero(ser.Type()).Interface())
-	return sw.Build()
+	return sw.Build(), nil
```
mdumandag, line 125: where do we return an error in this function? is the deferred function above recovers from panic and return an error automatically?

serialization/compact_config.go
```diff
@@ -51,3 +63,55 @@ func (cc *CompactConfig) ensureSerializers() {
 		cc.serializers = make(map[string]CompactSerializer)
 	}
 }
+
+func (cc *CompactConfig) checkNoDefaultSerializer() error {
+	// note that the check we are doing is just information.
+	// the default serializers are looked uo before the compact serializer,
```
mdumandag, line 69:
```suggestion
	// the default serializers are looked up before the compact serializer,
```

serialization/compact_config.go
```diff
@@ -51,3 +63,55 @@ func (cc *CompactConfig) ensureSerializers() {
...
+	// so the user cannot override any default serializer.
```
mdumandag, line 70: this part is different from the java implementation and the order defined in https://docs.hazelcast.com/hazelcast/latest/serialization/serialization#steps. why this is the case in go client?

serialization/compact_config.go
```diff
@@ -51,3 +63,55 @@ func (cc *CompactConfig) ensureSerializers() {
...
+	var cr CompactReader
+	var cw CompactWriter
```
mdumandag, line 72: why is these two here?

serialization/compact_config.go
```diff
@@ -51,3 +63,55 @@ func (cc *CompactConfig) ensureSerializers() {
...
+		reflect.TypeOf([]float64{}),
+		reflect.TypeOf(types.UUID{}),
+		reflect.TypeOf(b),
+		reflect.TypeOf(types.Decimal{}),
+		reflect.TypeOf(JSON("")),
+		reflect.TypeOf([]interface{}{}),
+		reflect.TypeOf(types.LocalDate{}),
+		reflect.TypeOf(types.LocalTime{}),
+		reflect.TypeOf(types.LocalDateTime{}),
+		reflect.TypeOf(types.OffsetDateTime{}),
```
mdumandag, line 105: this might be fine for now, but I believe we can easily forget to update this list, when we add a new type to support to serialization service. is there a way to do it there without defining these here?

internal/proto/codec/builtin.go
```diff
@@ -1175,20 +1175,20 @@ func DecodeNullableForSchema(frameIterator *proto.ForwardFrameIterator) *iserial
 	return DecodeSchema(frameIterator)
 }
 
-func EncodeListMultiFrameForFieldDescriptor(message *proto.ClientMessage, fields []*iserialization.FieldDescriptor) {
+func EncodeListMultiFrameForFieldDescriptor(message *proto.ClientMessage, fields []iserialization.FieldDescriptor) {
```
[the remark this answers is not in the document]
yuce: Non-pointer values are usually preferred over pointer ones, one of the reasons for that is they are easier on the GC than the pointers.

internal/serialization/default_compact_writer.go
```diff
@@ -318,14 +318,14 @@ func (r *DefaultCompactWriter) End() {
 }
 
 func (r *DefaultCompactWriter) getFieldDescriptorChecked(fieldName string, fieldKind pubserialization.FieldKind) FieldDescriptor {
-	fd := r.schema.GetField(fieldName)
-	if fd == nil {
+	fd, ok := r.schema.GetField(fieldName)
+	if !ok {
 		panic(ihzerrors.NewSerializationError(fmt.Sprintf("invalid field name: '%s' for %v", fieldName, r.schema), nil))
```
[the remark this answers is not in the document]
yuce: Yes, the panic is caught in the serialization system and sent as an `error` to the user.

internal/serialization/compact_it_test.go
yuce: Replaced with the correct ref here: https://github.com/hazelcast/hazelcast-go-client/pull/942/commits/62dd8899f77fe3fbee9e3e258f319335907831c7

internal/serialization/compact_serializer.go
yuce: Yes, the following block at the top of this function converts the panic to an `error`:
```
	defer func() {
		if rec := recover(); rec != nil {
			err = makeError(rec)
		}
	}()
```

serialization/compact_config.go
yuce: I've added a note at: https://github.com/hazelcast/hazelcast-go-client/pull/942/commits/d6776c52c4a31420deea2b005bd514e0e35e5e7b

serialization/compact_config.go
yuce: https://github.com/hazelcast/hazelcast-go-client/pull/942/commits/d6776c52c4a31420deea2b005bd514e0e35e5e7b

serialization/compact_config.go
yuce: They are not necessary, removed at: https://github.com/hazelcast/hazelcast-go-client/pull/942/commits/5544c931150397ebc71360304bd91933a0b5bb87

mdumandag (approved):

PR #946, opened 2023-03-14
Added compact serialization docs

yuce:


serialization/api.go
```diff
@@ -463,100 +463,244 @@ const (
...
+	// ReadArrayOfDate reads and returns a slice of type.LocalDate pointers.
+	// The returned pointers may be nil.
 	ReadArrayOfDate(fieldName string) []*types.LocalDate
+	// ReadArrayOfTimestamp reads and returns a slice of type.LocalDateTime pointers.
+	// The returned pointers may be nil.
 	ReadArrayOfTimestamp(fieldName string) []*types.LocalDateTime
+	// ReadArrayOfTimestampWithTimezone reads and returns a slice of type.OffsetDateTime pointers.
+	// The returned pointers may be nil.
 	ReadArrayOfTimestampWithTimezone(fieldName string) []*types.OffsetDateTime
+	// ReadArrayOfCompact reads and returns a slice of values.
```
emreyigit, line 545: Is there any chance it could be nil?

serialization/api.go
```diff
@@ -463,100 +463,244 @@ const (
...
+	// WriteNullableInt64 writes an int64 pointer.
+	// The pointer may be nil.
 	WriteNullableInt64(fieldName string, value *int64)
+	// WriteNullableFloat32 writes a float32 pointer.
+	// The pointer may be nil.
 	WriteNullableFloat32(fieldName string, value *float32)
+	// WriteNullableFloat64 writes a float64 pointer.
+	// The pointer may be nil.
 	WriteNullableFloat64(fieldName string, value *float64)
+	// WriteNullableFloat64 writes a float64 pointer.
```
emreyigit, line 685:
```suggestion
 	// WriteArrayOfNullableBoolean writes a slice of bool pointers.
	// The pointers may be nil.
```

serialization/compact_config.go
```diff
@@ -25,10 +25,13 @@ import (
...
 )
 
+// CompactConfig contains compact serializers.
 type CompactConfig struct {
 	serializers map[string]CompactSerializer
 }
 
+// Clone creates a copy of the CompactConfig value.
+// This method is intended for internal use.
 func (cc *CompactConfig) Clone() CompactConfig {
```
emreyigit, line 35: If the intention is to use internally, is there anything disallow you to make it internal?

serialization/api.go
yuce: The slice can be nil, but it is treated as empty slice by Go.

serialization/api.go
yuce: Thanks, fixed at: https://github.com/hazelcast/hazelcast-go-client/pull/946/commits/74f94e6e5b43ddd566d7891b1ad926b69a6f6122

serialization/compact_config.go
yuce: No, there's not. It's not dangerous to call it by the user though.

serialization/api.go
yuce: The values themselves may be nil.

serialization/compact_config.go
emreyigit: Np for me, I raised this since other clients doesn't have that public API.

emreyigit (approved):

serialization/compact_config.go
yuce: We have this method for all types that represent configuration sections.

PR #950, opened 2023-03-16
Update SSL samples docs [API-1837]

yuce:


cluster/doc.go
```diff
@@ -102,5 +102,58 @@ You can configure the frequency of the reconnection attempts using config.Cluste
...
+	var config hazelcast.Config
+	config.Cluster.Network.SSL.Enabled = true
+
+If you need to set a custom tls.Config, for example to turn off verification, you need to assign the tls.Config before any other SSL configuration:
+Check out this page for further details about tls.Config options: https://pkg.go.dev/crypto/tls#Config
+
+	// warning: never set InsecureSkipVerify: true on production
+	// config.Cluster.Network.SSL.SetTLSConfig(&tls.Config{InsecureSkipVerify: true})
+
+You need to set the host name with:
```
srknzl, line 125: Why user needs to set the hostname here? Why do we say "need" instead of "can"? Is it also related with the paragraph above, I did not get it .

examples/ssl/mutual_authentication/main.go
```diff
@@ -24,31 +24,38 @@ import (
...
-	cfg = hazelcast.Config{}
-	cfg.Cluster.Network.SetAddresses("foo.bar.com:8888")
+	cfg.Cluster.Network.SetAddresses("localhost:5701")
+	ssl := &cfg.Cluster.Network.SSL
 	// TLS/SSL is enabled
-	cfg.Cluster.Network.SSL.Enabled = true
+	ssl.Enabled = true
+	// we set insecureSkipVerify, since the sample certificates for this example were self-signed.
+	// do not enable it in production.
+	ssl.SetTLSConfig(&tls.Config{InsecureSkipVerify: true})
```
srknzl, line 40: Isn't what we do is against "If you need to set a custom tls.Config, for example to turn off verification, you need to assign the tls.Config before any other SSL configuration:"? we are setting ssl config before setting tls config. (ssl.Enabled)

examples/ssl/mutual_authentication/main.go
```diff
@@ -24,31 +24,38 @@ import (
...
 	// Absolute paths of PEM files must be given
-	err := cfg.Cluster.Network.SSL.SetCAPath("/path/of/server.pem")
+	err := ssl.SetCAPath("examples/ssl/etc/ca.crt")
 	if err != nil {
-		return
+		panic(fmt.Errorf("loading CA certificate: %w", err))
 	}
 	// Add client certificate and client private key to TLS config by decrypting private key using password
-	err = cfg.Cluster.Network.SSL.AddClientCertAndEncryptedKeyPath("/path/of/cert.pem", "path/of/key.pem", "password")
+	err = ssl.AddClientCertAndKeyPath("examples/ssl/etc/client.crt", "examples/ssl/etc/client.key")
```
srknzl, line 48: Why don't we pass a password argument now? The comment above this line is not accurate now since we don't pass a password.

yuce: The previous example used an encrypted certificate, this one uses a non-encrypted one.

examples/ssl/mutual_authentication/main.go
yuce: Clarified at: https://github.com/hazelcast/hazelcast-go-client/pull/950/commits/ff7272c9f6d6de1cb8f9c78e15dab2027566f282

cluster/doc.go
yuce: Otherwise Go std lib returns an error that it must be set.

examples/ssl/mutual_authentication/main.go
yuce: Updated at: https://github.com/hazelcast/hazelcast-go-client/pull/950/commits/3eab877103d5fa0fe4afde51b5a1d36b4ef6a382

PR #964, opened 2023-08-04
[CN-922] Fix IPv6 address parsing error

### LINKED ISSUE #963: Error on connection IPv6 cluster
Go version: 1.19
Hazelcast Go Client version: latest/1.4.0
OS version : Linux/OSX

When I try to connect to IPv6 Hazelcast cluster, It fails as described below.

```go
conf := hazelcast.Config{}
conf.Cluster.Network.Addresses = []string{"2600:1f11:****:eb50:5689::"}

client, err := hazelcast.StartNewClientWithConfig(ctx,conf)
if err != nil {
   panic(err)
}
```

```
panic: invalid address '[2600:1f11:****:eb50:5689::]': address [2600:1f11:****:eb50:5689::]: missing port in address

goroutine 1 [running]:
main.main()
	.../hz-connection/go/main.go:19 +0xd4
exit status 2
```

It seems like the address parsing is not compatible with IPv6.
https://github.com/hazelcast/hazelcast-go-client/blob/master/internal/addr.go#L28

The issue is also reported in Operator repo.
https://github.com/hazelcast/hazelcast-platform-operator/issues/806

semihbkgr:
Fixes #963


internal/addr_test.go
```diff
@@ -40,6 +40,10 @@ func TestParseAddr(t *testing.T) {
 		{Addr: "foo", Target: parseAddrTarget{Host: "foo", Port: 0, Err: nil}},
 		{Addr: ":1234", Target: parseAddrTarget{Host: "127.0.0.1", Port: 1234, Err: nil}},
 		{Addr: "", Target: parseAddrTarget{Host: "127.0.0.1", Port: 0, Err: nil}},
+		{Addr: "169.254.172.39", Target: parseAddrTarget{Host: "169.254.172.39", Port: 0, Err: nil}},
+		{Addr: "169.254.172.39:5703", Target: parseAddrTarget{Host: "169.254.172.39", Port: 5703, Err: nil}},
+		{Addr: "fe80::43:ecff:fec9:1683", Target: parseAddrTarget{Host: "fe80::43:ecff:fec9:1683", Port: 0, Err: nil}},
+		{Addr: "[fe80::43:ecff:fec9:1683]:5705", Target: parseAddrTarget{Host: "fe80::43:ecff:fec9:1683", Port: 5705, Err: nil}},
```
yuce, line 46: @semihbkgr Could you also add a test case for `[fe80::43:ecff:fec9:1683` and `fe80::43:ecff:fec9:1683]:5678`?

internal/addr.go
```diff
@@ -26,16 +26,12 @@ import (
...
 		return defaultHost, 0, nil
 	}
-	if !strings.Contains(addr, ":") {
-		return addr, 0, nil
-	}
 	host, port, err := net.SplitHostPort(addr)
-	if err != nil {
-		return "", 0, err
+	if err != nil { // return the address itself if port is missing in the address
+		return addr, 0, nil
```
yuce, line 34: Returning the address on an error doesn't look correct to me. It would be better to detect whether the address is ipv6 without a port and return early but still return an error if there's a problem with parsing the address.

yuce (changes requested):
Some ideas...

internal/addr_test.go
semihbkgr: Ok, we assume that these are not valid addresses, right?

internal/addr.go
semihbkgr: I agree with you. It is not the best practice as It ignores the errors. It will lead to more potential bug because It doesn't cover any undesired cases.

internal/addr.go
```diff
@@ -44,8 +43,21 @@ func ParseAddr(addr string) (string, int, error) {
...
 }
+
+func hasPort(addr string) bool {
+	i := strings.LastIndex(addr, ":")
+	if i == -1 {
+		return false
+	}
+	if i == len(addr)-1 {
+		return true
+	}
```
yuce, line 60: Is this check necessary?

internal/addr.go
```diff
@@ -26,16 +26,15 @@ import (
...
 		return defaultHost, 0, nil
 	}
-	if !strings.Contains(addr, ":") {
+	if !hasPort(addr) {
 		return addr, 0, nil
 	}
 	host, port, err := net.SplitHostPort(addr)
 	if err != nil {
-		return "", 0, err
+		return host, 0, err
```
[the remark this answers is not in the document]
yuce: Could you return `ihzerrors.NewClientError("invalid address", err, hzerrors.ErrInvalidAddress)` here? `ihzerrors` is an import alias to `"github.com/hazelcast/hazelcast-go-client/internal/hzerrors"`.

internal/addr.go
```diff
@@ -44,8 +43,21 @@ func ParseAddr(addr string) (string, int, error) {
 	if host == "" || strings.TrimSpace(host) == "" {
 		host = defaultHost
 	}
-	if portInt < 0 { // port number should be more than 0
+	if portInt < 0 {
+		// the port number must be positive value
 		return "", 0, fmt.Errorf("invalid port number: '%d'", portInt)
 	}
```
yuce, line 49: How about:
```
	if portInt < 0 {
		return "", 0, hzerrors.ErrInvalidAddress
	}
	if host == "" || strings.TrimSpace(host) == "" {
		host = defaultHost
	}
```

yuce (changes requested):
Here are some ideas...

internal/addr.go
semihbkgr: I wouldn't support to return `hzerrors.ErrInvalidAddress` as it is not descriptive.

internal/addr.go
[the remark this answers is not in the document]
semihbkgr: `host` is already empty string when the error is not `nil`.

internal/addr.go
semihbkgr: It is necessary as we get `i+1` as an index of `addr` at the rest

internal/addr.go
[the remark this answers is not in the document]
yuce: So, why not return `""` and make the purpose explicit?

internal/addr.go
yuce: That condition makes the function return `true` for `192.168.1.1:`. Is that what was intended?

yuce: Returning `true` for something like `192.168.1.1` is misleading, and may be an indicator that `hasPort` isn't suitable for our case. Maybe moving the code that splits host and port outside of `ParseAddr`is better. How about something like:
```go

func ParseAddr(addr string) (string, int, error) {
	addr = strings.TrimSpace(addr)
	if addr == "" {
		return defaultHost, 0, nil
	}
	host, port, err := splitHostPort(addr)
	if err != nil {
		return "", 0, ihzerrors.NewClientError("", err, hzerrors.ErrInvalidAddress)
	}
	if port < 0 || port > 65535 {
		return "", 0, ihzerrors.NewClientError("port not in valid range", err, hzerrors.ErrInvalidAddress)
	}
	return host, port, nil
}

func splitHostPort(addr string) (host string, port int, err error) {
	idx := strings.LastIndex(addr, ":")
	if idx == -1 {
		return addr, 0, nil
	}
	port, err = strconv.Atoi(addr[idx+1:])
	if err != nil {
		return "", 0, fmt.Errorf("invalid port")
	}
	host, _, err = net.SplitHostPort(addr)
	if err != nil {
		return "", 0, err
	}
	if host == "" {
		host = defaultHost
	}
	return host, port, nil
}
```

semihbkgr: This `splitHostPort ` would panic when calling with `192.168.1.1:` arg.

yuce: What's the panic you're getting?
The following works OK without panics:
```go

func splitHostPort(addr string) (host string, port int, err error) {
	idx := strings.LastIndex(addr, ":")
	if idx == -1 {
		return addr, 0, nil
	}
	port, err = strconv.Atoi(addr[idx+1:])
	if err != nil {
		return "", 0, fmt.Errorf("invalid port")
	}
	host, _, err = net.SplitHostPort(addr)
	if err != nil {
		return "", 0, err
	}
	if host == "" {
		host = defaultHost
	}
	return host, port, nil
}

func main() {
	host, port, err := splitHostPort("192.168.1.1:")
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(host, port)
}
```

semihbkgr: It works well, and is better way than checking if address has port separately.

yuce (approved):

PR #977, opened 2023-09-25
Fixed Map LoadAll

yuce:
Calling `Map.LoadAll` with no keys must load all keys.


yuce:
Superseded by https://github.com/hazelcast/hazelcast-go-client/pull/1036

PR #1043, opened 2026-02-19
internal/cp: fix dropped errors

alrs:
This picks up three dropped `err` variables in `internal/cp`. The unit tests pass both before and after these changes.


yuce (approved):
Thanks for your PR!

PR #1052, opened 2026-04-27
[HZ-5444] Improve Map.GetAll Performance

yuce:
Although the `Map.GetAll` method must send keys for partitions concurrently, in the current implementation they were sent serially which caused a performance regression.
This PR fixes that by sending keys and receiving the values concurrently in separate goroutines.
This is accomplished by running `tryChan` method of the circuit breaker in a goroutine.

This PR also preallocates the frame buffer with the required size instead of growing it.

These improvements improved the GetAll latency 5 times.


emreyigit (approved):
LGTM

